PR #25, opened 2014-04-18
Throw LowMemoryException on getDrawable(final String aFilePath)

illarionov:
It is more a question rather than a pull request.
Why the `getDrawable(String aFilePath)` does not throw an `LowMemoryException`?  It is to maintain compatibility somwhere?
It looks very strange. The [TileSource#getDrawable(String aFilePath)](https://github.com/illarionov/osmdroid/blob/master/osmdroid-android/src/main/java/org/osmdroid/tileprovider/tilesource/ITileSource.java#L54) allows to throw this exception.


spyhunter99:
due to merge issues, i merged it into a separate branch. if everything passes, i'll merge it tonight

spyhunter99:
merged

Issue #46, opened 2014-06-17
Support up to 1E7 instead of 1E6


kjeremy:
We had to bump the precision up to 7 decimal places in an older version of osmdroid (I believe 3.0.4, my coworker did not leave great notes) and increase the zoom levels to 22-23 for sub meter accuracy.  I'd like to get this work into the latest version of osmdroid if at all possible.  Would something like this be useful?

PR #48, opened 2014-06-25
Convert from E6 to doubles.

kjeremy:
I haven't been able to test this yet but it should cover most if not all of the int to double case from #46


spyhunter99:
Couldn't this be included with backwards compatibility?
I.E. keep the current API as is, add additional methods that have double support.

I've ran into some precision related issues with osmdroid and the map extents/map center not aligning precisely where it's supposed to be. On an 8in table at zoom 15 or so, the map could be off by as much as 1/8 inch.

spyhunter99:
Also, a float may also be a better choice for internal presentation of points vs double. In the case of lat/lon, the precision available is most likely going to be good enough and the calculations could be handled by a gpu

neilboyd:
If you have double/float **and** int then you'd need to store both in order to avoid conversion, which will be a pain

spyhunter99:
I"m just saying the internal representation can be whatever, then what's presented via the API can be flexible via conversion, which is exactly what we have now. int/double available via API with an int for internal representation. The issue is that using ints for the internal representation has an accuracy limitation.

spyhunter99:
I did some work on this, pushed to osmdroid/PR48. it's based on your PR but alters the code to maintain the existing integer based APIs while using doubles internall and add additional methods to get the double values. it's also up to date now. this will need some performance impact/memory allocation assessments before merging

MKergall:
+10 on this change. I also suggest to:
- rename GeoPoint to LatLng
- and set latitude, longitude and altitude as public. 

Quite easy to do, and a huge improvement of compatibility with Google Maps V2 API (V1 being completely obsolete now).

kjeremy:
:+1:

spyhunter99:
i'm going to close this PR since it has many issues. However, it is operational on the branch "doubles". Once issue #237, we can make a merge decision and move forwards. Thanks for doing the initial work on this!

PR #76, opened 2014-11-08
APK provider to bundle tiles with app itself

everplays:
With this change tiles can be bundled into assets of apk itself. With
this, people can easily make apps that load first couple of zoom levels
from apk and for the rest normal network/cache behavior be used.

P.S. It's not magic or something that can't be done by developers themselves. It just makes it easy to have offline map with tiles already bundled into apk.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -35,3 +35,6 @@ proguard/
 build/
 .gradle
 
+# vim swap files
+*.swp
+*.swo
--- a/osmdroid-android/src/main/java/org/osmdroid/tileprovider/MapTileProviderBasic.java
+++ b/osmdroid-android/src/main/java/org/osmdroid/tileprovider/MapTileProviderBasic.java
@@ -1,6 +1,9 @@
 package org.osmdroid.tileprovider;
 
+import android.content.Context;
+
 import org.osmdroid.tileprovider.modules.INetworkAvailablityCheck;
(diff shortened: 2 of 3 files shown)


ziem:
How to display tiles from assets instead of Mapnik? I have custom painted tiles and I want to display them in some locations instead of Mapnik. In other locations I want to have Mapnik. Is it possible?

everplays:
@ziem Have you read doc block above the class definition? https://github.com/osmdroid/osmdroid/pull/76/files#diff-473fe0b9dc9b9867db4f90f434f03365R24 That's how you can display tiles from assets.

And yes, that is possible to display some tiles from assets and others from the source. You can implement your own tile provider just like what I've done for MapTileProviderBasic (First checks assets and then falls back to the default provider which is Mapnik in your case).

Although, I haven't tried it myself, I think you can achieve the same with showing two tile sources but one on top of another.

ziem:
@everplays yes. I managed to display tiles from assets using your classes. But they were replaced by Mapnik tiles bacause of: https://github.com/osmdroid/osmdroid/blob/master/osmdroid-android/src/main/java/org/osmdroid/tileprovider/modules/MapTileAssetsProvider.java#L145 .

Thanks for hints. I will try to implement my own tile provider.

PR #77, opened 2014-11-17
hikebikemap added

Sash0k:
Hello. I added tiles for hikebikemap. Can you include it into osmdroid, if possibly?


spyhunter99:
ok so this is merged. i looked at the map source in a browser and it's not much different than the standard osm data. The cool part is hill shading. I'm not sure how to turn it on and it's not included in your PR. in closing, the source has been added, however i'm not sure if it met your intent. feel free to reopen

PR #122, opened 2015-06-08
Update of several dependencies and plugins. Moved some directories ar…

fpoyer:
…ound so that each module matches the Maven default configuration.

Updated dependencies:
Android 4.4.2            r3     -> r4
Google Play Services     23.0.0 -> 25.0.0
Compatibility v4 library 22.0.0 -> 22.2.0

Updated plugin:
com.jayway.maven.plugins.android.generation2:android-maven-plugin v3.8.2 replaced with com.simpligility.maven.plugins:android-maven-plugin v4.2.1 to allow for support of more recent versions of Maven.

Directories moved are AndroidManifest.xml, res and assets moved into src/main/ and java sources directory moved into src/main/java (instead of just src).


fpoyer:
This is my very first PR on an open-source project, so let me know if I did anything wrong.
I rebased and squashed my history into a single clean commit.

neilboyd:
Thanks for your contribution. Seems reasonable but I probably won't have time to look at it until at least next week.

neilboyd:
It works so I merged it. Thanks.

emdete:
this patch does not considere gradle, right? at least i do not see build.gradle mentioned which has refeferences that don't exists (for example the "-SNAPSHOT" ones).

neilboyd:
It is considered in as much as that I built the project in Android Studio which uses Gradle.

Please continue this discussion back on issue #126.

Issue #125, opened 2015-06-16
Feature request, change map tile providers during runtime


spyhunter99:
I have a requirement to change from the default map provider, which checks [assets, file cache, file system, online] to a file only map provider [just file system]. Unfortunately, this is set only during the constructor for osmdroid, meaning that these settings must be set in the layout file. This is not ideal and in my use case, i'd be prompting the user for a set of files to use, then update the osmdroid accordingly. I've already done the work to support this and will be attaching a pull request to this.

PR #136, opened 2015-07-10
#125 adding support to change tile providers at runtime

spyhunter99:
Redo, to keep standard formatting


spyhunter99:
@neilboyd ping

Issue #140, opened 2015-07-15
User selectable offline map sources


spyhunter99:
Referenced here
http://stackoverflow.com/questions/31347019/osmdroid-loading-multiple-offline-mbtiles-via-listview

PR #141, opened 2015-07-15
#140 adding the basics for user selectable offline map sources

spyhunter99:
#140 adding the basics for user selectable offline map sources instead of everything that's available offline and online


spyhunter99:
can we get this merged at some point?

spyhunter99:
ping

PR #143, opened 2015-07-18
Improvement: MapView.zoomToBoundingBox() can be used in onCreate() or…

k3b:
… onCreateView.

Implemented workaround from http://stackoverflow.com/questions/10411975/how-to-get-the-width-and-height-of-an-image-view-in-android/10412209#10412209

so that the MapView user does not have to do this.


k3b:
My change is not neccessary anymore. you can set the initial position (zoom, center, BoundingBoxE6) in

<pre>MapView.addOnFirstLayoutListener(new OnFirstLayoutListener() { ...});</pre>

mstaniszczak:
Thank you! You even not know, how long I looking for that:)

Issue #154, opened 2015-08-25
Allow to change OverlayManager


dozd:
1. It would be nice to allow change of the default `OverlayManager` in `MapView`.
2. Also, I'd like to see more hook methods in `Overlay` - when attached / deattached to `MapView`. But this could be done by custom implementation of OverlayManager to keep default one simple.

I can provide pull requests for both scenarios when approved.

spyhunter99:
go ahead and open a PR

spyhunter99:
@dozd ping. i'm eager to get the next versions of osmdroid out there. are you working on this?

dozd:
Sry, been out for a while. I'll try to prepare PR for 1. till wednesday.

spyhunter99:
quite alright, thanks for the PR. I'll second your comment on overlay manager extending abstract list. I'm not sure why. Looks like @neilboyd originally authored that part. It's probably some kind of concurrency wrapper for CopyOnWriteArrayList. The extends AbstractList<Overlay> could be removed and the getters and setters related to it be added to your interface. Let's get Neil's opinion on this. There may be a reason why it is the way it is.

PR #179, opened 2015-10-26
#154 Possibility to change the OverlayManager in MapView.

dozd:
I don't like the idea of extending List in OverlayManager interface much but it would be big refactoring.


spyhunter99:
love it. i'll get it out tonight and hopefully get it merged

spyhunter99:
merged

PR #206, opened 2015-12-07
Implemented Animated Zoom To ZoomLevel

johnnynotsolucky:
Added an animated zoomTo() method as well as implemented it in the zoomToBoundingBox() by providing an animate flag.

The existing zoomIn() and zoomOut() methods call the new zoomToFixing() method.

Additionally, the ZoomAnimatorListener was renamed and now includes the @TargetApi() annotation which is required due to the listener only being available in a newer API.

### DIFF
--- a/OpenStreetMapViewer/src/main/java/org/osmdroid/samplefragments/SampleAnimatedZoomToLocation.java
+++ b/OpenStreetMapViewer/src/main/java/org/osmdroid/samplefragments/SampleAnimatedZoomToLocation.java
@@ -2,11 +2,6 @@ package org.osmdroid.samplefragments;
 
 import android.content.Context;
 import android.location.Location;
-import android.os.Bundle;
-import android.view.Menu;
-import android.view.MenuInflater;
-import android.view.MenuItem;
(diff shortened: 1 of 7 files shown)


spyhunter99:
Awesome! would it be possible to add an example on the sample app?

spyhunter99:
This doesn't compile. Please make sure you build the entire project using maven. We don't have a 100% conversion to gradle at the moment and thus there's a lot still using maven

osmdroid-android/src/main/java/org/osmdroid/views/MapController.java:[431,84] error: inconvertible types

osmdroid-third-party/src/main/java/org/osmdroid/google/wrapper/MapController.java:[14,7] error: MapController is not abstract and does not override abstract method zoomToFixing(int,int,int) in IMapController

tyronetudehope:
My bad, I didn't think to build all the modules.

I added `Animated Zoom to Location` sample to the samples list, which simply zooms to the the user's current location when the marker is tapped.

spyhunter99:
cool. i like.

spyhunter99:
thanks for the contribution!

Issue #194, opened 2015-11-13
build faild~~ no support:compatibility


ikkyu0319:
Error:(32, 13) Failed to resolve: android.support:compatibility-v4:23+

neilboyd:
Please read the relevant [wiki](https://github.com/osmdroid/osmdroid/wiki) for your build method

spyhunter99:
you may need to update your android sdk

spyhunter99:
I'm going to go ahead a close this. Please reopen if this is still a problem for you @ikkyu0319. No one else appears to be having the problem and most likely the issue is that you need to run the android sdk updater tool

rastikw:
What helped me was to change the dependencies line to:
```
compile 'com.android.support:support-v4:23+'
```

ZKjellberg:
When building using **gradle**, the **OpenStreetMapViewer\build.gradle** will fail as it cannot discover
>compile 'android.support:compatibility-v4:23+'

As @rastikw suggested, replacing this with the below value will resolve the issue.
>compile 'com.android.support:support-v4:23+'

PR #208, opened 2015-12-08
Fixed gradle build error for support:compatibility 

ZKjellberg:
Current project features a build error as it is unable to resolve **support:compatibility** in the **OpenStreetMapViewer\build.gradle**. This error is referenced in issue #194.

> Error:(32, 13) Failed to resolve: android.support:compatibility-v4:23+

This is resolved by replacing the dependency:
> compile 'android.support:compatibility-v4:23+'

with

> compile 'com.android.support:support-v4:23+'

### DIFF
--- a/OpenStreetMapViewer/build.gradle
+++ b/OpenStreetMapViewer/build.gradle
@@ -2,7 +2,7 @@ apply plugin: 'com.android.application'
 
 android {
     compileSdkVersion 23
-    buildToolsVersion "23.0.1"
+    buildToolsVersion "23.0.2"
 
     defaultConfig {
         applicationId "org.osmdroid.example"
@@ -27,7 +27,6 @@ android {
         abortOnError false
     }
 }
-
 
 dependencies {
     compile 'com.android.support:support-v4:23+'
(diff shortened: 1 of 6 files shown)


johnnynotsolucky:
It should be possible to remove the dependency altogether as osmdroid handles all its own API conditions. 

The project builds without the dependency, but that's as far as I've tested it.

spyhunter99:
@johnnynotsolucky it is required for the example app. When removing it from the maven build, it fails. Also fails with 'gradle clean build'

spyhunter99:
Zach, wouldn't it be better to use 'buildToolsVersion=23+'? that would reduce our need to maintain these files everytime google decides to push an update

ZKjellberg:
@johnnynotsolucky As @spyhunter99 has stated, if the user does a clean build, the dependency is still required. For example, the OpenStreetMapViewer module's **MapActivity extends FragmentActivity** from this support library. I see it as still necessary to change this gradle dependency as the original does not appear to exist and this allows building.

@spyhunter99 I actually did not intend to push up those changes yet for building. I should have branched and made a seperate pull request to acknowledge the cleanup, but forgot pull requests automatically pull updates. (Slightly rusty on my git, relearning bad behavior.) As for your suggestion to use **buildToolsVersion=23+**,  Android Studio does not seem able to accept this usage. I recognize it is a hassle to maintain the minor updates, but figure it would be worthy to tweak every so often with minor maintenance.

PR #209, opened 2015-12-09
Improved Readme Screenshots

ZKjellberg:
Generated new screenshots with modern UI.

The previous screenshots appeared to be from Android 2.2 Froyo which is ancient.


neilboyd:
Thanks, that's great! :+1: 
It would be even better if you could also replace the third screenshot - the one with the description on the marker :)

ZKjellberg:
@neilboyd I could not figure out how to reproduce the description marker inside the Sample application. Single and long presses do not generate the dialog.

Additionally, I think the sample app should have better names for each Activity to better depict their purpose. I will try to better name the activities and take new screenshots later today. If you can provide guidance on the description marker, I will include an image of.

neilboyd:
There used to be a sample with this. Maybe it's not there any more.

neilboyd:
Reminder to self: also update Google Play

spyhunter99:
I think the sample is still there. Launch the app, hit the first item. Then the menu, samples, and it's one of those.

ZKjellberg:
@spyhunter99 I followed your instructions to discover the second set of Samples. Sadly, the description marker does not appear to work correctly as the Marker is very small and is not legible.

screenshot_20151209-131541

I tested using a Nexus 5 (screenshot source) and a Nexus 7 (2013) which are both running Android 6.0.

ZKjellberg:
Updated the screenshots as mentioned. As the description Activity did not properly display, I did not include it. If it is possible to fix the code, I would be happy to add an image in the future. (Note, I am simply using the build-in Android Studio screenshot tool which provides the frame and fancy effects.)

Additionally, I updated some of the sample code to feature titles for the samples and added default zoom levels and locations to improve the user experience compared to the default instance of seeing multiple worlds.

PR #211, opened 2015-12-09
Tile Caching to Include an Expires header instead of Last Modified Date

johnnynotsolucky:
I've modified the caching process include an `Expires` header at the top of each cached tile. This replaces, for the most part, the `file.lastModifiedDate()` check. 

The idea is that if an `Expires` header is received by the `MapTileDownloader`, the date is added as a header to the tile file. If there was no `Expires` header, then the `OpenStreetMapTileProviderConstants.DEFAULT_MAXIMUM_CACHED_FILE_AGE` is used instead.

If there are existing files with no Expires headers, then the original `file.lastModifiedDate()` condition is used to determine whether the file has expired or not.

I'm not sure if this would be something that could be included in the project as I think simply setting your own maximum cached file age is enough generally. However it may be useful if you'd like server to determine how long files should be cached.

P.S. I added code to the `StreamUtils` class which comes directly from the `DiskBasedCache` class from the Android Volley framework. I'm not 100% sure where to include a reference to the framework project?

### DIFF
--- a/osmdroid-android/src/main/java/org/osmdroid/tileprovider/modules/MapTileDownloader.java
+++ b/osmdroid-android/src/main/java/org/osmdroid/tileprovider/modules/MapTileDownloader.java
@@ -218,7 +218,11 @@ public class MapTileDownloader extends MapTileModuleProviderBase {
 					mFilesystemCache.saveFile(tileSource, tile, byteStream);
 					byteStream.reset();
 				}
-				final Drawable result = tileSource.getDrawable(tile, byteStream);
+
+				tile.readHeaders(byteStream);
+				byteStream.reset();
+
+				final Drawable result = tileSource.getDrawable(byteStream);
 
 				return result;
 			} catch (final UnknownHostException e) {
--- a/osmdroid-android/src/main/java/org/osmdroid/tileprovider/modules/MapTileDownloader.java
+++ b/osmdroid-android/src/main/java/org/osmdroid/tileprovider/modules/MapTileDownloader.java
@@ -216,8 +216,6 @@ public class MapTileDownloader extends MapTileModuleProviderBase {
 				}
 				tile.setExpires(dateExpires);
 
-				tile.writeHeaders(out);
-
 				StreamUtils.copy(in, out);
 				out.flush();
 				final byte[] data = dataStream.toByteArray();
@@ -228,9 +226,6 @@ public class MapTileDownloader extends MapTileModuleProviderBase {
 					mFilesystemCache.saveFile(tileSource, tile, byteStream);
 					byteStream.reset();
 				}
-
-				tile.readHeaders(byteStream);
-				byteStream.reset();
 
 				final Drawable result = tileSource.getDrawable(byteStream);
 
--- a/osmdroid-android/src/main/java/org/osmdroid/tileprovider/modules/MapTileDownloader.java
+++ b/osmdroid-android/src/main/java/org/osmdroid/tileprovider/modules/MapTileDownloader.java
@@ -10,6 +10,7 @@ import java.io.OutputStream;
 import java.net.UnknownHostException;
 import java.text.SimpleDateFormat;
 import java.util.Calendar;
+import java.util.Date;
 import java.util.Locale;
 import java.util.concurrent.atomic.AtomicReference;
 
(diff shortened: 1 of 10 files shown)


osmdroid-android/src/main/java/org/osmdroid/tileprovider/modules/MapTileDownloader.java
```diff
@@ -191,11 +193,21 @@ public Drawable loadTile(final MapTileRequestState aState) throws CantContinueEx
...
 
-				
 				in = c.getInputStream();
 
 				final ByteArrayOutputStream dataStream = new ByteArrayOutputStream();
 				out = new BufferedOutputStream(dataStream, StreamUtils.IO_BUFFER_SIZE);
+
+				String expires = c.getHeaderField("Expires");
+				if(expires == null) { //Expire yesterday
+					SimpleDateFormat dateFormat = new SimpleDateFormat("EEE, dd MMM yyyy HH:mm:ss z");
```
spyhunter99: Please pull this into a constants class

osmdroid-android/src/main/java/org/osmdroid/tileprovider/tilesource/BitmapTileSourceBase.java
```diff
@@ -137,10 +167,26 @@ public Drawable getDrawable(final InputStream aFileInputStream) throws LowMemory
...
+		} catch (final IOException e) {
+			Log.e(IMapView.LOGTAG,"IOException loading bitmap");
 		}
 		return null;
 	}
 
+	private boolean readExpiresHeader(MapTile aTile, InputStream inputStream) throws IOException {
+		try {
+			String expires = StreamUtils.readString(inputStream);
+			SimpleDateFormat dateFormat = new SimpleDateFormat("EEE, dd MMM yyyy HH:mm:ss z");
```
spyhunter99: move to constants

osmdroid-android/src/main/java/org/osmdroid/tileprovider/tilesource/ITileSource.java
```diff
@@ -39,11 +39,13 @@
...
 	 * Get a rendered Drawable from the specified file path.
 	 *
+	 * @param aTile
+	 * 						a tile
 	 * @param aFilePath
 	 *            a file path
 	 * @return the rendered Drawable
 	 */
-	Drawable getDrawable(String aFilePath) throws LowMemoryException;
+	Drawable getDrawable(MapTile aTile, String aFilePath) throws LowMemoryException;
```
spyhunter99: this part erks me a bit. Is there a way to do perform the same function without modifying the interface? there's a ton of tile sources out there. merging this would cause compatibility issues

spyhunter99:
see comments i left on the commits. i think it's a great idea and appreciate the contribution. Make sure the entire project builds.

johnnynotsolucky:
I've pulled out all the strings into the constants class and have restored the ITileSource interface. 

As a work-around, I added methods to the MapTile class to read and write header data to streams. An additional method, `skipOverHeaders`, skips over all stream headers to position the stream at the start of the bitmap data which effectively restores the original functionality of both the `getDrawable()` methods in BitmapTileSourceBase.

neilboyd:
I just had a quick look at this, but if I understand it correctly you're writing a string with the expiry time at the beginning of the tile files that are written to disk.  Doesn't that mean that it's not a valid image file any more?

I can see a few alternative possibilities:

1. Write to the *end* of the file. Maybe that keeps the file valid.
2. Write a proper header to the image file, assuming that's possible. But that would require it to be decoded when downloading which is too big of an overhead.
3. Write the expiry time somewhere else, eg an extra file next to the image file.

I appreciate the contribution :+1:

johnnynotsolucky:
Your understanding is correct, and yes, the tiles would no longer be valid images. I think having an extra file might be the most practical solution as it could probably more easily extended should there ever be a need for additional information linked to a tile in the disk cache.

neilboyd:
Off the top of my head, a file with the same name as the tile but with a json extension and json content.
In order to avoid looking for this file when you know it's not there, add a property to tile module/source to say whether it's expected.

johnnynotsolucky:
Sounds good. Would it be quicker to write data directly to the file instead of reading/writing json content? We can bypass the need for extra processing to map data to json, and vice versa, and instead control the order in which information is written, and then simply read the information back in the same order.

neilboyd:
I don't know. When all you have is a hammer everything looks like a nail! I think json processing is pretty efficient these days. It only needs to be processed when it's read, not when it's written. Compared to processing a bitmap it shouldn't be much.

spyhunter99:
On the other hand, if we add json processing, it'll most likely add another dependency. I'm already at the dex limit and always appreciate anything that keep the method counts low

johnnynotsolucky:
Instead of JSON, we could have a simple properties file with one property per line? Its easy and fast enough to process

spyhunter99:
i'm down with a properties file. no extra dependencies and easy to read

johnnynotsolucky:
I've pulled out the code which wrote header data to the tile itself, and replaced with a properties file with same name as each tile. Only `MapFilesystemProvider` and `TileWriter` make use of the methods at the moment. 

The two new methods in `MapTile`, `writeProperties()` and `readProperties()` take a stream as a parameter so that we don't limit their usefulness if other providers needed to make use of them.

The properties themselves are stored as key/value pairs, with a single pair to a line in the file.

spyhunter99:
Sorry to be a bother on this one, but doesn't java.utils.Properties have a load method? https://docs.oracle.com/javase/7/docs/api/java/util/Properties.html#load(java.io.InputStream). That would reduce your code footprint a lot.

johnnynotsolucky:
Haha yes a load *and* a store method! Thank you, I didn't even think about that

johnnynotsolucky:
Alright, I've implemented the same thing using java.utils.Properties, much less code and the logic is simpler too.

neilboyd:
Link to issue #213 as an alternative

spyhunter99:
I'm going to go ahead and close this PR, pending the merge of the sqlite based solution #213. I appreciate your hard work that went into it. If it's any consolidation, most of your code was moved over to the sqlite based solution.

PR #215, opened 2015-12-14
OSMapViewer Code Cleanup - Permissions and other fixes.

ZKjellberg:
Went through many files trying to clean up readability.

For SampleLoader, I cleaned up the permission system as it was requesting many permissions which are automatically granted. The project only requires two permission groups Location & Storage, helping simplify the sample code. In addition, I grouped AndroidManifest permissions into groups for easy understanding of permissions and expectations.

Updated deprecated code in MapActivity to be used directly inside its Fragment. (Dialog)
Updated build.gradle to match AndroidManifest.xml versioning.
Minor formatting cleanup in various files.

### DIFF
--- a/OpenStreetMapViewer/build.gradle
+++ b/OpenStreetMapViewer/build.gradle
@@ -8,8 +8,8 @@ android {
         applicationId "org.osmdroid.example"
         minSdkVersion 8
         targetSdkVersion 23
-        versionCode 16
-        versionName "5.0-SNAPSHOT"
+        versionCode 18
+        versionName "5.1-SNAPSHOT"
     }
     buildTypes {
         release {
--- a/OpenStreetMapViewer/src/main/AndroidManifest.xml
+++ b/OpenStreetMapViewer/src/main/AndroidManifest.xml
@@ -12,11 +12,14 @@
         android:largeScreens="true"
         android:normalScreens="true" />
 
-    <uses-permission android:name="android.permission.ACCESS_FINE_LOCATION" />
-    <uses-permission android:name="android.permission.ACCESS_COARSE_LOCATION" />
-    <uses-permission android:name="android.permission.ACCESS_WIFI_STATE" />
+    <!-- PROTECTION_NORMAL permissions, automatically granted -->
     <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
+    <uses-permission android:name="android.permission.ACCESS_WIFI_STATE" />
     <uses-permission android:name="android.permission.INTERNET" />
+
+    <!-- DANGEROUS PERMISSIONS, must request -->
+    <uses-permission android:name="android.permission.ACCESS_COARSE_LOCATION" />
+    <uses-permission android:name="android.permission.ACCESS_FINE_LOCATION" />
     <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE" />
 
     <uses-feature
--- a/OpenStreetMapViewer/src/main/java/org/osmdroid/MapActivity.java
+++ b/OpenStreetMapViewer/src/main/java/org/osmdroid/MapActivity.java
@@ -1,9 +1,6 @@
 // Created by plusminus on 00:23:14 - 03.10.2008
 package org.osmdroid;
 
-import android.app.AlertDialog;
-import android.app.Dialog;
-import android.content.DialogInterface;
 import android.os.Bundle;
 import android.support.v4.app.FragmentActivity;
 import android.support.v4.app.FragmentManager;
@@ -14,52 +11,21 @@ import android.support.v4.app.FragmentManager;
  * @author Manuel Stahl
  *
  */
-public class MapActivity extends FragmentActivity
-{
-
-    private static final int DIALOG_ABOUT_ID = 1;
-	private static final String MAP_FRAGMENT_TAG = "org.osmdroid.MAP_FRAGMENT_TAG";
+public class MapActivity extends FragmentActivity {
+    private static final String MAP_FRAGMENT_TAG = "org.osmdroid.MAP_FRAGMENT_TAG";
 
     // ===========================================================
     // Constructors
     // ===========================================================
     /** Called when the activity is first created. */
     @Override
-    public void onCreate(final Bundle savedInstanceState)
-    {
+    public void onCreate(final Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
-
         this.setContentView(org.osmdroid.R.layout.main);
-
         FragmentManager fm = this.getSupportFragmentManager();
-
 		if (fm.findFragmentByTag(MAP_FRAGMENT_TAG) == null) {
 			MapFragment mapFragment = MapFragment.newInstance();
 			fm.beginTransaction().add(org.osmdroid.R.id.map_container, mapFragment, MAP_FRAGMENT_TAG).commit();
 		}
-    }
-
-    @Override
-    protected Dialog onCreateDialog(final int id)
-    {
-        Dialog dialog;
-
-        switch (id) {
-            case DIALOG_ABOUT_ID:
-                return new AlertDialog.Builder(MapActivity.this).setIcon(org.osmdroid.R.drawable.icon)
-                        .setTitle(org.osmdroid.R.string.app_name).setMessage(org.osmdroid.R.string.about_message)
-                        .setPositiveButton(android.R.string.ok, new DialogInterface.OnClickListener() {
-                            @Override
-                            public void onClick(final DialogInterface dialog, final int whichButton)
-                            {
-                                //
-                            }
-                        }).create();
-
-            default:
-                dialog = null;
-                break;
-        }
-        return dialog;
     }
 }
--- a/OpenStreetMapViewer/src/main/java/org/osmdroid/MapFragment.java
+++ b/OpenStreetMapViewer/src/main/java/org/osmdroid/MapFragment.java
@@ -1,22 +1,10 @@
 // Created by plusminus on 00:23:14 - 03.10.2008
 package org.osmdroid;
 
-import org.osmdroid.constants.OpenStreetMapConstants;
-import org.osmdroid.samplefragments.BaseSampleFragment;
-import org.osmdroid.samplefragments.SampleFactory;
-import org.osmdroid.tileprovider.tilesource.ITileSource;
-import org.osmdroid.tileprovider.tilesource.TileSourceFactory;
-import org.osmdroid.views.MapView;
-import org.osmdroid.views.overlay.MinimapOverlay;
-import org.osmdroid.views.overlay.ScaleBarOverlay;
-import org.osmdroid.views.overlay.compass.CompassOverlay;
-import org.osmdroid.views.overlay.compass.InternalCompassOrientationProvider;
-import org.osmdroid.views.overlay.gestures.RotationGestureOverlay;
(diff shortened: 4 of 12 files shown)


ZKjellberg:
I wanted to help cleanup the MapActivity as well, but then realized how it is depended upon for demonstration of SampleFragments. This is a bit strange of an approach as it is difficult for people to discover this hidden Menu option for additional samples. A better implementation could be another entry on the launch Activity which would open a Fragment with a ListView of these samples and inflate them?

Additionally, I think SampleLoader should be renamed similar to MainActivity to highlight its importance as the launch activity as many will assume MapActivity is the starter due to its presence on the root and its name being the only class with Activity. I assume these problems are from a historical evolution, but think they should be addressed as developers exploring the code will get easily confused or simply be unaware of the many interesting samples available.

spyhunter99:
I feel the same about map activities options menu for additional samples. Previous versions went right to map activity but then you couldn't hit the opening menu for the other stuff. It is a bit strange. Feel free to move stuff around into a more logical organization (code and presentation)

spyhunter99:
Doesn't compile....<> diamond operator

ZKjellberg:
The diamond operator is an Java 1.7 language feature, if you are compiling inside Android Studio it should be fine. Since Android Lint suggests adopting the diamond operator for the assumed type, I often follow their guidance.

What environment are you building in which is causing the issue? I assume maven might not adopt the some of the newer build features and cause a conflict. No issue with restoring this as you have already done. I will avoid implementing future diamond operators.

I believe this error is caused because the root pom.xml has  the following:

´´´xml
<plugin>
  <groupId>org.apache.maven.plugins</groupId>
  <artifactId>maven-compiler-plugin</artifactId>
  <version>2.3.2</version>
  <configuration>
    <source>1.6</source>
    <target>1.6</target>
  </configuration>
</plugin>
´´´

This should target 1.7 to be compatible with the latest Android code as many projects adopt these features, especially with Android Lint suggesting quick fixes using this features. [Stackoverflow Comment](http://stackoverflow.com/questions/29258141/maven-compilation-error-use-source-7-or-higher-to-enable-diamond-operator)

´´´xml
<plugin>
    <groupId>org.apache.maven.plugins</groupId>
    <artifactId>maven-compiler-plugin</artifactId>
    <version>3.1</version>
    <configuration>
        <source>1.7</source>
        <target>1.7</target>
    </configuration>
</plugin>
´´´

(Can I not use XML for code pasting?)

spyhunter99:
osmdroid supports back to api 7 i believe. Does 1.7 java support go back
that far? Android Studio is not a build system. Maven and gradle are. Just
because it builds in android studio, which i believe uses "gradle assemble"
by default, doesn't mean the whole project works. As I found out yesterday
with osmbonuspack, 'gradle clean build' fails due to lint failures and
apparently no one noticed for quite some time. I also 'think' that gradle
assemble skips all unit tests too, which is terrible practice to rely on.

mvn clean install

gradle clean build

On Tue, Dec 15, 2015 at 2:33 AM, Zachary Kjellberg <notifications@github.com

neilboyd:
This answer on [Stack Overflow](http://stackoverflow.com/a/13550632/260221) is not really the same thing, but the idea that you can use Java 7 *language features* seems reasonable, ie features that are just syntactic sugar, including the diamond operator. I guess Lint would tell you if you used a non-supported feature.
I couldn't find any better reference, eg whether Java 7 will work on an API 8 device. The only way to find out is to try it.
I tend to agree to go with what Android Studio suggests.

ZKjellberg:
I should change my wording to reference gradle. Since Android Studio defaults to gradle configurations, I have a habit of referencing it since that is the default workflow. I definitely appreciate the caution with validating all build methods still work.

Just went through process to set up my environment for gradle from command line (using gradlew and had to add JAVA_HOME to path) and downloaded maven to test the clean install. **gradle clean build** compiles with the Java 1.7 language.

When I attempt to use **mvn clean install**, I updated the root pom.xml with the above code to allow Java 1.7, but I am receiving the following error:

```
[ERROR] Failed to execute goal on project OpenStreetMapViewer: Could not resolve dependencies for project org.osmdroid:OpenStreetMapViewer:apk:5.1-SNAPSHOT: Failed to collect dependencies at android:android:jar:[(6.0_r1,): No versions available for android:android:jar:[(6.0_r1,) within specified range -> [Help 1]
```
I am trying to follow the [wiki entry for setting up Maven](https://github.com/osmdroid/osmdroid/wiki/How-to-build-OsmDroid-from-source) and I am confused how to use the [Maven Android SDK Deployer](https://github.com/mosabua/maven-android-sdk-deployer). I set my ANDROID_HOME to the SDK Android Studio downloaded. I am not sure how I should be using this repository, is it a modified Android SDK?

Once I can resolve this build system, I will test compatibility with Android 1.7 language with maven. Again, this is not a required change as it is very minor, but I'd like to highlight that these are automatic fixes from the Android Studio Inspection which means they may return for future commits by other users.

spyhunter99:
I bet that if we update the maven pom file to use 1.7 source, this would
probably be resolved.

For the maven setup, Google doesn't allow publishing any of the android sdk
artifacts to a centralized repo (like maven central, jcenter, etc),
effectively preventing maven users from doing normal things. The android
sdk deployer basically repackages everything, then deploys to maven local.
In some cases, it compiles android libraries from source, other files are
just copied and pom'd. On a mac, set the ANDROID_HOME =
/Users/{USERNAME}/Library/Android/sdk. Windows and others, it depends on
how you installed android studio (installer vs zip and the android sdk
itself). Does this help?

On Tue, Dec 15, 2015 at 8:54 AM, Zachary Kjellberg <notifications@github.com

ZKjellberg:
@spyhunter99 I set ANDROID_HOME to target my SDK folder on my Windows dev machine as you stated. When I call $ANDROID_HOME, it recognizes the path, yet **mvn** still fails.

Do I actually need the Maven Android SDK Deployer repo or should simply adding the path for my local Android SDK be enough? I'll troubleshoot this tomorrow further, I am no longer at my office to test.

spyhunter99:
Yes you need the maven android sdk deploy repo. git clone it

ZKjellberg:
To clarify, I should clone the Maven Android SDK Deployer repo and set it as my ANDROID_HOME variable? This seems strange since it is just POMs and would break any other reference on the machine which might use that variable for access to tools such as ADB. I will test tomorrow.

spyhunter99:
no.

ANDROID_HOME = (whereever the android sdk is on your machine)

android sdk deployer, just git clone it, then run mvn install -fn
the clone location doesn't matter

ZKjellberg:
Working through getting this maven build to work. Thanks for the explanation. I downloaded the **Maven Android SDK Deployer** and ran the suggested command **mvn install -fn**. I was having issues as I never installed the **Google Play Services** package.

After installing the missing package and rerunning **mvn install -fn**, I was able to run mostly successfully run a **mvn clean install** for osmdroid. All projects except for **OSMdroid 3rd party examples** compiled successfully.

See here for the [gist output of my mvn clean install](https://gist.github.com/ZKjellberg/e67c7e438a5fe979cb92)

See below for an exerpt of the primary errors:

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.1:compile (default-compile) on project GoogleWrapperSample: Compilation failure: Compilation failure:
[ERROR] /C:/Users/KjellbergZ/AndroidStudioProjects/osmdroid/GoogleWrapperSample/src/main/java/org/osmdroid/google/sample/GoogleWrapperSample.java:[12,38] package org.osmdroid.google.wrapper.v2 does not exist
[ERROR] /C:/Users/KjellbergZ/AndroidStudioProjects/osmdroid/GoogleWrapperSample/src/main/java/org/osmdroid/google/sample/GoogleWrapperSample.java:[13,49] package org.osmdroid.tileprovider.tilesource.bing does not exist
[ERROR] /C:/Users/KjellbergZ/AndroidStudioProjects/osmdroid/GoogleWrapperSample/src/main/java/org/osmdroid/google/sample/GoogleWrapperSample.java:[26,31] cannot find symbol
[ERROR] symbol:   class Constants
[ERROR] location: package org.osmdroid.thirdparty
[ERROR] /C:/Users/KjellbergZ/AndroidStudioProjects/osmdroid/GoogleWrapperSample/src/main/java/org/osmdroid/google/sample/GoogleWrapperSample.java:[124,47] cannot find symbol
[ERROR] symbol:   variable MapFactory
```

neilboyd:
I still have trouble getting to build faultlessly with mvn, but usually if you run the command again without clean then it works second or third time

ZKjellberg:
Near the end of this issue discussion. Thanks again for helping walk through the issues. I'd like to fix up the Maven Build Wiki as this was a very confusing process. I figure rewording it slightly should make it a little easier for others.

Running **mvn install** without clean did successfully compile the **OSMdroid 3rd party examples** but fails when running the **osmdroid-android-it - Integration tests**. Since it is a permission denied error, it might be an Android 6.0 permission issue, so it may just be a quick fix to add the permission request. 

```
[INFO] 02c6381fd0218c8f_LGE_Nexus5 :       Start [19/20]: org.osmdroid.views.util.OpenStreetMapTileProviderDirectTest#test_getMapTile_found
[INFO] 02c6381fd0218c8f_LGE_Nexus5 :       org.osmdroid.views.util.OpenStreetMapTileProviderDirectTest#test_getMapTile_found
[INFO] 02c6381fd0218c8f_LGE_Nexus5 :       java.io.FileNotFoundException: /sdcard/osmdroid/OpenStreetMapTileProviderTest.png: open failed: EACCES (Permission denied)
at libcore.io.IoBridge.open(IoBridge.java:452)
at java.io.FileOutputStream.<init>(FileOutputStream.java:87)
at java.io.FileOutputStream.<init>(FileOutputStream.java:127)
at java.io.FileOutputStream.<init>(FileOutputStream.java:116)
at org.osmdroid.views.util.OpenStreetMapTileProviderDirectTest.test_getMapTile_found(OpenStreetMapTileProviderDirectTest.java:57)
at android.test.AndroidTestRunner.runTest(AndroidTestRunner.java:191)
at android.test.AndroidTestRunner.runTest(AndroidTestRunner.java:176)
at android.test.InstrumentationTestRunner.onStart(InstrumentationTestRunner.java:555)
at android.app.Instrumentation$InstrumentationThread.run(Instrumentation.java:1879)
Caused by: android.system.ErrnoException: open failed: EACCES (Permission denied)
at libcore.io.Posix.open(Native Method)
at libcore.io.BlockGuardOs.open(BlockGuardOs.java:186)
at libcore.io.IoBridge.open(IoBridge.java:438)
... 15 more

[INFO] 02c6381fd0218c8f_LGE_Nexus5 :       End [19/20]: org.osmdroid.views.util.OpenStreetMapTileProviderDirectTest#test_getMapTile_found
```

ZKjellberg:
Reworked some of the wording on the [Build with Maven](https://github.com/osmdroid/osmdroid/wiki/How-to-build-OsmDroid-from-source) wiki page, hopefully making it easier for future users attempting to mess with the command line.

For the above error in org.osmdroid.views.util.OpenStreetMapTileProviderDirectTest has the following code:
```
Line 52: final String path = "/sdcard/osmdroid/OpenStreetMapTileProviderTest.png";
Line 57: final FileOutputStream fos = new FileOutputStream(path);
```

My device does not have the desired image **OpenStreetMapTileProviderTest.png**. How would this be generated to pass this test?

PR #228, opened 2016-01-04
New Permission Check implementation

ZKjellberg:
The former permission check implementation only checked for a single permission for each session. This new implementation will request both permissions on first launch. Additionally, I improved the workflow for a better user experience with an initial Toast to explain why they are being prompted for the permission.

Commit is a bit confusing. I moved the ListView choice selection to the top, and rewrote both permission methods from scratch.

[Travis CI states the build is passing](https://travis-ci.org/ZKjellberg/osmdroid/builds/100106300)


spyhunter99:
despite the travis failures, i'm going to merge it. the failures are not code or test related.

PR #247, opened 2016-01-19
using batchmode and adding updated flag for Maven builds

mosabua:
I tried building locally and I did not get the compile problems you see on travis. So at least theoretically this should work

-B removes the log lines for KB download stuff
-U forces to check in upstream repo for artifacts

hth

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -59,7 +59,7 @@ before_script:
   - cd temp
   - git clone https://github.com/simpligility/maven-android-sdk-deployer
   - cd maven-android-sdk-deployer
-  - mvn install -fn
+  - mvn clean install -fn -B -U
   - cd ..
   - cd ..
 
@@ -71,9 +71,9 @@ script:
   #build using maven with integration tests
   #we build 3 times due to issues with the maven-android-plugin. root cause is unknown but seems to be related to how artifacts are resolved for android projects
   #in case you're wondering -fn means fail never, which loosely means, if the build fails, the exit code is always 0, which is basically all CI engines look for.
-  - mvn clean install -fn
-  - mvn install -fn
-  - mvn install 
+  - mvn clean install -fn -B -U
+#  - mvn install -fn
+#  - mvn install 
   - mvn android:undeploy
   #build using gradle
   - gradle -version


spyhunter99:
sweet! much appreciated

neilboyd:
When I look at the Travis log it's full of errors, but they're just ignored because of -fn
Or am I missing something?

mosabua:
Which log are you looking at specifically? And yes -fn ignores errors.

mosabua:
Also... you have to merge and try and see if it works.

mosabua:
Haha .. I just noticed you have branch build with travis and this branch passed.. and I did take out the repeated maven invocations. Let me check more in the log.

mosabua:
Looking at the travis log I still see compile problems. I do NOT get them locally.

mosabua:
You should see if you can manually run the SDK deployer locally and then an osmdroid build. That works for me.

neilboyd:
I've been trying it a few times, and it seems to sometimes work and sometimes not.  I'm not sure what's different between the times it works and the times it doesn't.
But surely I shouldn't need to clean maven-android-sdk-deployer every time?
And shouldn't osmdroid work without `-fn -B -U`? Or at least without `-fn`?

mosabua:
Well.. thats what I am saying. It works without -fn for me just fine. Reproducibly..

spyhunter99:
the current travis file, clones the maven-sdk-deployer, then install -fn. Then builds osmdroid.

On my box, if i wipe out the .m2 repo, the problem is reproduced. Travis builds are essentially new VMs with every build and thus an empty local .m2 repo.

mosabua:
Let me try with a clean setup ..

neilboyd:
Thank you for helping Manfred, I very much appreciate it :)

Please could you explain your thinking behind these changes that you made to .travis.yml:

For maven-android-sdk-deployer:
Why clean?  On the Travis server it's already a new box, and locally it takes forever.
Why -U?

For osmdroid:
Why -fn?  We want it to just pass without errors.
Why -U?  This causes it to get other osmdroid dependencies from Maven Central, whereas we should be building against our local version.

Sorry if I'm just not understanding Maven properly ;)

neilboyd:
When I said "other osmdroid dependencies", I mean, for example, this:

> [INFO] Downloading: https://oss.sonatype.org/content/repositories/snapshots/org/osmdroid/osmdroid-android/5.1-SNAPSHOT/osmdroid-android-5.1-20160117.193112-7.aar

mosabua:
clean is probably not necessary

-fn not either - it just causes it to keep running even if it fails

download from snapshots.. that looks like a bug in your project setup.

spyhunter99:
yup still an issue.
https://travis-ci.org/osmdroid/osmdroid/builds/103978076 this is without -fn

neilboyd:
Hx'm x

Issue #250, opened 2016-01-21
Memory leak in NetworkAvailabliltyCheck (ConnectivityManager stores reference to activity)


roshakorost:
I am using leak canary (https://github.com/square/leakcanary) to find leaks in my application.
Activities that use MapView have next leak:

![screenshot_20160121-141032](https://cloud.githubusercontent.com/assets/2326845/12480242/8d623de6-c049-11e5-8cb5-3f8718262501.png)

The issue is because 
`NetworkAvailabliltyCheck` used activity context to create `ConnectivityManager` but it supposed to use application context instead. 

as you can see there is next call sequence

````java
protected MapView(final Context context,
					  final ResourceProxy resourceProxy, MapTileProviderBase tileProvider,
					  final Handler tileRequestCompleteHandler, final AttributeSet attrs) {
		super(context, attrs);
		mResourceProxy = resourceProxy;
		this.mController = new MapController(this);
		this.mScroller = new Scroller(context);

		if (tileProvider == null) {
			final ITileSource tileSource = getTileSourceFromAttributes(attrs);
			tileProvider = isInEditMode()
					? new MapTileProviderArray(tileSource, null, new MapTileModuleProviderBase[0])
					: new MapTileProviderBasic(context, tileSource);
		}
    ....
}

public MapTileProviderBasic(final Context pContext, final ITileSource pTileSource) {
		this(new SimpleRegisterReceiver(pContext), new NetworkAvailabliltyCheck(pContext),
				pTileSource, pContext);
}

public NetworkAvailabliltyCheck(final Context aContext) {
		mConnectionManager = (ConnectivityManager) aContext
				.getSystemService(Context.CONNECTIVITY_SERVICE);
          isX86= "Android-x86".equalsIgnoreCase(Build.BRAND);
}
````

I think we should create MapTileProviderBasic with application context to prevent the leak.
so we should create it in next way: 
````java
 new MapTileProviderBasic(context.getApplicationContext(), tileSource);
````

I can make pull request to fix the issue.

spyhunter99:
yes please! never heard of that project but i love it already!

roshakorost:
Btw, as a workaround in client app I added 
`getSystemService(Context.CONNECTIVITY_SERVICE);` line of code in the application class in `onCreate()` method.
It creates singleton instance of `ConnectivityManager` using application context, so when `NetworkAvailabliltyCheck` is initialized it uses `ConnectivityManager` instance that was created in application with correct context.

[...]

PR #251, opened 2016-01-21
#250 fixed memory leak in NetworkAvailabliltyCheck

roshakorost:
`MapTileProviderBasic` is created with application context instead.


spyhunter99:
thanks!

Issue #268, opened 2016-02-15
kill the resource proxy


spyhunter99:
when using tools like osmbonuspack, the creation of markers and various overlays references MapView.resourceProxy. If you're defining the mapview in a layout file, there's no clear way to set the resource proxy class, thus you can run into issues with resources not found. adding a quick method to change it shortly after construction time well help resolve this

MKergall:
No way to get rid of this ResourceProxy? 
Is it still required, now that osmdroid is an aar, including regular Android resources?

neilboyd:
@MKergall No it shouldn't be necessary any more. This was mentioned in #146 but not implemented.

spyhunter99:
oh boy... i smell osmdroid 6.0 is around the corner...

spyhunter99:
So question a few questions on the removal of the resource proxy. Let's take the MyLocation Overlay. It fires up and looks for org.osmdroid.R.person.png and uses that for visualization.

If someone wants to override that, do we then change the constructor or whatever to pass in the resource id or a drawable of what they want icon to be?

MKergall:
Don't change the constructor, but add a setter: MyLocation.setIcon(Drawable). 
Allowing to change icon at runtime. And homogeneous with Marker. 

BTW: MyLocationNewOverlay.setPersonIcon(Bitmap icon) already exists. 

Same approach can be applied to DirectedLocationOverlay icon.

spyhunter99:
Quick update on this, I started a branch for this and although it looks
feasible, although it's a lot of work. Pretty much every constructor and
every file needs to be touched. There's tons of stuff related to options
menus that pulls out various strings and drawables....

My open concern is how would libraries that use osmdroid override some of
the functionality, string values, images, etc

spyhunter99:
ready to merge, pending review

spyhunter99:
merged

PR #271, opened 2016-02-20
#268 kill the resource proxy

spyhunter99:
@neilboyd @MKergall 

here's the initial cut for killing the resource proxy. I think I got everything but there's probably edge cases for overriding default icons


neilboyd:
I had a quick look at the diff and it looks fine.
The only thing I noticed is that in ScaleBarOverlay you used String.format instead of using Resources.getString with format parameters.

PR #347, opened 2016-07-02
Bug/#335 fix for a whole lot of memory leaks

spyhunter99:
This is a large commit (sorry!) that resolves a great deal of memory leaks. It's tested by running a slightly altered version of the instrumentation tests which loops for 100 iterations per sample fragment. If it doesnt crash on api 8 or api 10 devices (which has max heap of 25 MB) then we're in good shape. @MKergall may want to take a look at the changes for FolderOverlay and a few minor changes to Marker.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -109,9 +109,9 @@ before_script:
   - adb shell chmod 777 /mnt/sdcard
   - adb shell mkdir /sdcard/osmdroid/
   - adb push osmdroid-forge-app/world.map /sdcard/osmdroid/world.map
-  - adb push resources/usgsbase.gemf /sdcard/osmdroid/usgsbase.gemf
-  - adb push resources/usgstopo.sqlite /sdcard/osmdroid/usgstopo.sqlite
-  - adb push resources/usgssat.zip /sdcard/osmdroid/usgssat.zip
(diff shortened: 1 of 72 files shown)


MKergall:
Huge work. 
onDetach should also be added to Polygon and Polyline overlays, and to MinimapOverlay.

spyhunter99:
Great catch

spyhunter99:
long story short, nice quite ready to merge yet. I did add the detach methods for polyline and polygon. I also had to modify the IRegisterReceiver class to have a destroy method. I then found another slow leaker that was making the test process unstable over longer durations. I did add leak canary, ACRA and some code to dump hrpof files to find the error. Eventually I found a slow leak using the hprof file and MAT, ArrayAdapter in the FragmentSamples. It may be a bug in the API8 codebase.

spyhunter99:
Turns out it wasn't an android bug, just a matter of poor documentation with list view fragments.

spyhunter99:
MinimapOverlay will be covered by the TilesOverlay that it inherits from. Polyline and polygon were added

spyhunter99:
yay! finally got ci to pass, any more suggestions @MKergall ?

spyhunter99:
whoops, didn't mean to merge this, too late now i guess

PR #357, opened 2016-07-22
Doubles

Travlo26:
Convert Doubles


spyhunter99:
@Travlo26 have you noticed any performance issues or higher than expected memory utilization?

kjeremy:
Yes! This makes me so happy

spyhunter99:
Glad to be of service, I just hope it didn't cause any breaking API
changes. For the most part, it should be just a simple upgrade for everyone

PR #362, opened 2016-08-08
Support implementation of own projection classes

jklimke:
Hi,

we needed for a project a custom projection class for our osmdroid mapview. This was currently not possible since the projection member of the map view was read only for derived classed and the reset of the projection was not done using a setter method (that could be overwritten).

So i did the following
* switched projection to be protected instead of private and added protected setter for projection in order to use own projection class

Also i updated the build tools to the current version of 23.0.3

I think there is no real reason for not providing write access to the projection member for derived classes. 

It would be nice if that could be merged into the main repo


spyhunter99:
looking good. thanks!

PR #373, opened 2016-08-15
Overlay descendants without Context-dependecy that can be created in AsyncTask (again)

k3b:
I have a database with 18000 geo-positions (jpg-photos with exif-geo-info) that are displayed as clusters in a map. 
Every time visible map area or the zoom level changes, the visible clusters are recalculted from the database: 
An AsyncTask creates Clusterpoints (inherited from Overlay) and put them into a FolderOverlay.
When all visible Clusterpoints are calculated in the background, the FolderOverlay is added to the map. 

Overlay and descendants can only be created in background task if they donot acces Android-Context.

Up to osmdroid-5.1 I used ResourceProxy for creating Overlays in AsyncTask. 

Since Osmdroid-5.2 has removed ResourceProxy this merge request also removes Android-Context dependency from some Overlay descendants
so that these can be created in AsyncTask again.

These Overlay descendants are now Android-Context-free so they can be created in AsyncTask (outside main gui thread):

* Overlay, 
* OverlayWithIW, 
* ItemizedOverlay
* Polyline, 
* Polygon, 
* PathOverlay, 
* NonAcceleratedOverlay, 
* MapEventsOverlay, 

I deprecated the original constructors and created a new one without Context.
Old Osmdroid programs will get a deprecated compiler warning but should continue working.

This osmdroid issue relates to https://github.com/k3b/APhotoManager/issues/59


spyhunter99:
wow a lot of work. i had initially considered this approach with the 5.1 to 5.2 refactor. The scale bar was the only thing that uses the one field. I have no issues with the PR other than the scary commit message that says don't merge.

@MKergall does this cause any issues from OBP?

MKergall:
I don't see particular issues from OBP.

PR #504, opened 2016-12-26
Update README.md

adi9090:
There is no .jar file on https://jcenter.bintray.com/org/osmdroid/osmdroid-android/5.6.1/ it is .aar

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ repositories {
 }
 
 dependencies {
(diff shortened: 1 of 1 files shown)


spyhunter99:
when an artifact is published with the pom correctly defined as type=aar, then at least in theory @aar shouldn't be required

spyhunter99:
thanks for the contribution though!

PR #506, opened 2016-12-26
Fixed array out of bounds bug

miguel-porto:
For some reason this bug didn't happen before 5.6. This is just an emergency fix, but I need to better understand what's going on before this line, as this fix should not be needed, given the view bounding box test earlier in the code.


miguel-porto:
This is probably related to #507, there may be some issue in the view bounding box calculation.

Issue #440, opened 2016-10-13
Possible duplicate code in DefaultOverlayManager


cfsbhawkins:
I have been looking over the code in the overlay manager, because I'm looking at optimizing the overlay manager to only render line segments that are on screen as well as any type of overlay, I think someone here was talking about an issue with large polylines and hardware acceleration, I adopted some of their code in a custom folder overlay and it works real well. I was just hoping to intercept the on draw calls in the overlay manager instead but I noticed on the onDraw command it looks like the code is duplicated, any reason for this or was this a merge error?

spyhunter99:
which code?

cfsbhawkins:
If they both are needed to be there Im sure it could be optimized too, why have two for loops, when they could be done in the same loop too. Just thinking if one had a lot of overlays that call might be inefficient. 

I too am curious what the shadow call does.

cfsbhawkins:
If it should be there, though not implemented, I am proposing more of a single for loop rather than two separate loops. Just thinking speed, but if the shadow is useless then maybe remove it. It could always be done on the on draw anyways.

cfsbhawkins:
I will do a test with my changes removing the shadow calls and see if it breaks anything in tests, If it works I will open a pull request for the fix.

spyhunter99:
@cfsbhawkins any updates on this? Getting ready to cut another version soon

cfsbhawkins:
@spyhunter99 I have been testing it let me do one today with a few thousand markers, so far so good. No adverse side affects. If anything were skipping one for loop cycle on every on draw. I had actually been working on a new overlay manager that only draws what is in the current bounding box. I still have a bit of work todo since lines and polygons sometimes do some funny things. I'm hoping that I can get it to the point where we can enable Hardware Acceleration and not have issues with large polylines.

cfsbhawkins:
@spyhunter99 We have been doing some testing here and see no issues with taking out the shadow. That being said I can open a pull request with my changes, but I'm trying to think what we should do. I would love to remove the boolean shadow from the onDraw method, but that would break OSMBonus Pack and anyone who created their own overlays as well, I could create a new method with deprecated tile on the current onDraw or I can just remove the duplicated code from the DefaultOverlayManager and we always pass false to on draw. 

Let me know your thoughts.

spyhunter99:
Your approach sounds perfect. We'll schedule the removal of the additional parameter with 6.0 or maybe later

PR #526, opened 2017-01-09
Possible duplicate code in DefaultOverlayManager #440

cfsbhawkins:
… overlay to be called twice, while code should not have executed on one of them, it caused duplicate loops as discussed in issue #440


spyhunter99:
nice! i was wondering when this was going to be opened

PR #561, opened 2017-02-18
Feature/#560

Haukinger:
This adds a vertical offset for markers while being dragged to prevent them from hiding beneath the dragging finger tip.

The offset defaults to zero, so no visible change unless you want to use the feature.


spyhunter99:
i'll test it out this weekend, PR looks good though

spyhunter99:
@Haukinger  do you by chance have an example on how to use this?

Haukinger:
I create the marker like this:

                var marker = new Marker( _mapView )
                             {
                                 Draggable = true,
                                 Position = _checkpoints[ checkpointIndex ],
                                 DragOffset = 10,
                             };

That's in xamarin, though, but in native java I'd put `marker.setDragOffset(10);`. Anyway, it moves the marker up by 10mm when clicking, and look's like that: [gif](http://imgur.com/a/tl43w)

PR #574, opened 2017-03-29
Bug/#353

shiroyagi:
Here is the change attached to the issue (having made some progress with git).

### DIFF
--- a/osmdroid-android/src/main/java/org/osmdroid/views/overlay/Polyline.java
+++ b/osmdroid-android/src/main/java/org/osmdroid/views/overlay/Polyline.java
@@ -7,13 +7,16 @@ import android.graphics.Paint;
 import android.graphics.Path;
 import android.graphics.Point;
 import android.graphics.Rect;
+import android.util.Log;
 import android.view.MotionEvent;
 
+import org.osmdroid.api.IMapView;
 import org.osmdroid.util.BoundingBox;
 import org.osmdroid.util.GeoPoint;
 import org.osmdroid.util.GeometryMath;
 import org.osmdroid.views.MapView;
 import org.osmdroid.views.Projection;
+import org.osmdroid.views.overlay.OverlayWithIW;
 import org.osmdroid.views.util.constants.MathConstants;
 
 import java.util.ArrayList;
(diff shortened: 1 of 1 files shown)


shiroyagi:
I don't understand the cause of the CI build failures, any ideas?

spyhunter99:
Reason: android and travis. I restarted the jobs. Super excited btw!

shiroyagi:
Sorry, I've just spotted a problem: a76f83d should be undone.  I didn't check closely enough and mRepeatPath is incompatible with the new code.

shiroyagi:
I've reset the last commit and pushed it to github, hopefully that is the right way to do it?

spyhunter99:
testing now...sorry it's taken me so long!

spyhunter99:
@shiroyagi i checked out your branch and put the sample app on my device. Just so i'm understanding the purposes of the PR since i've haven't been following things too closely.

 - turn on hw map acceleration
 - then draw polylines
 - performance should be better
 - lines shouldn't disappear while zooming

My observed behavior is that at higher altitude zoom levels (1), if i draw a bunch of lines, then zoom in to zoom 4, they disappear with `OpenGLRenderer: Shape too large to be rendered into a texture (7005x6805, max=4096x4096)`

I also see lines disappearing at all zoom levels (drawn at zoom=z, disappears at zoom=z+3)

Performance is differently better but...

with hw acceleration off, there's no problem

shiroyagi:
That is odd and almost sounds like you are testing different code: it shouldn't be using a texture any more.  I wonder if different vendors or API version implement drawLines() in a different way?  I haven't been able to replicate the problem but I'm using a different app and probably different hardware.

One strange thing I have noticed when building and testing osmdroid is that I have to run 'gradlew osmdroid-android:install' twice to get a correct build.

shiroyagi:
Some of the code seems to be using the implementation in osmdroid-third-party:
Googlev1WrapperSample.java:import org.osmdroid.api.Polyline;

spyhunter99:
yup i was on the wrong branch, sorry! testing now

spyhunter99:
ok tested, it works great!. I did notice an issue with polygons with hw acceleration turned on, but that's present on both origin and your fork (slow panning at high zoom levels 20+)

shiroyagi:
Great, yes I haven't touched polygons.  MKergall did that separately [here](https://github.com/MKergall/osmbonuspack/blob/master/OSMBonusPackTuto/src/main/java/com/example/osmbonuspacktuto/Polygon.java)

Issue #619, opened 2017-06-16
Enhancement: more flexible CacheManager


monsieurtanuki:
The idea is to give more flexibility to CacheManager.

For the moment the constructor needs a MapView parameter. But:
* not that much from that MapView is needed
* it prevents developers from using CacheManager without a MapView - for instance in an admin dashboard

The idea is to add an alternate constructor that does not need a MapView parameter.

For the moment the features are limited to "performing a task on each tile that match a bounding box / geo points". But I have a simple example where I'm restricted by that behavior:
* I need a not-too-detailed map of Paris
* I also need a detailed map of a district of Paris
* And some day, eventually, I don't need my detailed map on a subset of Paris but I want to keep my big map of Paris

With the current CacheManager, I'm not able to get rid of the district tiles without potentially getting rid of some of the city tiles.
The idea is to develop new constructors for CacheManagerTask, with List<MapTile>s as an input parameter, and to create new methods (based on already existing ones) that return the MapTile's that match a bounding box / geo points.
With that I could easily say:
1. give me the list of the tiles I want to get rid of (bounding box of the district on some zoom levels)
2. give me the list of the tiles I want to keep (bounding box of the city on sometimes the same zoom levels)
3. give me the list of tiles of list 1 that are not in list 2 (using List related methods)
4. perform a CleaningTask on that list

In the process, there will be slight code refactoring, including removing code duplicates (counting the total tiles vs performing a task on each tile).

MKergall:
I think it makes sense. 
Maybe this could also clean the mess around loading based on "a list of geopoints" (pushing tiles list building in a dedicated method). 

I guess @spyhunter99 answer : propose a PR ...

spyhunter99:
100% open to change. Open a PR. This is a community driven product, not a business driven one (to my knowledge, there is no corporate or government sponsorship or funding for osmdroid).

PR #620, opened 2017-06-18
feature/#619 more flexible CacheManager

monsieurtanuki:
* More flexible CacheManager
* Slight impact of CacheManager refactoring on the other 2 classes
* Slight correction on the manifest not related to #619 but needed when testing

### DIFF
--- a/osmdroid-android/src/main/java/org/osmdroid/tileprovider/cachemanager/CacheManager.java
+++ b/osmdroid-android/src/main/java/org/osmdroid/tileprovider/cachemanager/CacheManager.java
@@ -36,9 +36,11 @@ import java.net.HttpURLConnection;
 import java.net.URL;
 import java.net.UnknownHostException;
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.Date;
 import java.util.HashSet;
 import java.util.Iterator;
+import java.util.List;
 import java.util.Set;
 
 ;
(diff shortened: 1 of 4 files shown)


secaw:
This is such a great improvement - now we can also download tile overlays like OpenSeaMap together with the base map.
Tested this & works.
Thank You!

osmdroid-android/src/main/java/org/osmdroid/tileprovider/cachemanager/CacheManager.java
```diff
@@ -61,24 +63,31 @@
...
     }
 
     public CacheManager(final MapView mapView, IFilesystemCache writer) {
-        mTileProvider = mapView.getTileProvider();
-        mTileWriter = writer;
-        mMapView = mapView;
+        this(mapView.getTileProvider(), writer, mapView.getMinZoomLevel(), mapView.getMaxZoomLevel());
+    }
+
+    public CacheManager(final MapTileProviderBase pTileProvider,
```
spyhunter99: since this is a new constructor, please add a javadoc comment that references the issue for the PR and an @ since 5.6.5

spyhunter99:
@monsieurtanuki mentioned in this issue, but not in the code (?) was the addition of a new constructor for Cache Manager. Was that added or am i just missing it, or is this a 'push to another issue' task?

other than this and the minor javadoc issue, i don't see any obvious problems

monsieurtanuki:
@spyhunter99 I just added a slight Javadoc to the new `CacheManager` constructor. In order to clear the confusion, I'm talking about the constructor referred to in the issue description and added in the previous code push.

PR #627, opened 2017-07-04
Feature/#624

monsieurtanuki:
Added method IFilesystemCache.getExpirationTimestamp: gets the cache expiration timestamp of a tile in milliseconds, or null if expiration timestamp is not supported or if the tile is not cached

Added CacheManager.isTileToBeDownloaded: answers the question "Should we download this tile?", either because it's not cached yet or because it's expired

Added CacheManager.forceLoadTile: actual tile download, regardless of the tile being already present in the cache


osmdroid-android/src/main/java/org/osmdroid/tileprovider/modules/IFilesystemCache.java
```diff
@@ -52,4 +52,15 @@ boolean saveFile(final ITileSource pTileSourceInfo, MapTile pTile,
...
      */
 	boolean remove(ITileSource tileSource, MapTile tile);
+
+	/**
+	 * Gets the cache expiration timestamp of a tile
+	 *
+	 * @since 5.6.5
+	 * @param pTileSource
+	 * @param pTile
+	 * @return cache expiration timestamp in milliseconds,
```
spyhunter99: in time since UTC epoch

monsieurtanuki:
feature/#630
* ExpirableBitmapDrawable: created two new states: SCALED and NOT_FOUND
* SqlTileWriter: made static the getIndex methods (to be used in MapTileSqlCacheProvider), uniformized the primary key use
* MapTileSqlCacheProvider: removed useless mMaximumCachedFileAge (in db we already store the expiration timestamp, not the last modified timestamp), fixed method hasTile, used * SqlTileWriter.getIndex and getTileCursor
* MapTileModuleProviderBase: used the new ExpirableBitmapDrawable.getState method
* MapTileProviderArray: used the new ExpirableBitmapDrawable.getState method
* MapTileProviderBase: stored the scaled tiles asap
* DefaultConfigurationProvider: raised cacheMapTileCount from 9 to 24

spyhunter99:
not sure how there's a merge conflict...

monsieurtanuki:
Indeed, this "9" becomes "24" looked rather innocent...

How could we fix that? Is it possible for you to just deliver a simple modification of "DefaultConfigurationProvider" with "24" instead of "9"?

monsieurtanuki:
Sorry for the confusion with branches, I'm a Github rookie. I still don't understand the roots of the conflicts with `SqlTileWriter`

This is a delivery for new features, that entails the following impact:

feature/#624
* IFilesystemCache: added method getExpirationTimestamp: gets the cache expiration timestamp of a tile in milliseconds, or null if expiration timestamp is not supported or if the tile is not cached
* CacheManager: added method isTileToBeDownloaded (answers the question "Should we download this tile?", either because it's not cached yet or because it's expired), added method forceLoadTile (actual tile download, regardless of the tile being already present in the cache)

feature/#630
* ExpirableBitmapDrawable: created two new states: SCALED and NOT_FOUND
* SqlTileWriter: made static the getIndex methods (to be used in MapTileSqlCacheProvider), uniformized the primary key use
* MapTileSqlCacheProvider: removed useless mMaximumCachedFileAge (in db we already store the expiration timestamp, not the last modified timestamp), fixed method hasTile, used SqlTileWriter.getIndex and getTileCursor
* MapTileModuleProviderBase: used the new ExpirableBitmapDrawable.getState method
* MapTileProviderArray: used the new ExpirableBitmapDrawable.getState method
* MapTileProviderBase: stored the scaled tiles asap

feature/#634
* new class MapTileApproximater: a tile provider that computes approximation of tiles based on the tiles of the same region, but on lower zoom level tiles. Useful in offline mode. To be used as a nth provider.
* CacheAdapter: small bug fix
* IFilesystemCache: added method loadTile
* MapTileFilesystemProvider: added a TileWriter member, used TileWriter.loadTile method int TileLoader.loadTile
* MapTileModuleProviderBase: added method tileLoadedScaled, fixed methods tileLoaded* with explicit calls to ExpirableBitmapDrawable.setState
* MapTileProviderBase: moved method getBitmap to MapTileApproximater, used new method MapTileApproximater.approximateTileFromLowerZoom in ZoomInLooper.handleTile
* MapTileProviderBasic: included MapTileApproximater to the provider array
* MapTileSqlCacheProvider: used the SqlTileWriter.loadTile method int TileLoader.loadTile
* SqlArchiveTileWriter: small bug fix, added tool methods for uniform db access, implemented new method loadTile
* SqlTileWriter: added tool methods for uniform db access, implemented new method loadTile
* TileWriter: added method getFile, implemented new method loadTile

spyhunter99:
looking at this now...

Issue #624, opened 2017-06-28
Proposal: Only let tiles from MapTileSqlCacheProvider expire when a network is available


secaw:
Background: users don't want to have their tiles expire, when there's no possibility to reload them.
Offline maps should always stay available independent from expiration time as long as the user is offline,

Possible solution:
New constructors in MapTileSqlCacheProvider with networkAvailabliltyCheck as parameter analog to MapTileDownloader.
Then if tiles are expired, check if network is available before actually setting them as expired.

spyhunter99:
great idea, the only catch is that it would have to be, expire if the currently selected tile source is network reachable. If you're connected, it doesn't necessarily mean that the tile source is available.

monsieurtanuki:
@spyhunter99: I have a suggestion connected to this one.

In `CacheManager`, add the concepts of
- `isTileToBeDownloaded` - a mix of "tile that's never been downloaded" and "tile that has already been downloaded but is now expired and potentially to be deleted by the next clean up operation"
- `forceLoadTile` - a method that loads a tile, regardless of its presence in the `TileWriter`. Merely a snippet of the current `loadTile` method
With those guys I'm now able to protect the `MapTile` region I cached in the first place.

Does that make sense? If so, I'm ready to develop and deliver by the end of the week.

secaw:
@spyhunter99 : Good point, I'd have to dig a bit deeper, don't quite understand yet how the map tile loading chain including managing expired tiles exactly works...

@monsieurtanuki : Sounds interesting to me. In regards to the CacheManager I would also have two suggestions:
- Add a callback to cleanAreaAsync analog to downloadAreaAsync, so that e.g. we can clean different sources one by one
- Add a callback method 'onTaskCancelled' to CacheManagerCallback, so that e.g. we are able to react on progress dialog dismissals

Should we open another issue for the cache manager topics?

monsieurtanuki:
@secaw:
1. clean up code is in `SqlTileWriter.runCleanupOperation`
2. callbacks can already be added, in branch `osmdroid-parent-5.6.5` (cf. feature/#619)
3. I believe you can already check `CacheManagerTask.onCancelled`, but I'm not sure that answers your `onTaskCancelled` need

Regarding answers 2 and 3, developers only need to code about 3 lines; something like
`final CacheManagerTask task = new CacheManagerTask(...);
task.addCallback(...);
execute(task);`

spyhunter99:
@monsieurtanuki fire at will

[...]

PR #631, opened 2017-07-15
Bug/#624

shiroyagi:
Change SqlTileWriter's use of getTileFileSystemCacheTrimBytes() to hopefully what was intended.  Use average tile size rather than 4kB estimate to avoid purging too much of the cache.  See comments on #624.


shiroyagi:
Also added some info functions to SqlTileWriter to get database size and the age of the tile that expires first.  I find these useful for keeping an eye on the cache and checking it behaves as expected.

PR #651, opened 2017-08-04
Feature/#640

spyhunter99:


monsieurtanuki:
Checking why it failed...

monsieurtanuki:
It looks like there were problems with `Bug445Caching`.
I then decided to code a more verbose and more accurate version.
Tried to push it, but it went into the twilight zone.
If ever I cannot find a way to put it back in the normal process, here's the code
[Bug445Caching.java.txt](https://github.com/osmdroid/osmdroid/files/1199620/Bug445Caching.java.txt)

(Yes I know I need to learn a lot about git)

PR #657, opened 2017-08-10
Update CacheManager.java - Add Alert-Confirmation before canceling

iwanmcm:
In my opinion the user should be asked, if he wants to cancel the map download. Since it may be the case that the cancel action was not intended.
I added an option "verifyCancel" which can be set to true or false. On true it will display an alert asking for confirmation.


spyhunter99:
all tests have failed, please revisit the test failures

spyhunter99:
lots of GC happening. looking at this now

spyhunter99:
see https://github.com/osmdroid/osmdroid/pull/664 closing this PR

PR #666, opened 2017-08-13
feature/#646: MapTileApproximater with several sources

monsieurtanuki:
`MapTileApproximater` now accepts not only one but several providers. They are added using new method `addProvider(final MapTileModuleProviderBase pProvider)`.
When `MapTileApproximater.getTileLoader().loadTile(MapTile mapTile)` is called, we return an approximation from the best possible bitmap, which comes from the best possible zoom level among all providers, and from the first provider on the ordered providers list for that best zoom level.

Files impacted:
* `MapTileApproximater`: changed the constructors, added method `addProvider(final MapTileModuleProviderBase pProvider)`, refactored `approximateTileFromLowerZoom`methods
* `MapTileModuleProviderBase`: `TileLoader` is more `public`, `getTileLoader()` is upgraded from `Runnable` to `TileLoader`, new method `loadTile(final MapTile pTile)` deprecates `loadTile(MapTileRequestState pState)`, with an obvious impact to the classes that extend `MapTileModuleProviderBase`
o `GeoPackageMapTileModuleProvider`
o `MapsForgeTileModuleProvider`
o `MapTileAssetsProvider`
o `MapTileDownloader`
o `MapTileFileArchiveProvider`
o `MapTileFilesystemProvider`
o `MapTileProviderTest`
o `MapTileSqlCacheProvider`
* `MapView`: unrelated Android Studio warning fix
* `MapTileProviderBase`: changed use of `MapTileApproximater.approximateTileFromLowerZoom`
* `MapTileProviderBasic`: changed use of `MapTileApproximater`

### DIFF
--- a/osmdroid-android/src/main/java/org/osmdroid/tileprovider/modules/MapTileApproximater.java
+++ b/osmdroid-android/src/main/java/org/osmdroid/tileprovider/modules/MapTileApproximater.java
@@ -29,7 +29,6 @@ public class MapTileApproximater extends MapTileModuleProviderBase {
 
     private final List<MapTileModuleProviderBase> mProviders = new ArrayList<>();
     private int minZoomLevel;
-    private int maxZoomLevel;
 
     /**
      * @since 5.6.6
@@ -57,16 +56,14 @@ public class MapTileApproximater extends MapTileModuleProviderBase {
 
     private void computeZoomLevels() {
         boolean first = true;
+        minZoomLevel = OpenStreetMapTileProviderConstants.MINIMUM_ZOOMLEVEL;
         for (final MapTileModuleProviderBase provider : mProviders) {
             final int otherMin = provider.getMinimumZoomLevel();;
-            final int otherMax = provider.getMaximumZoomLevel();;
             if (first) {
                 first = false;
                 minZoomLevel = otherMin;
-                maxZoomLevel = otherMax;
             } else {
                 minZoomLevel = Math.min(minZoomLevel, otherMin);
-                maxZoomLevel = Math.max(maxZoomLevel, otherMax);
             }
         }
     }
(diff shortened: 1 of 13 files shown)


osmdroid-android/src/main/java/org/osmdroid/tileprovider/modules/MapTileApproximater.java
```diff
@@ -59,60 +87,70 @@ protected String getThreadGroupName() {
...
             }
-            final Bitmap bitmap = approximateTileFromLowerZoom(tileSource, mReader, pState.getMapTile());
-            if (bitmap == null) {
-                return null;
+            return null;
+        }
+    }
+
+    /**
+     * Approximate a tile from a lower zoom level
```
spyhunter99: do you mean lower numerical (slippy zoom level) or altitude?

monsieurtanuki: slippy zoom level

spyhunter99:
can you add a sample fragment that shows how to use this or is this enabled by default?

monsieurtanuki:
I guess one could say it is enabled by default, as `MapTileApproximater` is set up and used inside `MapTileProviderBasic`.

spyhunter99:
@monsieurtanuki gotcha. Do we have a procedure to test this? I know the issue discusses that this is most relevant when offline, but is offline a requisite for testing this?

monsieurtanuki:
@spyhunter99 The obvious test situation is when offline. There's no reason why it shouldn't work online, except that the download provider will be triggered before.

Oh, if you're online but your download provider's max zoom level is 19, then the MapTileApproximater will be used for 20+ zoom levels.
But I guess I have a little bug to fix then, as I use the other providers to compute MapTileApproximater's max zoom level. And if I compute 19 as the max zoom level for MapTileApproximater, it won't be used for levels 20+. I'll fix that tomorrow.

spyhunter99:
Might want to just set it to 22

spyhunter99:
omg i love it!

monsieurtanuki:
Glad you like it!

PR #672, opened 2017-08-18
Feature/#670

monsieurtanuki:


spyhunter99:
for the record, i love reading your commit messages. most detailed commit messages, ever. shouldn't be a problem merging this however this commit pr adds a number of api changes. It's probably a low probability that changing the internals of the getTileLoader will have any impact on users. I image that few users of osmdroid extend or modify something this low level but it's something for future consideration.

PR #686, opened 2017-08-26
Feature/#665 support for geopackage feature tiles and conversion to osmdroid overlays

spyhunter99:


spyhunter99:
ehh should probably fix the fixme's

PR #695, opened 2017-08-29
add constructor InfoWindow(View, Mapview)

sonwh98:
Adding this constructor allows views added programmatically instead of declared in an xml layout. I needed this constructor so that i can create a webview in an InfoWindow without using xml


sonwh98:
The test is aborted because it timed out. Adding this constructor should be unrelated to the test failing

Issue #476, opened 2016-11-18
Polygon drawing issue


hlaurent:
I want to draw large polygons over my map.
I am using a fork of the Polygon.java class keeping the exact same drawing logic.

When I pan the map, the polygon might not be drawn correctly as it looks like some segments are drawn going around the earth.

Not sure if I am clear so I add a set of screenshots beginning and finishing with the correct drawings.
![device-2016-11-18-115339](https://cloud.githubusercontent.com/assets/467298/20427872/5e7b968c-ad86-11e6-9dbf-03591360db20.png)
![device-2016-11-18-115358](https://cloud.githubusercontent.com/assets/467298/20427873/5e7ce334-ad86-11e6-8efd-b040c318dd18.png)
![device-2016-11-18-115419](https://cloud.githubusercontent.com/assets/467298/20427870/5e751fc8-ad86-11e6-877b-942bee225764.png)
![device-2016-11-18-115429](https://cloud.githubusercontent.com/assets/467298/20427871/5e77eb4a-ad86-11e6-94ae-739566aadd5e.png)
![device-2016-11-18-115440](https://cloud.githubusercontent.com/assets/467298/20427874/5e7e3c48-ad86-11e6-849b-b12448bcb8b4.png)

spyhunter99:
I've seen this before.  I think it's date line related.

MKergall:
Similar issue had been raised (and fixed) on Polyline: https://github.com/MKergall/osmbonuspack/issues/170

Solution had not been applied yet to Polygon.

hlaurent:
Thanks, that solved the issue.

The buildPathPortion method of polygon.java should me modified like this:
```
/**
     * Note - highly optimized to handle long paths, proceed with care.
     * Should be fine up to 10K points.
     */
    protected void buildPathPortion(Projection pj){
        final int size = mConvertedPoints.size();
        if (size < 2) // nothing to paint
            return;

        // precompute new points to the intermediate projection.
        if (!mPrecomputed){
            for (int i=0; i<size; i++) {
                final Point pt = mConvertedPoints.get(i);
                pj.toProjectedPixels(pt.x, pt.y, pt);
            }
            mPrecomputed = true;
        }

        Point projectedPoint0 = mConvertedPoints.get(0); // points from the points list
        Point projectedPoint1;

        Point screenPoint0 = pj.toPixelsFromProjected(projectedPoint0, mTempPoint1); // points on screen
        Point screenPoint1;

[...]

PR #698, opened 2017-09-02
#476 applying deedoc's patch that fixes polygon drawing issues with p…

spyhunter99:
…ole wrapping


spyhunter99:
@MKergall any opinions on this?

MKergall:
I didn't tested yet. Looking at the code: 
General idea seems good (even if comments would be helpful to understand the logic). 
But it's a draft: I'm quite sure that it doesn't handle properly neither holes, nor tap detection. 
And - as already raised - I don't like the principle of vertical replication at all.

spyhunter99:
oook, it's at least an improvement but may not a solution then?

spyhunter99:
@MKergall so i just tested this. Points for holes are reversed, counter clockwise creates the whole. The strange part is that the polygon is double painted and the hole will retain the first paint's fill color. See https://youtu.be/S4Pf2u9WSyQ

On master, its clockwise holes that function normally with a transparent hole. I'm not entirely sure what's going on here. Although this patch improves upon the initial bug, it creates several other bugs. I am therefore hesitant to merge this at this time. What are your thoughts?

PR #702, opened 2017-09-02
Bug/#683

monsieurtanuki:
Fixed the calculation of bounding box zoom.

### DIFF
--- a/osmdroid-android/src/main/java/org/osmdroid/util/BoundingBox.java
+++ b/osmdroid-android/src/main/java/org/osmdroid/util/BoundingBox.java
@@ -66,11 +66,26 @@ public class BoundingBox implements Parcelable, Serializable, MapViewConstants {
 	// ===========================================================
 
 	/**
+	 * Use {@link #getCenterWithDateLine()} instead to take date line into consideration
 	 * @return GeoPoint center of this BoundingBox
 	 */
+	@Deprecated
 	public GeoPoint getCenter() {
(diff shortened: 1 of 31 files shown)


osmdroid-android/src/main/java/org/osmdroid/util/BoundingBox.java
```diff
@@ -66,13 +66,28 @@ public BoundingBox concat(BoundingBox bb2){
...
 	public GeoPoint getCenter() {
 		return new GeoPoint((this.mLatNorth + this.mLatSouth) / 2.0,
 				(this.mLonEast + this.mLonWest) / 2.0);
 	}
 
+	/**
+	 * This version takes into consideration the date line
+	 * @since 5.6.6
+	 */
+	public GeoPoint getCenterWithDateLine() {
```
spyhunter99: why not just put this code in get center?

monsieurtanuki: I didn't want to break anything. I don't know who uses `getCenter`: maybe people now admit that `east` and `west` in `BoundingBox` only mean `longitude1` and `longitude2`, and populate them without any relevant order.

spyhunter99:
hmm ok leave it as is then, makes sense

spyhunter99:
can you merge master into this branch? it should clear up the travis failures

monsieurtanuki:
Crashed again :(

Issue #329, opened 2016-06-06
Support Higher Zoom Level Than 22


schrieveslaach:
At the moment osmdroid is restricted to use a maximum zoom level of 22 even if the tile server provides much higher resolution, for example 26.

The implementation of `TileSystem` restricts the maximum zoom level to 22. Is it possible to raise the maximum to higher numbers? It seems that using higher zoom level than 22 cause integer overflows. 

I started a new fork trying to address this problem but it is not working. I appreciate any help on that.

spyhunter99:
See the doubles branch

spyhunter99:
do you have a publicly accessible source for zoom 22 for testing?

spyhunter99:
I just ran some tests on the doubles branch. Although you can zoom in as far as you want, there's still some overflow happening somewhere. When zooming out, the map is centering somewhere unexpected past zoom 21

spyhunter99:
i just did a quick look into this. It's going to be a lot of work but not impossible.
scrollable area limits will need to be reworked

i also tried to narrow down where the overflow is i think i found it in Project.toMercatorPixels(int x, int y);

Root cause: we're using android's View.scrollTo to handle planning the map around. It uses an x,y coordinate in screen pixels that is integer based.  To overcome the overflow issue, we'd have to write our own mechanism to pan the map that's not based on integers nor based on android's View class. Looks like a ton of work.

schrieveslaach:
@spyhunter99, I ported `TileSystem` to long and I came across the same issues. In a lot of places I needed to convert from long to int... I also tried to `PointF` and `RectF` but I do not have a running version yet. 

What would be good approach to implement this? I'm not very familiar with the internal...

spyhunter99:
This is my current understanding and i'm probably wrong...

Essentially, at every zoom level, it creates a canvas that is equal to the size of the tile in pixels * the total number of tiles on at the zoom level. It doesn't actually allocate the ram for it but that's the size of the canvas. When scrolling (panning on the map), android moves the current view of the canvas around which triggers tile loading, etc. Android uses screen coordinates for everything (integer based). At a high enough zoom level, the conversion from lat/lon to pixel coordinates results in a value > Integer.MAX. The negative value will then send the map view flying off somewhere else. We use Android's scrolling mechanism that requires a Point instance (Integer coordinates). If we can get rid of that and use either doubles or longs everywhere, then the problem would be solved. However, using our own scrolling mechanism is not a trivial task

[...]

PR #722, opened 2017-09-26
Feature/#183-2 Show map only once

Maradox:
`MapView` now has a method `setMapRepetitionEnabled` which allows to configure if the map should be presented only once, or repeat in all directions.

Files impacted:
* `MapView`: For configuration and adaption of `scrollTo`
* `TilesOverlay`: Contains a new variable `wrapEnabled` which is passed by the `MapView`
* `TileLooper`: In case of wrapping disabled, `handleTile` will only be called if the tile is part of the one map
* `TileSystem`: New methods for converting coordinates without wrapping at the borders.

Sample:
* `SampleAssetsOnlyWithoutWrapping`  which can be found in: `More Samples / TileProviders / Assets Only Without Wrapping`


spyhunter99:
@monsieurtanuki you may want to take a look at this.

@Maradox whoa! awesome! can't wait to try it out

monsieurtanuki:
@spyhunter99 You're right, I guess there will be conflicts with my #329 code. I should be able to deliver on Sept. 28th. What would be the best solution to integrate those 2 deliveries on the same code?

spyhunter99:
great work @Maradox! I did find some strange behavior but it is an overall improvement. With the new sample, i can scroll the map off of the screen, however it looks like anything that is considered a fling action will snap the map back on screen. What do you think about forcing the map to always stay on screen?

I also modified the drawing samples to prevent the map wrapping and ran into an issue where by the line and polygons are still repeated (meaning they are painted on the now black surface when the draw area for the line/polygon is off screen). Similar issue with polygons. Maybe @MKergall knows why.

Maradox:
@spyhunter99 I missed the case of flinging when already off-screen. I think this has to do with how the fling is calculated: it flings but not farther than the mapSize. I will have to check if I can do something about that in the case when the fling already occurs outside the bounds.

Regarding forcing the map to stay on screen. I guess we could for example force the map to not scroll further than 1/2 of the mapSize on screen.

It should be further noted that currently (even without this PR), the calculation of the fling does not respect the `scrollableAreaLimit`. If I remember correctly this calculation happens in a class which does not have access to this information.

I will check the problems regarding the lines and polygons (hopefully next week).

Maradox:
@spyhunter99 I fixed the issue regarding the polyline and polygon. I also added a sample for each in `drawing`.

Maradox:
@spyhunter99 I also fixed the problem with the fling when the map is offscreen. If the map is partly offscreen, a fling is allowed to scroll in every direction `+/- mapSize`. I'm not sure if this is the best solution, since now the fling behaves differently dependent on if the map is fully onscreen or partly offscreen - but for now it should fix the problem you encountered.

spyhunter99:
So i just merged the other PR, unfortunately this means there's a bunch of merge issues. I can look at it when i have time, but if you feel obliged, by all means.

spyhunter99:
@Maradox i just looked at it, a few of the conflicts look easy to merge, however i'm not sure i understand the logic differences. Sorry for the merge issue. The plus side is that this library is evolving (rapidly)!

Maradox:
I should be able to solve the conflicts until next friday.

Maradox:
@spyhunter99 

I started solving the conflicts. However I found some issues with the latest PR, which make it difficult to make this feature work again. The following mentioned problems were discovered on the current master-branch, so without any changes from this PR.

1. The drawing of polylines and polygons does not work very well if zoomed out and there are several maps visible - the lines are distributed over several maps. For the new features this means that part of the polylines are outside the map.

2. In earlier versions the map-tiles started at scroll 0/0. This behavior changed and the scroll for the map tiles start at different values now. This may also be the reason for the other mentioned problem. It makes it difficult to use the scroll-position for calculations, as for example the fling.

screenshot_20171005-180655

monsieurtanuki:
@Maradox There's a code collision between your code and mine (#329): both were committed at the same time and dealt with the same classes.
I had to destroy the logic of map scrolling as `View.scrollTo` (limited to `int` values) in order to accept `long` map offsets (necessary in order to deal with zoom levels above 23).
I am currently working on #725 and I'll probably PR by Sunday.

Maradox:
@monsieurtanuki I was aware of that. I encountered the mentioned problems on master (so without any changes from this PR). I hope the polyline issues I encountered are also related to #725, not sure though.

Do you have any information about the scroll-position in relation to the map-tiles? In earlier versions the tile: 0/0 started at map-scroll: 0/0. Thus the scroll position of the map could be easily used to calculate the location of map-tiles and the strart/end of maps (I used this information for example when calculating certain flings). Now this relation seems to be gone however and I have currently no idea in which way a certain tile correlates to a certain scroll-position. Is this an intended change? If yes, could you explain how the relation between a tile and its pixel-position works?

monsieurtanuki:
@Maradox All you need can be found in `Projection` (and possibly `org.osmdroid.util.TileSystem`).

I don't know exactly what you're looking for, but you should have a look at methods like `Projection.toPixels` or `Projection.getPixelFromTile`.

The code triggered by `Projection(MapView mapView)` might be useful to understand how the scrolls are handled, as well as `MapView.scrollTo`:
````java
setMapScroll(x, y);
resetProjection();
invalidate();
````

I hope this helps.

spyhunter99:
looking at resolving the conflicts now, this is one hell of a merge

Maradox:
@spyhunter99 You don't need to. I should be finished soon.

Maradox:
@spyhunter99 I merged master into it. Please note that it is currently a WIP and should not be merged.

@monsieurtanuki For the scroll and fling I would like to limit the scroll to the borders of the map. It seems to me due to the last changes on master, the scroll-position now references the center of the map. But even if I take that into consideration, my calculations do not work when the mapSize is bigger than the screenSize.

I also found that the current implementation of `setScrollableAreaLimit` does not work correctly. I created an issue: https://github.com/osmdroid/osmdroid/issues/742
It would help me a lot if this method would be implemented correctly, so I could adapt it to this PR as well.

spyhunter99:
@Maradox just give the word when you're comfortable with it. I'll run some basic validation tests and then get it merged

spyhunter99:
@Maradox sorry more conflicts, these look easy to fix though

Maradox:
I will try to finish this PR this week.

spyhunter99:
@Maradox is there more work to be done with this?

Maradox:
There are still two problems:

1. When using a bounding box and you fling, at the borders the scroll snaps back. This is rather an issue of the bounding box + fling than this PR. Though, when showing only one map, a bounding-box will be used most of the time and thus the issue becomes more pressing.

2. Polylines and polygons leave the map, when you scroll the map off the center (not sure how to fix that)

screenshot_20171103-104220

Maradox:
It would also be good to review the code, since there are a lot of changes and I replaced many "core" methods by extending them with the `wrapEnabled` parameter (which in some cases is not actually a wrap but a clip. However, I was not sure if it would be better to keep the naming consistence or switch between `wrapEnabled` and `clipEnabled` depending on the method-details.).

Furthermore, I use for example `wrapEnabled` in `ClipToLong `. I'm not sure if this is the correct way to do it and if there will be issues in some cases.

From my point of view the PR could be merged if the code has been reviewed. But there is still work to do, to make it fully workable with the bounding-box.

spyhunter99:
@Maradox do you think it would be possible to make a 3rd option to enable horizontal wrapping but not vertical?

Maradox:
@spyhunter99 I guess so. `wrapEnabled` could be split into `verticalWrapEnabled` and `horizontalWrapEnabled` and all the methods changed in this PR would need to be refactored accordingly. I'm not 100% sure however if there may occur unwanted side-effect with this approach.

Maradox:
@spyhunter99 If you think it is important to the project, I might work on it next week. In any case it would be good to know if the current implementation seems ok to you, or if the passing of the `wrapping` information should be handled differently.

spyhunter99:
well, let's get the single instance of the map working correctly first. i was going to poke around with it this weekend. It's the overlays that aren't be drawn on the non-wrapped map. I vaguely remember where this is calculated. I think it was in projection

spyhunter99:
I got it! The required changes are in the LinearRing class, getBestOffset method. I added a simple boolean flag and now polylines do not just around when panning!

spyhunter99:
currently reviewing the changes and adding a quick sample for plotting markers. found some interesting behavior. I can plot icons north of the map tile (at min zoom). The icon does display and it's latitude value is interesting. It looks like like an artifact of the Mercator projection possibly where by the stretching of coordinates becomes exponential

spyhunter99:
It also appears that projection fromPixels can return a geopoint that is invalid, meaning past the limit. For plotting east/west of the map, the geopoint returned can be past the +-180 longitude

spyhunter99:
image

spyhunter99:
image

spyhunter99:
@Maradox i was able to push the changes to your fork/branch (surprised it worked actually).

@Maradox @monsieurtanuki @MKergall up to you all on if any action is required to resolve this oddity. I think we should correct the invalid longitude at a minimum so that the point is somewhere in bounds.

The strangeness with the latitude projection is up for debate. Not sure what to do with that one. I think the icon should be within the bounds of the map. If replication is OFF, then should we ignore icons based on the plotable range or "fix" it so that the icon is at least rendered at the max bounds?

An alternative for Project.fromPixels could return null or throw an exception should one of these conditions occur

Maradox:
@spyhunter99 I would not limit the icons to the bounds. If the user places an icon outside the bounds he may have a reason to do so (e.g.: For example saving a marker for later use).

From a developers perspective this could also be a feature. Lets say you want to develop a game where users have to name countries. You could start the game by placing markers which the names of the countries on the right side of the world, and the user drags the icon to the correct country.

So I think there might be usecases where markers outside the map could be viable. Thus, I would not restrict them to the bounds of the actual map. Of course it could be implemented as an option.

monsieurtanuki:
> It also appears that projection fromPixels can return a geopoint that is invalid, meaning past the limit. For plotting east/west of the map, the geopoint returned can be past the +-180 longitude

@spyhunter99 The "out of the world" problem doesn't come from `Projection.fromPixels`, which calls `TileSystem.getGeoFromMercator`, that calls `TileSystem.getLatitudeFromY01` and `TileSystem.getLongitudeFromX01`.
And in both methods, results are clipped into min/max latitude/longitude.
But if you saw this happening please provide me with the input parameter values.

spyhunter99:
@monsieurtanuki this was on this branch, using the new sample for plotting markers with map replication turned off. This branch has a bunch of changes and I would appreciate your time looking at the deltas. I don't see any problems other than this one. Anyhow, with replication off and I long press to plot somewhere east or west of the map bounds, then tap on the new icon to bring up the popup, it's easy to reproduce. Look two screen shots up

monsieurtanuki:
@spyhunter99 I won't be able to have a look at it before next Monday. What I can already say is that I'm not in favor of computing the `GeoPoint` of a pixel obviously beyond the map clip area (as in your screenshots)

@Maradox I like the idea of a game with a scrollable map embedded in a dashboard. Should the `MapView` take care of everything, including specifying a clip area dedicated to the map? Should it be done by putting a smaller `MapView` inside/on top of a dashboard `View`? I don't know.

Maradox:
@monsieurtanuki The game with the countries was only meant as an example for why geoPoints of markers should not be limited to `[-180, -90, 180, 90]`.

spyhunter99:
ok, since there is no consensus and there are obviously use cases for both operating scenarios, i'll "fix" it in the demo app and add some commentary related to that effect in the javadocs of the aar and in the apk

spyhunter99:
@MKergall since you were asking for this feature (seems like years ago), do you have any further comments on this? I think this is one of the last items remaining for 6.0.0

Maradox:
One remaining issue is that the zoom does not work correctly in combination with a bounding-box (not related to the PR itself). Which is especially problematic for this PR, because in most cases you will want to limit the scroll when wrapping is disabled. (issue: https://github.com/osmdroid/osmdroid/issues/762).

MKergall:
@spyhunter99  (sorry to react so late)
I tried, and ... Hummm, yes, I have a comment: 
Basically, my wish was to block vertical replication, and to have horizontal replication. This is what all map toolkits are doing, and I think this should be the default behaviour. 
And allowing people to activate vertical replication (but why?) , or block horizontal replication (why? we already have the ability to limit to a BoundingBox) are nice to have. 

And it seems we still can't have that: setMapRepetitionEnabled operates on both vertical and horizontal replications... 

The work already done is impressive. I hope my request may just be a simple API adjustment...

spyhunter99:
bugger. I'll see if i can rework it to be an enumeration type of setting, which would enable 3 modes, default, single map and horizontal replication

MKergall:
Or replace: 

- mapRepetitionEnabled by 2 separate booleans: mVerticalRepetition (default=false) and mHorizontalReplication (default=true)
- and setMapRepetitionEnabled by: setHorizontalReplication and setVerticalReplication

Same changes in TilesOverlay and TileLooper. 
Apparently, we then just have to handle the right boolean at the right line inside TileLooper.loop. 

If you want, I can do that beginning next week.

spyhunter99:
@MKergall go for it. I spent an hour or so attempting the enum route and managed to break everything from zoom animations to all forms of tile looping. Maybe you'll have more luck

MKergall:
Far more complicated than I expected. Working on TileLooper booleans is not sufficient at all. 
Code can be surprising: MapView.mapRepetitionEnabled default to true, when TileLooper.wrapEnabled default to false. 
Still not found how/when the black background is set. 

@Maradox Help! I'm giving up for now.

Maradox:
@MKergall I will try to work on it this week. 

Since you asked why blocking horizontal replication is necessary since there is a bounding box: There are some limitations to that approach (especially in regard to custom maps). For example you cannot zoom further out than the size of the map, because otherwise the repetition becomes visible.

Maradox:
I created an issue for the split between horizontal and vertical repetition for future discussions: https://github.com/osmdroid/osmdroid/issues/787

pmsferreira:
@Maradox where can I find SampleAssetsOnlyWithoutWrapping sample?

Maradox:
@pmsferreira I think `SampleAssetsOnlyRepetitionModes` is what you are looking for.

Issue #353, opened 2016-07-19
Polyline Path too large to render into a texture


2ndGAB:
Hello,

I'm creating a polyline of about 12kms, correctly displayed at zoom level 13 and 14 but for higher zoom level, the path is not displayed anymore and I indefinitely get the following message:

   W/OpenGLRenderer: Path too large to be rendered into a texture

Why? 
How to fix it please?

MKergall:
Known issue: https://github.com/MKergall/osmbonuspack/issues/168

Turn HW acceleration off. No other way: OpenGL cannot handle large values.

2ndGAB:
Turn HW acceleration off ?? :-(((
Ok saw the issue.

spyhunter99:
Just on the map view, not the whole app, that would be silly

2ndGAB:
well, I think that the map view is the one who need HW acceleration unfortunately.

spyhunter99:
In either case, nothing we can do about it at the moment.

2ndGAB:
@spyhunter99 I think [this solution](http://fiskurgit.github.io/osmdroid_polyline_performance) is a good approach.
I'l give it a try.

spyhunter99:
Nice find! Clipping at map bounds is critical

spyhunter99:
It may be useful to include a similar algorithm in osmdroid. A clip to
bounds method. It be also apply to polygons too.

MKergall:
Warning: Clipping at map bounds would be nice, but is not obvious. 

A "simple" optimization was made in PathOverlay: when the 2 points of a segment are outside the view, don't draw it. This solution is wrong, introducing numerous issues (glitches on segments), because there are various cases of a segment traversing the view even if both points are outside the bounds. 

The fiskur solution seems to be the same: "only render the points that will appear on screen". So as he says, he also get "occasional glitches". 

In addition, this simple optimization is not sufficient to avoid the "path too large" issue, because you can have a segment with one point inside the view, and the second point very far away. 

So, to be able to turn HW acceleration on, a real clipping on bounds seems necessary. 
It's really a pitty that Android/OpenGL doesn't provide that out of the box!

2ndGAB:
@MKergall . I'm going to give it a try in the next days and see how to optimize it.
Not drawing a segment is maybe better than not drawing anything at all but in case, it could be not so difficult to create an extra point on the bound for segment starting inside and going outside.

MKergall:
[...]

PR #723, opened 2017-09-27
Feature/#329 long based coordinates with max zoom 29

monsieurtanuki:
Zoom level is now up to 29.


monsieurtanuki:
@spyhunter99 Any help regarding conflict resolution is welcomed, such as magical git instructions.

spyhunter99:
mmm you need to merge master into your branch.
assuming your local master is up to date
git checkout master
git pull origin master
git checkout feature/#329
git merge master
git push origin feature/#329

spyhunter99:
super exciting by the way

spyhunter99:
This is amazing

spyhunter99:
Testing results:
- With hardware acceleration turned on
  - holy crap this is fast
  - lines are ok to max zoom (expected)
  - polygons still have the same issue as before, disappears at zoom + 2 (expected)
  - polygons still have the world wrapping issue (expected)
- With hardware acceleration turned off
  - map means slightly slower but its probably close to about usual or it's just my phone
  - polylines seem fine up to max zoom
  - polygons, at high zooms, the border drawn around the polygon disappears at z=25. Not sure why, maybe @MKergall would have some insight. Holes are fine. Still has the wrapping issue (expected)
  - latlon grid lines i saw one case where the label disappeared. It may be my crummy math though and I don't think it handles past zoom 20 anyhow, so this is somewhat expected. No worries, not asking you to fix this

fantastic work!

OpenStreetMapViewer/src/main/java/org/osmdroid/samplefragments/tileproviders/SampleVeryHighZoomLevel.java
```diff
@@ -0,0 +1,58 @@
...
+import org.osmdroid.tileprovider.modules.MapTileAssetsProvider;
+import org.osmdroid.tileprovider.modules.MapTileModuleProviderBase;
+import org.osmdroid.tileprovider.tilesource.ITileSource;
+import org.osmdroid.tileprovider.tilesource.XYTileSource;
+import org.osmdroid.tileprovider.util.SimpleRegisterReceiver;
+import org.osmdroid.util.GeoPoint;
+import org.osmdroid.views.overlay.ScaleBarOverlay;
+
+/**
+ * A lousy example of very high zoom levels.
```
spyhunter99: This is brilliant

osmdroid-android/src/main/java/org/osmdroid/views/MapView.java
```diff
@@ -342,20 +339,35 @@ public Projection getProjection() {
 		return mProjection;
 	}
 
+	/**
+	 * Use {@link #resetProjection()} instead
```
spyhunter99: i'm a little confused by the javadoc

monsieurtanuki: The only use of `setProjection` was `setProjection(null)`. The actual use was to reset the projection, so that it gets recomputed later in `getProjection` (after a change of zoom level or center).

OpenStreetMapViewer/src/main/java/org/osmdroid/samplefragments/tileproviders/SampleVeryHighZoomLevel.java
monsieurtanuki: Thank you! I like it when I just click on the zoom out button, from level 29 down to 0.
Btw if you have access to "less abstract" tiles up to zoom level 29, or if you feel like adding polygons / polylines as examples, don't hesitate to enhance this demo.

monsieurtanuki:
@spyhunter99 What is the "wrapping" issue with `Polygon`s? Is that something like that:
* in very low zoom levels
* `GeoPoint`s can be theoretically projected on the screen in several `Point`s (modulo world size)
* and as it's the top left `Point` that is chosen by default
* depending on the map scroll, the `Polygon`s can have different shapes, including unexpected ones?

Is there currently an open issue on that?

spyhunter99:
#476 #353 https://github.com/MKergall/osmbonuspack/issues/168
https://www.youtube.com/edit?o=U&video_id=S4Pf2u9WSyQ
https://www.youtube.com/edit?o=U&video_id=EQsi38PIYm8

spyhunter99:
for clarify, that youtube video was with the patch as applied in the open PR with hardware acceleration off

monsieurtanuki:
As far as I understood the expected results, I've just fixed [https://www.youtube.com/watch?v=S4Pf2u9WSyQ](https://www.youtube.com/watch?v=S4Pf2u9WSyQ)
Might solve [https://www.youtube.com/watch?v=EQsi38PIYm8](https://www.youtube.com/watch?v=EQsi38PIYm8) too.
Might even solve [MKergall/osmbonuspack#168](https://github.com/MKergall/osmbonuspack/issues/168) as `Polygon`'s coordinates were truncated into `intE6`, which is not the case anymore.
If needed I could have a look at performances issue #476 and #353. Is #476 fixed or no - there's a PR pending, right?

spyhunter99:
test results
 - hw acceleration off
   - polyline disappears at zoom 25
   - polygon border disappears at zoom 25
 - hw acceleration on
   - polyline disappears at zoom 26
   - polygon disappears when zooming in, it seems to be linked to large polygons that are primarily all offscreen

spyhunter99:
regardless, i don't think any of these are major blockers for getting this pr merged. it's already a significant improvement. they didn't build rome in a day

monsieurtanuki:
device-2017-09-30-183753
@spyhunter99 I'm a bit surprised: could you give an example about `Polygon`'s border not displaying at zoom 25? Attached is an example of `Polygon` at zoom level 29.

In the meanwhile I'll fix `Polyline` as I fixed `Polygon`.

spyhunter99:
sure thing, i'll pull again, clean build, then take a video capture. I had draw the polygon at something like zoom =2 then zoomed in, keeping the border on screen at all times

monsieurtanuki:
I changed idea: I won't edit `Polyline` as I don't have a clear understanding of what is naturally expected on the screen in case of very low zoom level.

@spyhunter99 Understood: I just got the same bug by drawing on zoom level 0 and zooming to level 22.

monsieurtanuki:
I'll fix the `Polygon` border issue in another PR. From what I understood, the problem is that computed pixel coordinates get "too big" for a `Path`, especially with hardware acceleration flag on (but not that big either: it's still below `Integer.MAX_VALUE`). The solution will be to approximate the compute pixels to (sic) "not too big" values.

For the current PR, no further dev on my side.
@spyhunter99 Ready to merge?

spyhunter99:
I'm fine with it. @MKergall any opinions?

Issue #725, opened 2017-10-01
Polyline and Polygon issues at higher zoom levels with/wo hardware acceleration


spyhunter99:
Priority:
This is blocking issue for 6.0.0 release. If we can't get everything rendering at the higher zoom levels, we may have to set the max zoom to the level at which we can render the content at.

Goals:
Provide proper rendering of polylines and polygons at any zoom level supported by the map view/projection with and without hardware acceleration turned on (really on is the goal)

Current issues after the merge of #329 #723 

- With hardware acceleration off
  - [ ] polyline disappears at zoom 25
  - [ ] polygon border disappears at zoom levels > 22
  - [ ] polygon has strange rendering behavior when the map wraps n/s, e/w
 - hw acceleration on
   - [ ] polyline disappears at zoom 26
   - [ ] polygon disappears when zooming in, it seems to be linked to large polygons that are primarily all offscreen see #353. Small polygons are not effected by this. The solution is an accurate clipping mechanism that supports all these strange edge cases when stuff is within or outside of the view port, holes and all at poles

Gotcha's
- Polygons are complex. 
   - "holes" that are partially offscreen, fill colors.
   - polygons in which all points are off screen, however the view port is within the polygon
   - polygons in which all points are off screen, however the view port is partially over the polygon
   - polygons in which all points are off screen, however the view port is completely over the polygon
 - lines
   - 1 or more points are on screen others are off screen
   - All points are on screen
   - All points are off screen, however the viewport covers an area in between some of the points
   - Poll/dateline wrapping

Considerations
- Performance is a big one

Directly related issues
 - MKergall/osmbonuspack#168
 - #353 Polyline Path too large to render into a texture
 - #476 Polygon drawing issue
 - #675 Polygon issues with the dateline
 - https://www.youtube.com/edit?o=U&video_id=S4Pf2u9WSyQ
 - https://www.youtube.com/edit?o=U&video_id=EQsi38PIYm8

Indirectly related issues
 - #595 scaling during animations of polyline

monsieurtanuki:
Some remarks

We shouldn't have different codes for `Polygon` and `Polyline`. They should both use `Path`. A `Polyline` behaves as a `Polygon` except that there are no holes, no fill and no `Path.close()`.

In my tests polygon borders even disappear at level 23 (when I draw a polygon at level 0)

The expected behavior of `Polygon`s in zoom level 0 is not clear for me, because we potentially have several pixels that match the same `GeoPoint`.
[...]

PR #729, opened 2017-10-06
Feature/#725

monsieurtanuki:


monsieurtanuki:
@spyhunter99 The good news is that I've just fixed #725.
The bad news (is that really news?) is that I'm once again stuck with branch conflicts. Luckily I practice meditation, which prevents me from collapsing each time I try to PR (and almost systematically fail because of conflicts). I think I would be more relaxed if I eventually get rid of the optimistic concept of git predictability. And call the doctor each time I have a pain in the git.
Doctor @spyhunter99, sorry to bother you again, would you help PR?

monsieurtanuki:
I switched off my computer, switched it on, performed random tasks on git, and the conflicts disappeared.
I guess part of my problems come from Android Studio's dealing with git.

PR #741, opened 2017-10-16
Feature/#736

monsieurtanuki:
New class `ZoomButtonsOverlay` to be used in `MapView` instead of `ZoomButtonsController`.


spyhunter99:
i like it, just waiting on travis

MKergall:
Well, I'm not so enthousiastic... The design, why not. But the API... 
I want to set it (center,bottom) as it was before: why is it so complex and unclear? 
Previous button was vanishing automatically when no pan - this feature disapeared.

spyhunter99:
hmm i see your point, it is significantly different than the old setup

spyhunter99:
moving this conversation to https://github.com/osmdroid/osmdroid/issues/823

PR #746, opened 2017-10-18
Bug/#742 ScrollableAreaLimit not working

monsieurtanuki:
Fixed, cf. demo "More Samples / Events / Limited scroll area"

### DIFF
--- a/OpenStreetMapViewer/src/main/java/org/osmdroid/samplefragments/events/SampleLimitedScrollArea.java
+++ b/OpenStreetMapViewer/src/main/java/org/osmdroid/samplefragments/events/SampleLimitedScrollArea.java
@@ -3,19 +3,14 @@ package org.osmdroid.samplefragments.events;
 import org.osmdroid.samplefragments.BaseSampleFragment;
 import org.osmdroid.util.BoundingBox;
 import org.osmdroid.util.GeoPoint;
-import org.osmdroid.views.MapView;
-import org.osmdroid.views.Projection;
-import org.osmdroid.views.overlay.Overlay;
+import org.osmdroid.views.overlay.Polygon;
 
-import android.content.Context;
-import android.graphics.Canvas;
 import android.graphics.Color;
-import android.graphics.Paint;
-import android.graphics.Point;
-import android.graphics.Rect;
 import android.view.Menu;
 import android.view.MenuInflater;
 import android.view.MenuItem;
+
+import java.util.ArrayList;
 
 /**
  * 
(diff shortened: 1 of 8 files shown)


Maradox:
This works beautifully. Good job

Maradox:
There seems to be an issue with the polygon in the sample (the limit itself works fine) .

First screen shows the whole area.

screenshot_20171023-120109

When I zoom more in, the polygon is always too short at the bottom of the screen.

screenshot_20171023-120125

monsieurtanuki:
Very strange indeed. I don't have this behavior on my smartphone: we both see the same south map limit, but on my screenshot the polygon is correctly displayed.
20171123 bug742

Maradox:
I just checked, and it happens for all polygons and polylines. (also in the sample: drawing/polygon and drawing/polyline). It seems to effect only the lower part. Could this something have to do with the toolbar? Because the "cropped" area seems to be as high as the toolbar.

monsieurtanuki:
@Maradox Then it could be a problem with too small a clipping area for Polygons (cf. #725).
In `LinearRing`, `Polygon`s and `Polyline`s are clipped into a small [-600, -600, 1400, 1400] area.
Why so? In order to deal with
* very high pixel values (that go beyond the int values, for instance on zoom 29), and 
* some kind of Android bug related to hardware acceleration
But why -600 and 1400? Those are just experimental values estimated on my smartphone (something like a 2000x2000=4Mb area), as there are no explicit values given by Android in order to avoid the bug.

My assumption is that you have a very high definition that goes beyond the high y=1400 pixel limit. That would explain why the lower part of your `Polygon`s are truncated to a y=1400 line.
* Is that relevant to you?
* Could you try to play with `LinearRing.mClipMax` (why not 1800) and `mClipMin` (why not -200) values?

Maradox:
I tested with a Nexus5, which has a resolution of 1080×1920. 
`mClipMax = 1400` and `mClipMin = -1000` seems to solve the problem for that device. However, many devices have a higher resolution than that. All recent samsung galaxies for example have a resolution from 1440x2560 to 1440x2960. 

Could the clipping area not be dependent on the device size? Is there a way to convert the device size in `mClipMin` and `mClipMax`? I guess the maximum device length (width or height), has to be lower than `mClipMax-mClipMin`. What I'm not sure about is, where the `0-line` of `mClipMin` and `mClipMax` is.

monsieurtanuki:
I suggest we continue on #725 ...

PR #747, opened 2017-10-19
Bug/#745 Double-Tap-Zoom ignores location of zoom

monsieurtanuki:
Fixed the bug, as far as I understand it (I just compared with an old version of the code).


spyhunter99:
testing now

spyhunter99:
looks great as usual. thanks for the pr!

PR #751, opened 2017-10-23
Bug/#745bis zoom in map center

monsieurtanuki:
The map center after the double-tap zoom-in animation now matches the end of the animation.

### DIFF
--- a/osmdroid-android/src/main/java/org/osmdroid/views/MapController.java
+++ b/osmdroid-android/src/main/java/org/osmdroid/views/MapController.java
@@ -8,6 +8,7 @@ import android.annotation.TargetApi;
 import android.graphics.Point;
 import android.graphics.Rect;
 import android.os.Build;
+import android.util.Log;
 import android.view.View;
 import android.view.animation.Animation;
 import android.view.animation.Animation.AnimationListener;
@@ -188,6 +189,7 @@ public class MapController implements IMapController, OnFirstLayoutListener {
      */
     @Override
     public void setCenter(final IGeoPoint point) {
+        Log.e("MapController", "setCenter");
         // If no layout, delay this call
         if (mMapView.mListener != null) {
             mMapView.mListener.onScroll(new ScrollEvent(mMapView, 0, 0));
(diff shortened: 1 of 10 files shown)


Maradox:
It seems to me there is still a little shift at the end of the animation. Could this be a rounding problem? Otherwise it is much better than before.

monsieurtanuki:
I saw that little shift too. May be a rounding problem. I'd need to understand more about `View` animation in order to try to fix it.

spyhunter99:
@monsieurtanuki is this ready to merge?

Maradox:
I think this should not be merged until the shift-issue is resolved. Not sure what causes the shift though.

osmdroid-android/src/main/java/org/osmdroid/views/MapController.java
```diff
@@ -471,8 +473,15 @@ protected void onAnimationStart() {
     }
 
     protected void onAnimationEnd() {
+        final double newZoom = mMapView.mTargetZoomLevel.get();
+        final Projection newProjection = mMapView.getProjection().getOffspring(
+                newZoom, mMapView.getProjection().getScreenRect());
+        final IGeoPoint newCenter = newProjection.fromPixels(
```
spyhunter99: possibly related to the open issue. not saying there's something wrong with this, just leaving myself a note to look at this in the near future

monsieurtanuki: This code is 95% correct. The shift @Maradox and I experienced is only something like a 5 pixel shift, just after the end of the zoom animation (where I need to say: "now that the animation is done, let's center on that GeoPoint").
I need to understand more about the animation in order to compute (or retrieve) the animation's last map center.

monsieurtanuki: Just committed a fix:
* in fact the zoom-in animation refreshes the `MapView`'s `Projection` many times, with increasing values of scales from 1f to 2f - the question being how to get the correct map center after the animation
* in the constructor of `Projection`, I now compute a new member (`mCurrentCenter`) - the screen center as a `GeoPoint`
* after the zoom-in animation, I reuse the previous `Projection`'s `mCurrentCenter` for my `MapView` refresh

@Maradox Do you find this version convincing?

Maradox:
Sometimes there is still a tiny little shift, but it is much better than before and doesn't always occur.
I think the remaining issue is neglectable for now and thus the issue can be closed. Maybe in the future that issue could be fully resolved, but I don't know how. It is interesting though, that the issue didn't occur in previous versions of osmdroid.

Issue #734, opened 2017-10-11
Maximum Zoom: floating-point


Maradox:
Since in the current master it is now possible to zoom by floating-point values, it would be nice to also allow a floating-point for the maximumZoom: `setMaxZoomLevel(double)`.

spyhunter99:
Good idea

spyhunter99:
@Maradox should this be a blocking issue for v6? or could it wait until a follow up version?

Maradox:
It would be more consistent with the rest of the changes of v6. Do we need to keep the old methods or could we just replace them with their `double` versions? Maybe I have time for a PR this week. Otherwise I guess v6 could be shipped without it.

PR #752, opened 2017-10-24
# Use double for maximum and minimum zoom level

Maradox:
`setMaxZoomLevel` and `setMinZoomLevel` now use `Double` instead of `Integer`.

Inside `CacheManager` the `double` are casted to `int`. I'm not sure this is the correct way to go though.

issue: https://github.com/osmdroid/osmdroid/issues/734


monsieurtanuki:
I need to have a look at `CacheManager`. Some zoom `int` castings make sense as we download `int` zoom tiles. But there must be other places where we actually lose precision.

spyhunter99:
compile failure

Maradox:
@spyhunter99 Compiler error should be fixed now.

PR #753, opened 2017-10-26
simple Robelectric test for the StarterMap activity; LeakCanary comme…

spyhunter99:
…nted out


pasniak:
Wouldn't you want to keep the LeakCanary? I am not really familiar how it works (another reason I commented it out for this demo).

spyhunter99:
@pasniak yes. leak canary is used for debug builds of the demo app. it traps and detects memory leaks. i opened the pr just to get a better view of the deltas

Issue #761, opened 2017-11-03
Min and max zoom are ignored for pinch-zoom animation


Maradox:
When using the 2 finger pinch-zoom gesture, the minimum and maximum zoom level are ignored for the animation.

Example: max-zoom = 10, current zoom = 9
The user can start a pinch-zoom gesture which zooms beyond zoom 10. On releasing the fingers, the zoom snaps back to 10.

Expected behaviour:
The animation for the pinch-zoom gesture should respect the minimum and maximum zoom levels.

PR #767, opened 2017-11-09
Bug/#761 Respect minimum and maximum zoom during pinch-zoom

Maradox:
The pinch-zoom gesture now respects the minimum and maximum zoom. 

Previously it was possible to exceed the minimum and maximum zoom of the mapview  during a pinch-zoom. When the gesture was finished, the zoom snapped to the minimum and maximum zoom. Now the pinch-zoom respects the minimum and maximum zoom and does not allow to exceed these values even during the gesture.

The new behaviour can be tested via  the `MinMaxZoomLevel` sample.

Issue: https://github.com/osmdroid/osmdroid/issues/761


spyhunter99:
sweet, waiting for travis. good work

Issue #781, opened 2017-11-23
ItemizedOverlay onDrawItem with map zoom


Maradox:
It would be nice if `ItemizedOverlay.onDrawItem()` would also include the map-zoom.
This would allow subclasses to use the zoom for custom logic.

## Issue Type

[ ] Question
[ ] Bug
[X] Improvement
[ ] Build system related
[ ] Performance
[ ] Documentation

PR #782, opened 2017-11-23
#781 Add zoom and scale parameters in onDrawItem of ItemizedOverlay

Maradox:
This PR adds the zoom and scale of the map to the parameters of `ItemizedOverlay.onDrawItem(...)`. This allows subclasses to use this parameters for custom logic regarding drawing the icons.

issue: https://github.com/osmdroid/osmdroid/issues/781

### DIFF
--- a/osmdroid-android/src/main/java/org/osmdroid/views/overlay/ItemizedOverlay.java
+++ b/osmdroid-android/src/main/java/org/osmdroid/views/overlay/ItemizedOverlay.java
@@ -157,7 +157,7 @@ public abstract class ItemizedOverlay<Item extends OverlayItem> extends Overlay 
 
             pj.toPixels(item.getPoint(), mCurScreenCoords);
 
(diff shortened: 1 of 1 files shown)


Maradox:
Please note that `ItemizedOverlay` already has its own implementation which does not need these parameters. So I understand if you think this should not be merged and in such a case developers should develop their own overlay class.

spyhunter99:
since this is an api change (which is fine) would it make more sense to just pass in a reference to the mapview itself? Then any needed values can be queried for rendering purposes and we won't need to update the api again then next time a new value is available in mapview and it is required for rendering

Maradox:
I thought about it, but felt that having access to the mapview in this method could be a breach of separation of concerns. However, the mapview is already available in the `onDraw` method itself, so it may not be that big of a deal.

spyhunter99:
if the map view is available onDraw for the overlay class, then let's just make it available for the itemized icon overlay

Maradox:
@spyhunter99 It now uses the mapview as a parameter.

Issue #787, opened 2017-11-27
Support horizontal and vertical repetition configuration


Maradox:
Support the configuration of horizontal as well as vertical repetition.

Repetition cases:
* None
* Vertical
* Horizontal
* Vertical and Horizontal

spyhunter99:
#183 related

MKergall:
Tested: yes!

PR #788, opened 2017-11-27
Feature/#787 Repetition Modes

Maradox:
`MapView` now supports the configuration of the repetition in the horizontal and vertical direction:
* `setHorizontalMapRepetitionEnabled(boolean)`
* `setVerticalMapRepetitionEnabled(boolean)`

Issue: https://github.com/osmdroid/osmdroid/issues/787

### DIFF
--- a/OpenStreetMapViewer/src/main/java/org/osmdroid/samplefragments/SampleFactory.java
+++ b/OpenStreetMapViewer/src/main/java/org/osmdroid/samplefragments/SampleFactory.java
@@ -20,9 +20,11 @@ import org.osmdroid.samplefragments.cache.SampleSqliteOnly;
 import org.osmdroid.samplefragments.animations.AnimatedMarkerTimer;
 import org.osmdroid.samplefragments.data.AsyncTaskDemoFragment;
 import org.osmdroid.samplefragments.drawing.DrawPolygonHoles;
+import org.osmdroid.samplefragments.drawing.DrawPolygonWithoutVerticalWrapping;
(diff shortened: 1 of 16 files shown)


Maradox:
1. If preferred, I could change the API of the `MapView` to use an `Enum` for the repetition modes.

2. I am currently not sure if the adjustment in `LinearRing` is correct (at the location of the TODO).

spyhunter99:
Great work! After some quick testing, this appears to be correct with one exception. When panning the map, (vertical off, horizontal on), the algorithm that determines where the draw overlays needs to be adjusted. Currently the plotted points/lines/polygons scroll all the way off screen. The previous behavior (vertical on, horizontal on) this can be viewed easily by zooming to 0, then panning the map around. As the line/polygon/point pans off screen, it will get redrawn on the most appropriate world.

spyhunter99:
@Maradox looking at linear ring now, that's the problem area

spyhunter99:
might have fixed it, probably not as efficient as it could be though..

spyhunter99:
@MKergall  how's this?

MKergall:
Not tried yet, but it seems good! Absolutely no need for an Enum. 
If you merge on master, I will test. 

@Maradox You seem reluctant to change default behaviour, and default vertical replication to false. Is it just to respect compatibility, or do you have another reason?

spyhunter99:
@MKergall merged, it's all you

Maradox:
@MKergall No, just for compatibly reasons.

PR #789, opened 2017-11-28
bug/#762: taking the current scale into account

monsieurtanuki:
Impacted classes:
* `MapView`: modified method `checkScrollableAreaBoundingBox` in order to take into account the current scale
* `Projection`: added member `Matrix mScaleMatrix` in order to deal with the scale factor without the rotation; added method `scalePoint` in order to apply `mScaleMatrix` to a point; added helper method `applyMatrixToPoint`

Bad news: the map angle isn't taken into account. The results with an angle do not look like a big bug but are not perfect.
I think that in general many things should be fixed regarding the map angle, including the MiniMap behavior. In another issue I guess...


spyhunter99:
map angle is described in another ticket. testing this now

Issue #791, opened 2017-11-28
Add directional arrows to polylines


jmarks213:
## Issue Type
[x] Improvement

I needed to add directional arrows to the polylines for indicating direction between points and wanted to share the end result.  Maybe someone else will find it useful.

For example,
![screenshot_20171128-155839](https://user-images.githubusercontent.com/15203304/33344851-51cdc122-d458-11e7-8aa4-1abaccca855d.jpg)

spyhunter99:
#611 related issue

This is pretty cool! Testing now, i can see lots of use cases for aviation maps, vectors and waypoints, etc

pasniak:
Is it doable to have labels aligned to polylines? Something like:
![screenshot from 2017-12-21 20-44-18](https://user-images.githubusercontent.com/19778666/34281899-c64affee-e68f-11e7-9e78-24576939669e.png)

monsieurtanuki:
@pasniak I think this goes far beyond the "arrow" feature for many reasons, among them: what parametrized text should we display and how should we split the text if the segment zigzags? In comparison, for arrows we just always display the same small `Path` (or bitmap in  #809).
That being said, the implementation of the angled text would be something like
````java
canvas.save();
canvas.rotate(orientation, centerX, centerY);
// here it's a bitmap, but you can put a text instead
canvas.drawBitmap(bitmap, centerX - bitmap.getWidth() / 2, centerY - bitmap.getHeight() / 2, null);
canvas.restore();
````

spyhunter99:
@pasniak one approach you could take is to look at the milstd2525 renderer. There's a standardized graphic for this. It looks a lot like an air space corridor or something like that. Your (developer) workflow would be:

- add the renderer to your project
- add osmbonusback
- figure out what the symbol code is
- figure out your control points (basically the start/end points and a width)
- have the renderer make you a kml doc
- use osmbonus pack to convert the kml to a folder overlay
- add the folder overlay to the map

the following is what it looks like in the spec (milstd2525c), which is in the ball park of what you're looking for. http://www.dtic.mil/doctrine/doctrine/other/ms_2525c.pdf for the full spec

![image](https://user-images.githubusercontent.com/1975321/34316530-ccee8b78-e765-11e7-92e9-09d9c19805fd.png)

of course, you could just build the polygons yourself. The text labels can be created on the fly too with the marker class

PR #792, opened 2017-11-28
Add directional arrows on polylines

jmarks213:
for #791


monsieurtanuki:
@jmarks213  I like this idea of improving the look of the `Polyline`s, and that makes perfect sense to add arrows.

My comments:
* Would it be possible to add a `Bitmap` instead of other `Path`s?
* Would it be possible to do the same thing through a carefully set up `Paint`?
* It should be implemented for `Polygon`s too
* I don't think `LinearRing` is the perfect place to add the "arrow logo" code: I would keep `LinearRing` as is (the code was already big enough) and extend it into something like `ArrowLinearRing`, dedicated to the arrow logo additional feature. That would then work for `Polygon`s too. From what I read in your code, I would say that something like overriding `SegmentClipper.SegmentClippable.lineTo` would do the trick: each time I add a point with `lineTo`, I add the corresponding arrow logo (if the distance with the previous point is big enough). And that new class should be easy to implement, as you just have to move your code from `LinearRing`. Or even, `LinearRing` should have a settable array of `SegmentClippable` members, typically a standard one drawing the mere polyline (to be created, named `SegmentStandard`?), and another one displaying the arrow logos (to be created, named `SegmentArrow`?).

What do you think of it?

spyhunter99:
Ran into a few issues drawing very large polylines. All of the lines have the arrows set to true.
image

spyhunter99:
just pushed the sample along with fixing the merge conflicts

spyhunter99:
I also noticed when drawing, some of the lines would immediately have the arrows and some would not. Not entirely sure why but it may be related to multiple "worlds" being visible concurrently

spyhunter99:
image
last one.  seems to work fine with zooming, scaling and rotation

jmarks213:
Thanks for the comments!

@monsieurtanuki 
At first, I did use bitmaps to implement this but it ended up being messier than it needs to be.  You end up needing multiple bitmaps for each cardinal direction that, after rotating and resizing, end up being pixelized and elongated.  Drawing them directly creates a cleaner end result.  I think your idea of a separate class extending linear ring is a good idea.  I did not consider this being relevant for Polygon as it is not an object I have needed to use before.

@spyhunter99 
Thanks for the helpful feedback.  The default value for the length of the sides is a static number so it looks fine when the Polyline is thin but when the Polyline has more width the arrow is not also taking up more space.  I will change the default value to be based off the stroke width of the line instead of a static number.

I am not sure what you have done with the transparent Polylines?  I am not familiar with that case but it can be that the arrows would be preferred to also not be drawn.

monsieurtanuki:
@jmarks213 As you are in favor of my idea of a separate class, I'll work on that once your code is merged.

jmarks213:
@monsieurtanuki That's fine with me.  

I updated it as is to scale the arrows with the stroke width.  It looks better.  I do see the issue where the arrows are not drawing correctly over the polyline at the extreme zoom levels.  I will hold off on looking into it?  I think you may get to the fix with the refactor you have in mind.

screenshot_20171201-113627
screenshot_20171201-113617

monsieurtanuki:
> I do see the issue where the arrows are not drawing correctly over the polyline at the extreme zoom levels

@jmarks213 Could you be more specific / attach a screenshot? Or were you referring to @spyhunter99's screenshot on zoom level 1 with polylines on the left side of the screen and arrows on the right side?

jmarks213:
@monsieurtanuki Yes, the screenshots @spyhunter99 provided show the behavior.  The arrows are getting decoupled from the polyline.  If you rotate the map the arrows and polyline will sometimes line back up as expected.

spyhunter99:
it's acting like the map offset isn't being included.

image
image

spyhunter99:
the issues with the arrows also appears at all zoom levels once the map has wrapped at least once

jmarks213:
Okay, I understand it a bit better now.  The projected pixels are not updated to reflect the offset so they are not a good representation of the lines that are drawn to the screen.  I implemented @monsieurtanuki  thoughts and it indeed does draw correctly now.  I am now getting a rogue arrow for each line but I have to stop for the day.  I'll take another look tomorrow.


screenshot_20171202-112708

jmarks213:
I finished with ArrowsLinearRing.  It seems to be working fine for all cases now.  I added the arrows to Polygon as well.  

Have I overlooked anything?

monsieurtanuki:
@jmarks213 Did you manage to fix the rogue arrow bug? If you're done, we can merge.

jmarks213:
@monsieurtanuki Yes, it is fixed.

screenshot_20171205-091129
screenshot_20171205-091038

monsieurtanuki:
@jmarks213 Awesome!
I'll play with this new feature, and I may open related issues on the following topics:
* So far the arrows are displayed as `Path`s, but I think we should be able to display bitmaps instead (and even to orient them).
* I don't know how the distance between arrows is computed, but I believe there might be something counter-intuitive in that. For instance, imagine we have a polyline on the equator line, from W180 to E180. If there are only 2 points in my polyline, if there are 20 points, if there 100 points, should we have the same number of arrows? I don't know what users should expect. I think this is not the case in your code, because if you look at the round part of a polyline in your latest screenshots there are no arrows, because I suspect the round part is a succession of close points (where there's not enough space for arrows). Maybe we should rather talk in terms of distance _from the start of the polyline_: "I draw an arrow every x pixels I drew from the very polyline start, not from the current segment start".

jmarks213:
@monsieurtanuki Thanks for taking an interest and providing your comments.  

- As I mentioned before, I found the bitmap solution to be messier than it needs to be.  I look forward to seeing what you come up with.

- Your second point I agree with.  In the current form arrows are drawn at the midpoint between every two points not considering distance.  Drawing arrows at consistent distances along the line was something I considered originally but didn't take the time to look into; Arrows between points was more appropriate for my case.  I would like to keep arrows at mid points as an available option but I do welcome the additional option for arrows at distance along the line.

monsieurtanuki:
@jmarks213 About bitmaps, I'm pretty sure it's safe and easy, because that's the way markers are currently displayed.
In fact, in order to display a map rotation, the `MapView` canvas is rotated. And as you need markers to be displayed without rotation, you have to rotate them with the opposite rotation value.
Cf. the (edited) code in `ItemizedOverlay.onDrawItem`:
````java
canvas.save();
canvas.rotate(-aMapOrientation, x, y);
marker.copyBounds(mRect);
marker.setBounds(mRect.left + x, mRect.top + y, mRect.right + x, mRect.bottom + y);
canvas.scale(1 / scaleX, 1 / scaleY, x, y);
marker.draw(canvas);
marker.setBounds(mRect);
canvas.restore();
````

For the "consistent distance versus segment middle only" arrow display enhancement, we need to compute the actual coordinates (in order to compute the distance in pixels between segment points) and at the same time the truncated coordinates (that we now use in order to simplify the polylines).

MKergall:
My 2 cents: 
Bitmaps would be great for full flexibility on arrow shape. 
But then, the user can set polyline color and size: arrow bitmap should take that into account...

monsieurtanuki:
@MKergall My goal is not to display red or black arrows, but to give the user the possibility to draw (not necessarily oriented) bitmaps, e.g. flags.
Then the user can use whatever polygon line color, polygon background color and bitmap he wants. And we can provide as examples a black arrow bitmap (to be oriented) and a work-in-progress-cone bitmap (to be not oriented).

PR #793, opened 2017-11-29
Use apply on SharedPreferences. More efficient. Fixes Robolectric tests.

pasniak:


monsieurtanuki:
@pasniak You're right, `apply` is much better than `commit`.
Wouldn't it be even nicer if you created a static method in `DefaultConfigurationProvider` so that we can more easily implement this best practice; something like this?
````java
public static void commit(final SharedPreferences.Editor pEditor) {
    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.GINGERBREAD) {
        pEditor.apply();
    } else {
        pEditor.commit();
    }
}
````

Issue #801, opened 2017-12-10
MapTileApproximater OutOfMemoryError when zooming out deep


miguel-porto:
## Issue Type
[x] Bug

## Description and/or steps/code to reproduce the problem
With wi-fi turned on, pinch zoom out a large amount quickly and do not lift fingers (say, with fingers going from the screen corners to the center fast). The following exception occurs, because too many tiles are being loaded and scaled at the same time. Note that, until you lift up the fingers, the tiles will be loaded at the higher resolution (before the zoom started), and that makes hundreds of tiles being loaded and scaled and the same time at the lower zoom level, beforel you lift the fingers. And this throws this exception, causing the app to crash and vanish. I suggest that the resolution at which the tiles are being loaded should be updated during pinch zoom, to avoid loading hundreds of high res tiles simultaneously.

Anyway, congratulations for the MapTileApproximater - it *greatly* increased UX for offliners!

```
12-10 22:19:26.961 6813-6965/? I/art: Waiting for a blocking GC Alloc
12-10 22:19:26.968 6813-6968/? I/OsmDroid: Error downloading tile: /17/62449/50269
java.lang.OutOfMemoryError: Failed to allocate a 262156 byte allocation with 0 free bytes and 3GB until OOM
   at dalvik.system.VMRuntime.newNonMovableArray(Native Method)
   at android.graphics.Bitmap.nativeCreate(Native Method)
   at android.graphics.Bitmap.createBitmap(Bitmap.java:831)
   at android.graphics.Bitmap.createBitmap(Bitmap.java:808)
   at android.graphics.Bitmap.createBitmap(Bitmap.java:775)
   at org.osmdroid.tileprovider.modules.MapTileApproximater.getTileBitmap(MapTileApproximater.java:250)
   at org.osmdroid.tileprovider.modules.MapTileApproximater.approximateTileFromLowerZoom(MapTileApproximater.java:206)
   at org.osmdroid.tileprovider.modules.MapTileApproximater.approximateTileFromLowerZoom(MapTileApproximater.java:184)
   at org.osmdroid.tileprovider.modules.MapTileApproximater.approximateTileFromLowerZoom(MapTileApproximater.java:148)
   at org.osmdroid.tileprovider.modules.MapTileApproximater.approximateTileFromLowerZoom(MapTileApproximater.java:130)
   at org.osmdroid.tileprovider.modules.MapTileApproximater$TileLoader.loadTile(MapTileApproximater.java:111)
[...]

Issue #757, opened 2017-11-01
MapTileModuleProviderBase.loadMapTileAsync() causes ANR


Vandalko:
I don't have exact steps to reproduce right now - just running 6.0.0 version and trying to display a few markers. It does not occur each time.

```
Long monitor contention event with owner method=void org.osmdroid.tileprovider.modules.MapTileModuleProviderBase.loadMapTileAsync(org.osmdroid.tileprovider.MapTileRequestState) from MapTileModuleProviderBase.java:135 waiters=4 for 379ms
```

Here is the full trace:
```
"main" prio=5 tid=1 Blocked
  | group="main" sCount=1 dsCount=0 obj=0x74fc8fa0 self=0xb4d76500
  | sysTid=13815 nice=0 cgrp=default sched=0/0 handle=0xb6f85b34
  | state=S schedstat=( 0 0 0 ) utm=179 stm=51 core=2 HZ=100
  | stack=0xbe298000-0xbe29a000 stackSize=8MB
  | held mutexes=
  at org.osmdroid.tileprovider.modules.MapTileModuleProviderBase.loadMapTileAsync(MapTileModuleProviderBase.java:135)
  - waiting to lock <0x0730f43e> (a java.lang.Object)
  at org.osmdroid.tileprovider.MapTileProviderArray.mapTileRequestFailed(MapTileProviderArray.java:152)
  at org.osmdroid.tileprovider.modules.MapTileModuleProviderBase$1.removeEldestEntry(MapTileModuleProviderBase.java:122)
  at java.util.LinkedHashMap.addNewEntry(LinkedHashMap.java:189)
  at java.util.HashMap.put(HashMap.java:403)
  at org.osmdroid.tileprovider.modules.MapTileModuleProviderBase.loadMapTileAsync(MapTileModuleProviderBase.java:147)
  - locked <0x0037129f> (a java.lang.Object)
  at org.osmdroid.tileprovider.MapTileProviderArray.mapTileRequestFailed(MapTileProviderArray.java:152)
  at org.osmdroid.tileprovider.modules.MapTileModuleProviderBase$1.removeEldestEntry(MapTileModuleProviderBase.java:122)
  at java.util.LinkedHashMap.addNewEntry(LinkedHashMap.java:189)
  at java.util.HashMap.put(HashMap.java:403)
  at org.osmdroid.tileprovider.modules.MapTileModuleProviderBase.loadMapTileAsync(MapTileModuleProviderBase.java:147)
[...]

PR #804, opened 2017-12-11
Handling tiles that failed due to max queue size being reached

secaw:
Right now, when a tile can not be retrieved due to the max queue size being reached, the callback method `mapTileRequestFailed` is being called just like as if the corresponding provider doesn't have this tile. This is problematic, as `MapTileProviderArray` is then looking for the next appropriate provider for this tile, which we don't want in case the queue is full. This causes too much CPU and Memory load.
Related:  #757, probably also #801

This PR introduces a separation between `mapTileRequestFailed` and the new `mapTileRequestFailedExceedsMaxQueueSize` in:

- IMapTileProviderCallback.java
- MapTileModuleProviderBase.java
- MapTileProviderBase.java
- MapTileProviderArray.java

### DIFF
--- a/osmdroid-android-it/src/main/java/org/osmdroid/tileprovider/modules/MapTileProviderTest.java
+++ b/osmdroid-android-it/src/main/java/org/osmdroid/tileprovider/modules/MapTileProviderTest.java
@@ -1,6 +1,6 @@
 /*
  * WARNING, All test cases exist in osmdroid-android-it/src/main/java (maven project)
- * 
+ *
  * During build time (with gradle), these tests are copied from osmdroid-android-it to OpenStreetMapViewer/src/androidTest/java
  * DO NOT Modify files in OpenSteetMapViewer/src/androidTest. You will loose your changes when building!
  *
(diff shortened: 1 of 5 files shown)


secaw:
@spyhunter99 Why does it fail? Sorry, it's my first PR, anything I can do?

spyhunter99:
It's a combo of travis and the avd. I'll kick it

spyhunter99:
you have to go to travis and look at the raw log. Usually there is a reason for the failure. In this case, it's a compile error
````
/home/travis/build/osmdroid/osmdroid/osmdroid-android-it/src/main/java/org/osmdroid/tileprovider/modules/MapTileProviderTest.java:30: error: <anonymous org.osmdroid.tileprovider.modules.MapTileProviderTest$1> is not abstract and does not override abstract method mapTileRequestFailedExceedsMaxQueueSize(MapTileRequestState) in IMapTileProviderCallback
	final IMapTileProviderCallback mTileProviderCallback = new IMapTileProviderCallback() {
	                                                                                      ^
Note: Some input files use or override a deprecated API.
Note: Recompile with -Xlint:deprecation for details.````

secaw:
Oh, I see - thanks, I think I got it now...

secaw:
@spyhunter99 I think I resolved the compile error and reran the checks three times on the same code, always with different failures. Once two of the three checks passed, just one failed, once two failed and now all three failed again? Can't find anything meaningful in the log. What am I missing?

spyhunter99:
@secaw looking at this now

secaw:
@spyhunter99 Looking good now, thanks.

Issue #832, opened 2017-12-27
MapView should handle onPause/onResume of the parent application


spyhunter99:
## Issue Type

[ ] Improvement

## Description and/or steps/code to reproduce the problem

There's a whole bunch of strange and silly code that users of osmdroid have to work around in order to prevent memory leaks, sensor leaks, etc. This primarily happens when the application is paused (screen off or switch to a different app). This issue will add onPause and onResume on the map view and will clean up all the nonsense copy and paste boilerplate code.

This will be an API change that is unfortunately unavoidable. Affected classes

- OverlayManager
- DefaultOverlayManager
- MapView
- ZoomButtonsOverlay
- MyNewGpsLocationOverlay
- CompassOverlay

spyhunter99:
implemented on branch feature/#823

InI4:
That sounds good, though I need to work out the details to learn, what code I can throw away.

Issue #825, opened 2017-12-25
Zoom Buttons in 6.1.0


InI4:
## Issue Type

[2] Improvement
[1] Bug

## Description and/or steps/code to reproduce the problem
Tried moving from 5.6.5 to 6.0.0. I really like the smooth zooming!

Found the following 3 issues with the zoom buttons.

1. I have custom overlays and Zoom Buttons are hard to spot on top of my overlay. Would be nice to have a way to adjust alpha without writing a new overlay.
2. New Zoom Buttons touch each other, being very small for "thumbs", would be nice to have the buttons about as wide as the old ones. Allowing the programming to set this up (or oven to put buttons in top of each other, instead od left right) would be a nice to have.
3. I catch a onLongPress on my map/overlay. Every touch of the Zoom control buttons subsequently creates such a touch event on the map. I consider this a bug. My guess is, it has to do with the zoom animation making the touch event last longer.

### If it's a bug, version(s) of android this affects:

Tested with Android 4.2 and 6.0

### Version of osmdroid the issue relates to:

6.0.0-SNAPSHOT

spyhunter99:
Great feedback! Thanks

InI4:
Hmm, now bug or here? I have 2 major problems with Zoom Buttons right now:

1. As stated above, long press events on map coming from short press on zoom. I have no workaround for this, this makes my UI useless.
2. setBuiltInZoomControls() semantics has changes, as these calls do not automatically turn buttons on or off, but I need to send an invalidate() to the map, to be sure, the buttons get updated. It is at least a behaviour change asking to be documented.

spyhunter99:
working on #823, the long press should be fixable along with the rest of this issue

spyhunter99:
root cause, the current zoombuttons overlay implementation was catching onTouch instead of using the overlay based mechanisms for onSingleTapConfirmed. This was causing confusion of the various gesture detection code that was causing both the zoom control to fire and a long press. This was discovered with #831 (plotting circles). 

The fix, which i just implemented, reworks on the ZoomButtonsOverlay to behave like any other overlay. The down side (and probably the reason why it was implemented with onTouch and directly integrated with mapView) is that the drawing order of overlays can now cause issues where the zoom buttons are either below or above other overlays. The default setup should now draw, the tiles overlay, followed by zoom controls, then everything else. This will change if the built in zoom controls are turned off, then on again, which will paint the zoom buttons last.

[...]

Issue #823, opened 2017-12-24
Restore ZoomButton's auto hide feature


spyhunter99:
## Issue Type

[ ] Improvement

## Description and/or steps/code to reproduce the problem

The previous setup, using Android's built in zoom buttons would leak references while the animation for fading the buttons in and out.

https://github.com/osmdroid/osmdroid/pull/741
https://github.com/osmdroid/osmdroid/issues/736

This issue to restore that function

spyhunter99:
In order to support the auto hide, we'll need to add mapview handling for onpause and onresume, which means all app that use osmdroid will also need to add these calls to activity or fragment using osmdroid.

spyhunter99:
@monsieurtanuki the current issue with this branch is that the zoom controls work intermittently and i have no idea why. Once in a state of not working, panning the map, even just by one pixel, usually restores the functionality. 

I'm starting to think that it may be easier to create and attach a custom view/layout to the map view and just rely on that instead of calculating our own hit test. The other option is to revert the zoom buttons overlay back to the original android based zoom controls

monsieurtanuki:
@spyhunter99 Initially, when I developed the current version of the zoom buttons, it was supposed to be a quick win in order to get rid of a harmless exception caused by the standard Android zoom buttons, when I was interested by no specific issue.

If it's a pain in the neck to fine tune our own version of zoom buttons, I guess we can forget it and switch back to the standard version. And survive with that harmless exception that nobody cares about - except those who read the logs.

spyhunter99:
sadly i think we should. We can revisit this for 6.1 or some other future version. 6.0 has already taken much longer than expected. The core features and performance are more important that zoom buttons IMO

monsieurtanuki:
Definitely.

spyhunter99:
ok, working on it now...

monsieurtanuki:
Fixed by #1190. Feel free to reopen.

PR #833, opened 2017-12-28
ZoomButtonsOverlay auto hide + onPause/resume for map/overlays

spyhunter99:
#823 Zoom Buttons Overlay auto hide
#832 onPause/onResume for the mapView and for all overlays
#825 zoom buttons issues (for reference)


spyhunter99:
onDetached and onAttachedToWindow are the only android lifecycle things that we can actually use from mapview, primarily because the map view is a view, not an activity or fragment.

onDetached is usually called from the activity's onDestroy function. An onPause is therefore necessary in order to pause any sensor related stuff, background threads, etc

spyhunter99:
since we are deferring the zoom buttons stuff, this will be partially merged manually

Issue #835, opened 2017-12-30
Polyline rendering issue with repetition on


spyhunter99:
## Issue Type

[ ] Bug

## Description and/or steps/code to reproduce the problem

In the draw polyline sample, if I draw some reasonable length lines, then switch to panning move and pan the map in any direction, the previous configuration would draw lines/markers/polygons on instance of the world that was mostly within the view port. In other words, as you keep panning the map, the map overlays would continue to be drawn on a map instance that is on screen. This still functions fine for markers, but it stopped working recently for lines (vertical and horizontal) and for polygons (horizontal panning doesn't work as expected, however panning vertically does)

### Version of osmdroid the issue relates to:

6.0.0/master

monsieurtanuki:
I guess it's related to the recent "world replication on/off" devs. I'll have a look at it.

monsieurtanuki:
Solved in #838

Issue #820, opened 2017-12-24
Performance issue with the heatmap demo


spyhunter99:
## Issue Type

[ ] Performance

## Description and/or steps/code to reproduce the problem

More samples > data > heatmap with async loading

once the "heat map" loads, which is just a set of non overlapping polygons with transparency, scroll panning is awful with hardware acceleration is on. With hardware acceleration off, it's smooth.

### If it's a bug, version(s) of android this affects:

testing on android 6

### Version of osmdroid the issue relates to:

6.0.0-SNAPSHOT on master

monsieurtanuki:
I'll have a look at it.

monsieurtanuki:
Here we are talking about hundreds of `Polygon`s being displayed!
I guess that for that demo (with square areas), simple `Path`s would be faster, because each time we `draw` a `Polygon` a lot of computation is involved.
What I can do is try to optimize the `LinearRing.buildPathPortion` part.
But the `HeatMap` demo itself could be optimized too.

shiroyagi:
Yes, maybe manage the texture bitmap as mentioned on SO, or blend a hardware and a software layer if that is possible.

monsieurtanuki:
I just had a go with a simple `drawRect` instead of a `Polygon`, and it looks very smooth to me.
Is there a specific reason why we should use `Polygon` for Heat Map, or are alternate solutions like `drawRect` acceptable? Something like that:
````java
new Overlay() {
	private final GeoPoint mNorthWest = new GeoPoint(key.getLatNorth(), key.getLonWest());
	private final GeoPoint mSouthEast = new GeoPoint(key.getLatSouth(), key.getLonEast());
	private final Point topLeft = new Point();
	private final Point bottomRight = new Point();

	@Override
	public void draw(Canvas c, MapView osmv, boolean shadow) {
		osmv.getProjection().toPixels(mNorthWest, topLeft);
		osmv.getProjection().toPixels(mSouthEast, bottomRight);
		c.drawRect(topLeft.x, topLeft.y, bottomRight.x, bottomRight.y, paint);
	}
};
````

pasniak:
I have hardware acceleration on because otherwise map seems sluggish (on tablet).

Recently I added bunch of irregular map areas as Polygons (they are concave and most of the time they are outside the user's view) and when I switch on the layer, the map becomes painfully slow.

What are my options? Is it possible to disable HW acceleration on the fly?  I thought I had to set it before MapView is instantiated?

[...]

PR #838, opened 2017-12-31
Optimized the display for `Polyline`

monsieurtanuki:
Here are the improvements:
* for `Polyline` we now use `Canvas.drawLine` and we don't care about invisible segments: they may have an impact on visible Path but never on visible segments
* introduction of `ListPointL` as an optimization of `List<PointL>`, which is supposed to benefit both `Polyline` and `Polygon` display

New classes:
* `ListPointL`: an optimized version of `List<PointL>`
* `ListPointLTest`: a unit test class for `ListPointL`

Impacted classes:
* `LinearRing`: used new class `ListPointL` for constantly refreshed `List<PointL>`; handled the case where `mPath` is `null`; refactored a little
* `PathBuilder`: handled the case where `mPath` is `null`
* `Polyline`: removed the `Path` and handled the display with `Canvas.drawLine`
* `SegmentClipper`: added `boolean mPathMode`, to be set to `true` when dealing with a `Path` (if so we keep the invisible segments: they have an impact on Path inner area)

### DIFF
--- a/OpenStreetMapViewer/src/main/java/org/osmdroid/samplefragments/drawing/CustomPaintingSurface.java
+++ b/OpenStreetMapViewer/src/main/java/org/osmdroid/samplefragments/drawing/CustomPaintingSurface.java
@@ -115,8 +115,10 @@ public class CustomPaintingSurface extends View {
         if (map!=null){
             Projection projection = map.getProjection();
             List<GeoPoint> geoPoints = new ArrayList<>();
+            final Point unrotatedPoint = new Point();
             for (int i=0; i < pts.size(); i++) {
-                GeoPoint iGeoPoint = (GeoPoint) projection.fromPixels(pts.get(i).x, pts.get(i).y);
+                projection.unrotateAndScalePoint(pts.get(i).x, pts.get(i).y, unrotatedPoint);
+                GeoPoint iGeoPoint = (GeoPoint) projection.fromPixels(unrotatedPoint.x, unrotatedPoint.y);
                 geoPoints.add(iGeoPoint);
             }
             //TODO run the douglas pucker algorithm to reduce the points for performance reasons
--- a/osmdroid-android/src/main/java/org/osmdroid/util/LineBuilder.java
+++ b/osmdroid-android/src/main/java/org/osmdroid/util/LineBuilder.java
@@ -1,0 +1,48 @@
+package org.osmdroid.util;
+
+/**
+ * Created by Fabrice on 03/01/2018.
+ * @since 6.0.0
+ */
+
+public class LineBuilder implements PointAccepter {
+
+    private final ListPointL mPoints = new ListPointL();
+    private float[] mLines;
+    private int mSize;
+
+    @Override
+    public void init() {
+        mPoints.clear();
+        mSize = 0;
+    }
+
+    @Override
+    public void add(final long pX, final long pY) {
+        mPoints.add(pX, pY);
+    }
+
+    @Override
+    public void end() {
+        mSize = mPoints.size() * 2;
+        if (mSize < 4) {
+            return;
+        }
+        if (mLines == null || mLines.length < mSize) {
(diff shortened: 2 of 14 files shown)


monsieurtanuki:
@shiroyagi Could you please check if you have decent performances on polylines with this PR's optimizations?

shiroyagi:
With 3 points the performance is now similar to 5.6.6 however with 2k points it is still not good.  Also calling clearPath() does not clear the path but freezes it relative to the viewport.

monsieurtanuki:
...working on the subject...
As Path looks slower than drawLines, I'm going to develop a drastically different version for Polyline. There is a lot of code dedicated to offset: for Polygon holes and multiworld issues; I'll optimize it for Polyline.
I'll fix #835 too in the process.

monsieurtanuki:
I'm very optimistic. It will be OK by the end of the week.

monsieurtanuki:
@shiroyagi Now there's a specific optimized code for `Polyline`s. I hope it's fast enough for your needs: could you have a look at the performances?

The `Polygon`s are still slow with the hardware acceleration on, but that should rather be dealt in #820. I guess the idea would be to create a demo without `MapView`, just to infirm/confirm that dozens of `Path` cannot be displayed with decent performance, regardless of the osmdroid context.

monsieurtanuki:
@spyhunter99 I guess I just fixed #835 in this latest commit. Could you have a look at it?

shiroyagi:
Yes, this is much better thanks.  It is still not quite as good as 5.6.6 which was comfortably below the 60fps green line target in the GPU profiler.  It is now around and occasionally above this line (perhaps 30-50% slower and less consistent).  The proportion of update/execute rendering is roughly the same so the cause is not immediately obvious.  Could it be the clipping is constantly changing the list of lines passed whereas before it could be cached on the GPU?

shiroyagi:
With 4k points the height of the profiler's blue 'update' bars increases and the red 'execute' bars decrease as you zoom in until the chart is nearly all blue.   This is consistent with more time spent clipping but less being drawn .  Unfortunately the total height increases, but maybe with less frequent clipping performance could actually be better than 5.6.6?

monsieurtanuki:
I just had a look at the previous Polyline code:
* memory is more used nowadays, because int E6 coordinates and Point pixels used to be used, but it's not possible anymore (due to zoom 29) and we use now double and PointL. More intensive memory use may explain as side effect the random performance you now experience
* drawLines used to be called by chunk of 64 segments, but I use only one call of drawLines with a large float array. I can switch back to those more numerous calls of smaller arrays.

Of course, what I say only make sense if you have many many points and many many visible segments in your current performance test. Is that the case?

monsieurtanuki:
It also looks like there used to be the minimum of home made clipping, and that the actual clipping was taken into account by drawLines.

shiroyagi:
> drawLines used to be called by chunk of 64 segments

I only did that to keep the buffer reasonably small without impacting performance, it shouldn't actually be faster.

> the actual clipping was taken into account by drawLines.

ie. done in hardware when enabled?  This may be significantly more efficient (but problematic at zoom >=25).

spyhunter99:
@monsieurtanuki re: the map repetition bug#835, it's fixed for vertical panning but not horiztonal

monsieurtanuki:
@shiroyagi Just pushed my latest fixes: less clipping computation and less memory used during drawing.
Closer to the "original" fast code 6 months ago.

@spyhunter99 To me the demo "Draw a polyline on screen" works as expected: each polyline is drawn so that its center is as close to the screen center as possible (worldmapreplicationwise). I mean, I double-checked horizontally and vertically. What's the problem?
* this business rule is not what you expect
* it's not what you see

spyhunter99:
@monsieurtanuki appears to be fixed now

shiroyagi:
Great, that's fixed.  Same performance as before.

PR #840, opened 2018-01-03
Check for real network availabilty (connectivity) in NetworkAvailabliltyCheck

secaw:
Having experienced a problem with unstable network conditions, where osmdroid tried to load tiles unavailingly instead of approximate them when there was no actual connectivity, I discovered that `NetworkAvailabliltyCheck` was checking for `NetworkInfo.isAvailable()` instead of `NetworkInfo.isConnected()` as advised here: https://developer.android.com/training/basics/network-ops/managing.html
> Note that you should not base decisions on whether a network is "available." You should always check isConnected() before performing network operations, since isConnected() handles cases like flaky mobile networks, airplane mode, and restricted background data.


spyhunter99:
I'm OK with this

secaw:
Great, can be merged - tested locally and working.

PR #852, opened 2018-01-06
Fixed scroll limit offsets computation bug due to world replication

monsieurtanuki:


spyhunter99:
@monsieurtanuki sorry looks like a caused a conflict. assuming it's fixed, is this ready to merge?

PR #854, opened 2018-01-06
Bug/#853 - improve redraw of SimpleFastPointOverlay

miguel-porto:
SimpleFastPointOverlay layer when set with MAXIMUM_OPTIMIZATION (for displaying tens of thousands of points) had a redraw issue, in that the points were not correctly positioned during pinch zoom. This PR fixes the issue, making it feel smoother during pinch zooms.


miguel-porto:
Thanks for the promptness!

PR #867, opened 2018-01-14
setCenter now able to work for every GeoPoint within the scroll limits

monsieurtanuki:
How is it possible?
Now you can add an `int extraSize` parameter when you set the scroll limits.
And if you use
* `mapView.getWidth()/2` for longitude extra size limit
* `mapView.getHeight()/2` for latitude extra size limit
you have enough flexibility to
* comply with the limits
* and set the map center to a `GeoPoint` within these limits

### DIFF
--- a/OpenStreetMapViewer/src/main/java/org/osmdroid/samplefragments/events/SampleLimitedScrollArea.java
+++ b/OpenStreetMapViewer/src/main/java/org/osmdroid/samplefragments/events/SampleLimitedScrollArea.java
@@ -81,9 +81,15 @@ public class SampleLimitedScrollArea extends BaseSampleFragment {
 
 		mMapView.getController().setZoom(13.);
 
-		setLimitScrollingLatitude(true);
-		setLimitScrollingLongitude(true);
 		setHasOptionsMenu(true);
+
+		mMapView.post(new Runnable() { // "post" because we need View.getWidth() to be set
+			@Override
+			public void run() {
+				setLimitScrollingLatitude(true);
+				setLimitScrollingLongitude(true);
+			}
+		});
 	}
 
 	/**
@@ -93,7 +99,7 @@ public class SampleLimitedScrollArea extends BaseSampleFragment {
 		mMapView.getOverlays().remove(mNorthPolyline);
 		mMapView.getOverlays().remove(mSouthPolyline);
 		if (pLimitScrolling) {
-			mMapView.setScrollableAreaLimitLatitude(sCentralParkBoundingBox.getActualNorth(), sCentralParkBoundingBox.getActualSouth());
+			mMapView.setScrollableAreaLimitLatitude(sCentralParkBoundingBox.getActualNorth(), sCentralParkBoundingBox.getActualSouth(), mMapView.getHeight() / 2);
 			mMapView.setCenter(sCentralParkBoundingBox.getCenterWithDateLine());
 			mMapView.getOverlays().add(mNorthPolyline);
 			mMapView.getOverlays().add(mSouthPolyline);
@@ -110,7 +116,7 @@ public class SampleLimitedScrollArea extends BaseSampleFragment {
 		mMapView.getOverlays().remove(mWestPolyline);
 		mMapView.getOverlays().remove(mEastPolyline);
 		if (pLimitScrolling) {
-			mMapView.setScrollableAreaLimitLongitude(sCentralParkBoundingBox.getLonWest(), sCentralParkBoundingBox.getLonEast());
+			mMapView.setScrollableAreaLimitLongitude(sCentralParkBoundingBox.getLonWest(), sCentralParkBoundingBox.getLonEast(), mMapView.getWidth() / 2);
 			mMapView.setCenter(sCentralParkBoundingBox.getCenterWithDateLine());
 			mMapView.getOverlays().add(mWestPolyline);
 			mMapView.getOverlays().add(mEastPolyline);
--- a/OpenStreetMapViewer/src/main/java/org/osmdroid/samplefragments/events/SampleLimitedScrollArea.java
+++ b/OpenStreetMapViewer/src/main/java/org/osmdroid/samplefragments/events/SampleLimitedScrollArea.java
@@ -3,9 +3,8 @@ package org.osmdroid.samplefragments.events;
 import org.osmdroid.samplefragments.BaseSampleFragment;
 import org.osmdroid.util.BoundingBox;
 import org.osmdroid.util.GeoPoint;
-import org.osmdroid.views.overlay.Polygon;
+import org.osmdroid.views.overlay.Polyline;
 
-import android.graphics.Color;
 import android.view.Menu;
 import android.view.MenuInflater;
 import android.view.MenuItem;
@@ -25,7 +24,8 @@ public class SampleLimitedScrollArea extends BaseSampleFragment {
 
 	public static final String TITLE = "Limited scroll area";
 
-	private final int MENU_LIMIT_SCROLLING_ID = Menu.FIRST;
+	private final int MENU_LIMIT_SCROLLING_LAT_ID = Menu.FIRST;
+	private final int MENU_LIMIT_SCROLLING_LNG_ID = Menu.FIRST + 1;
 
 	private BoundingBox sCentralParkBoundingBox;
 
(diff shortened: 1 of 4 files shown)


secaw:
Fantastic, but the following happens: if I set `mapView.setScrollableAreaLimitLatitude(89.75,-89.75, mapView.getHeight()/2);` with `mapView.setVerticalMapRepetitionEnabled(false)`, then it's working great on Zoom 2, which is my min zoom: the map stops scrolling vertically at about half of my mapview height. But when I zoom to let's say level 4 and above, then the map again moves out of the view on max north / south scrolling. I take `pExtraHeight` is not a fixed pixel value, but somehow get's adjusted by zoom values? Can this be fixed?
Also, if I set max north / south in `mapView.setScrollableAreaLimitLatitude` to 90.0 / -90.0 I get an ANR.

secaw:
Probably unrelated to this PR, but first accured now, I get a NPE with `mapView.setBuiltInZoomControls(false);`:
Attempt to invoke virtual method 'void org.osmdroid.views.overlay.ZoomButtonsOverlay.setZoomInEnabled(boolean)' on a null object reference
May have something to do with latest work on built in zoom buttons by @spyhunter99 ?

monsieurtanuki:
@secaw 
> if I set mapView.setScrollableAreaLimitLatitude(89.75,-89.75, mapView.getHeight()/2);

Please try with min/max latitudes `TileSystem.MinLatitude` and `TileSystem.MaxLatitude` (+- 85.05112877980659): that makes no sense going beyond those values under the osmdroid tile system.
Hopefully things should work better; at least we'll be able to dismiss the side effects.
 
In first approach the "extra" sizes should not be impacted by the zoom level.

The NPE may indeed have to do with @spyhunter99 's latest devs around the zoom buttons.

secaw:
@monsieurtanuki Alright, if I set `mapView.setScrollableAreaLimitLatitude(TileSystem.MaxLatitude,-TileSystem.MaxLatitude, mapView.getHeight()/2)`, there's no vertical space at any zoom level.

monsieurtanuki:
@secaw Try to play with demo `SampleLimitedScrollArea`: I think you should be satisfied.
For instance, change the code in the constructor:
````java
sCentralParkBoundingBox = new BoundingBox(
	TileSystem.MaxLatitude, TileSystem.MaxLongitude,
	TileSystem.MinLatitude, TileSystem.MinLongitude);
````
There's one Android tricky part in the code: the `View.getHeight()` might not have been initialized when you set the scroll limits. Therefore its value is 0. Sad.
The trick is to use a `View.post` statement (cf. `SampleLimitedScrollArea.addOverlays`).
I hope this helps.

**Comment by secaw** [comment]
@monsieurtanuki You're right! Just checked with setting `mapView.setScrollableAreaLimitLatitude` in `onResume` instead `onCreate` and it works on the second call of `onResume` now. Fantastic really! Thanks a lot, will investigate further in order to force setting it in a more elegant way.

**Comment by monsieurtanuki** [comment]
@secaw I'm not sure there's an elegant way to do that. My knowledge on the subject comes from [SO](https://stackoverflow.com/questions/3591784/views-getwidth-and-getheight-returns-0).

**Comment by monsieurtanuki** [comment]
@secaw Btw, how good are we now regarding #851?
* may I merge the current code?
* are there additional features you were expecting?
* may I close the issue?

**Comment by secaw** [comment]
Yes, from my point of view definitely ready to merge and close. Thanks again.

File: osmdroid-android/src/main/java/org/osmdroid/views/Projection.java [hunk 1]
```diff
@@ -703,36 +722,49 @@ private void adjustOffsets(final long pDeltaX, final long pDeltaY) {
...
-
-		if (pPixelMax - pPixelMin < pScreenSize) {
-			return pPixelMax - pScreenSize / 2 - (pPixelMax - pPixelMin) / 2;
-		} else if (pPixelMin >= 0) {
-			return pPixelMin;
-		} else if (pPixelMax <= pScreenSize) {
-			return pPixelMax - pScreenSize;
-		} else {
-			return 0;
+	boolean setMapScroll(final MapView pMapView) {
```
spyhunter99: should this be private or protected?

osmdroid-android/src/main/java/org/osmdroid/views/Projection.java
```diff
@@ -703,36 +722,49 @@ private void adjustOffsets(final long pDeltaX, final long pDeltaY) {
 	}
 
 	/**
+	 * @param pPixelMin Pixel position of the limit (left)
```
spyhunter99: so this is public, why would someone want to call this? not clear (to me) based on the method name or the javadocs

osmdroid-android/src/main/java/org/osmdroid/views/Projection.java
```diff
@@ -675,17 +675,36 @@ public void adjustOffsets(final IGeoPoint pGeoPoint, final PointF pPixel) {
...
-		final long deltaY = checkScrollableOffset(top, bottom, mMercatorMapSize, mIntrinsicScreenRectProjection.height());
-		adjustOffsets(-deltaX, -deltaY);
+		adjustOffsets(pBoundingBox.getLonWest(), pBoundingBox.getLonEast(), false, 0);
+		adjustOffsets(pBoundingBox.getActualNorth(), pBoundingBox.getActualSouth(), true, 0);
+	}
+
+	/**
+	 * @since 6.0.0
+	 */
+	public void adjustOffsets(final double pNorthOrWest, final double pSouthOrEast,
```
spyhunter99: this is public, why would someone want to call this? not clear (to me) based on the method name or the javadocs. What offset is being adjusted?

osmdroid-android/src/main/java/org/osmdroid/views/MapView.java
```diff
@@ -667,14 +688,66 @@ public void setScrollableAreaLimit(BoundingBoxE6 boundingBox) {
...
+		mScrollableAreaLimitLatitude = true;
+		mScrollableAreaLimitNorth = pNorth;
+		mScrollableAreaLimitSouth = pSouth;
+		mScrollableAreaLimitExtraHeight = pExtraHeight;
+	}
+
+	/**
+	 * @since 6.0.0
+	 */
+	public void setScrollableAreaLimitLongitude(final double pWest, final double pEast,
```
spyhunter99: is extra width distance in meters, pixels or degrees longitude?

osmdroid-android/src/main/java/org/osmdroid/views/MapView.java
```diff
@@ -667,14 +688,66 @@ public void setScrollableAreaLimit(BoundingBoxE6 boundingBox) {
...
+	 * @since 6.0.0
+	 */
+	public void resetScrollableAreaLimitLongitude() {
+		mScrollableAreaLimitLongitude = false;
+	}
+
+	/**
+	 * @since 6.0.0
+	 */
+	public void setScrollableAreaLimitLatitude(final double pNorth, final double pSouth,
```
spyhunter99: is extra height distance in meters, pixels or degrees latitude?

OpenStreetMapViewer/src/main/java/org/osmdroid/samplefragments/events/SampleLimitedScrollArea.java
```diff
@@ -43,36 +43,87 @@ public String getSampleTitle() {
...
+	private final Polyline mSouthPolyline = new Polyline();
+	private final Polyline mWestPolyline = new Polyline();
+	private final Polyline mEastPolyline = new Polyline();
 
 
 	@Override
 	protected void addOverlays() {
 		super.addOverlays();
 
-		final Polygon polygon = new Polygon();
```
spyhunter99: this whole block goes from a polygon to a set of lines. is there a particular reason?

monsieurtanuki: Definitely: now you can set limits for latitude only and for longitude only. And for both. Therefore the Polygon was not relevant anymore, and was replaced by 2 times 2 Polylines as min/max limits.

osmdroid-android/src/main/java/org/osmdroid/views/MapView.java
monsieurtanuki: In pixels: currently including "pixel" in the parameter's name.

osmdroid-android/src/main/java/org/osmdroid/views/MapView.java
monsieurtanuki: In pixels: currently including "pixel" in the parameter's name.

osmdroid-android/src/main/java/org/osmdroid/views/Projection.java
monsieurtanuki: Currently commenting and downgrading access to package-private.

osmdroid-android/src/main/java/org/osmdroid/views/Projection.java
monsieurtanuki: My guess is package-private, as it's meant to be used only by `MapView`, which happens to be in the same package.

osmdroid-android/src/main/java/org/osmdroid/views/Projection.java
monsieurtanuki: It's public only in order to be accessed by `ProjectionTest`.

spyhunter99: would protected work for this?

osmdroid-android/src/main/java/org/osmdroid/views/MapView.java
spyhunter99: one last question, how does the "extra" distance effect scrolling? is it like padding?

osmdroid-android/src/main/java/org/osmdroid/views/Projection.java
monsieurtanuki: Doesn't work for `ProjectionTest` if `protected` in `Projection`.

osmdroid-android/src/main/java/org/osmdroid/views/MapView.java
monsieurtanuki: There's no real impact on the actual scrolling, except that you can scroll further "extra" pixels beyond the scroll limits.
And to address @secaw 's needs, you can put `MapView.getWidth()/2` and `MapView.getHeight()/2` as longitude/latitude extras: doing so the map can actually be centered via `MapView.setCenter` on any `GeoPoint` within the scroll limits, which was not the case before (before, it was like we had extras of 0 pixel).

spyhunter99:
snapshot posted

osmdroid-android/src/main/java/org/osmdroid/views/MapView.java
secaw: @spyhunter99 
> is it like padding?

Yes, would actually be most descriptive!
pPaddingPixelHeight / pPaddingPixelWidth
instead of pExtra... but well, I think now it's understandable too.

spyhunter99: @secaw see the updated docs
https://github.com/osmdroid/osmdroid/blob/master/osmdroid-android/src/main/java/org/osmdroid/views/MapView.java#L761

MKergall:
As @secaw, wanting to remove vertical replication, I'm trying to block vertical scrolling of poles at middle of the screen, so that map.setCenter at high and low latitudes still works. 

Trying this in onCreate: 
```
map.setVerticalMapRepetitionEnabled(false);
map.setScrollableAreaLimitLatitude(TileSystem.MaxLatitude,-TileSystem.MaxLatitude, map.getHeight()/2);

```
is close to my expectations, with just 1 remaining issue: map.getHeight returns 0. 
Same result when calling it in onResume. 
Hard-coding this extra pixel value works fine. 

I would be very interested by @secaw solution...

monsieurtanuki:
@MKergall I don't know if @secaw eventually used that syntax, but by using `View.post` statement you should have map's height and width (cf. `SampleLimitedScrollArea.addOverlays`).

secaw:
Sorry, somehow missed that, if it still helps or for future reference -
Yes, this does the trick even in `onCreate()`:
```
mapView.post(new Runnable() {
    @Override
    public void run() {
        mapView.setScrollableAreaLimitLatitude(TileSystem.MaxLatitude,-TileSystem.MaxLatitude, (int) mapView.getHeight()/2);
    }
});
```

PR #870, opened 2018-01-14
Bug/#865 fix for info window

spyhunter99:
Fixes the following

- info window not following markers map panning/zooming the map
- info window drifting when rotating the map
- info window not updating it's location while zooming


spyhunter99:
ahh the setRotation method allows you to rotate the marker's drawable separately from the map

PR #879, opened 2018-01-16
Feature/#878 Deprecations

cbalster:
I have two questions:

1) There are 2 sections that use zoom as int and which maybe should be reworked to use double. I will mark them separately. Do these warrant a new Issue? Any concerns about TODO tags in code?

2) Do you want PRs like this one, or would you rather like to do this in a more centralised manner?


osmdroid-android/src/main/java/org/osmdroid/views/overlay/gridlines/LatLonGridlineOverlay.java
```diff
@@ -45,7 +45,8 @@ private static void applyMarkerAttributes(Marker m){
 
     public static FolderOverlay getLatLonGrid(Context ctx, MapView mapView) {
         BoundingBox box = mapView.getBoundingBox();
-        int zoom = mapView.getZoomLevel();
+        //TODO change getIncrementor(), getStartEndPoints*() to use double?
```
cbalster: TODO tag okay? Create Issue?

osmdroid-android/src/main/java/org/osmdroid/views/overlay/ScaleBarOverlay.java
```diff
@@ -366,7 +366,7 @@ public void draw(Canvas c, MapView mapView, boolean shadow) {
 			return;
 		}
 
-		final int zoomLevel = mapView.getZoomLevel();
+		final int zoomLevel = (int) mapView.getZoomLevelDouble(); //TODO change calculation to double?
```
cbalster: TODO tag okay? Create Issue?

spyhunter99:
the double zoom level stuff is new to 6.0.0 thus we haven't deleted the old methods yet...

cbalster:
Sorry about that - somehow I missed that. I changed the commit to just replace some of the internal uses.
Just let me know if you want to do changes like this by yourselves - I only used this as an opportunity to familiarize myself with the codebase.

spyhunter99:
@cbalster for starters, thanks for the PR. All contributions are welcome

> There are 2 sections that use zoom as int and which maybe should be reworked to use double.  will mark them separately. Do these warrant a new Issue? 

The zoom level change is new to v6. We probably won't remove the integer based zoom apis for while.

>Any concerns about TODO tags in code?

Feel free to open issues if you see something.

> Do you want PRs like this one, or would you rather like to do this in a more centralised manner?

The general preference is to isolate work on a single issue to a single branch/pr. Mixing multiple issues can be confusing and cause delay. Also, try not to reformat the entire code case, makes it harder to review. Most of this is covered in the wiki

PR #887, opened 2018-01-21
Fixed crash in Polyline single tap

monsieurtanuki:
When running `LinearRing.isCloseTo`, the lines used to be drawn (again). Which was not only unnecessary, but also sometimes provoked an Android crash for some reason.
With this fix the lines are not drawn again, and therefore the Android crash doesn't happen anymore.


spyhunter99:
tested and verified. i'm a bit puzzled as i thought touch input was on the ui thread.

monsieurtanuki:
@spyhunter99 Well maybe it is in the UI thread, I don't know.
All I can tell is that it crashed during a `canvas.drawLines` statement:
* in an "Android funny way" (no exception, no logs, no nothing)
* for an operation that was irrelevant (drawing lines during a simple distance check)
* anyway in an ugly pattern - calling an `add` method for each point without calling `init` or `clear` before (for the record it usually crashed during my test on the 3rd click attempt, where enough points were added without initialization nor flush)
Well, now it's solved and that's good news.

spyhunter99:
yeah i saw two distinct crashes, a binder error message whereby the window was not found and the windeath seg 11 error. Good job finding the issue! I think we are much closer to release now

PR #889, opened 2018-01-21
Fixed MapView flinging limits

### LINKED ISSUE #845: Mapview repetition bug
There's a bug with horizontal and vertical map repetition, where the map jumps back when scrolled to the edges of a virtual 3x3 map repetition grid. (Might be related to #843)
To reproduce: Scroll the map (e.g. with zoom 2) past the current map in any direction to the next repetition map and further. As soon as you hit the edge of that one, in most cases it flips back to the center of that repetition map instead of seamlessly moving on to the next repetition (see image). This is most annoying with horizontal scrolling, but might also be confusing in the vertical direction. Interestingly, if you're in an 'edge map' and are zooming, you get to the center map again. Also the jump back might not happen when scrolling very slowly.

osmdroid_map-repetition-bug

6.0.0-SNAPSHOT

monsieurtanuki:
There's a good chance that will fix #845.


spyhunter99:
@monsieurtanuki i think this is causing build failures. one of the unit tests is failing

spyhunter99:
org.osmdroid.views.OpenStreetMapViewTest > test_toMapPixels_0_0[test1(AVD) - 2.2] [31mFAILED [0m
	junit.framework.AssertionFailedError: the computed center of the map is in the pixel center of the map (Y)(zoom=0.5291292591692422,center=76.32586730594576,-152.71419392472797,0.0) expected:<381.0> but was:<9.0>

	at junit.framework.Assert.fail(Assert.java:47)

:OpenStreetMapViewer:connectedDebugAndroidTest FAILED

PR #892, opened 2018-01-21
add GroundOverlay2

pasniak:
This is a first cut, it works better than GroundOverlay (with midpoint) from the bonuspack.


spyhunter99:
sounds good, i'll test it later today and whip up a sample. Do you have any relevant test images that can be used for demo/testing purposes? maybe we can find a weather image to use

spyhunter99:
I found this, https://radar.weather.gov/Conus/RadarImg/latest_Small.gif and was able to rig up a sample using it. Looking pretty good! Seems fast enough. Hard part was guessing the correct (or close enough) coordinates

image

pasniak:
that was fast :) it would be great to have some tool to find coordinates by choosing online image location and dragging corners in place...

spyhunter99:
good idea, this could probably be done with a four draggable markers,

Issue #688, opened 2017-08-27
Promiscuous offline tile provider


spyhunter99:
new feature for the offline users. alternate versions of the offline tile provider that ignores tile source names, providing the first tile available from any offline or cache source.

MKergall:
Not sure for this one. Tile sources can be very different (eg satellite photos). Mixing that could look like a bug. 
At least, the feature should be optional.

spyhunter99:
Definitely opt in. Not the default

secaw:
Would be a great feature! E.g. in my use case I already created my own version of `DatabaseFileArchive` to make it source insensitive. Other file archives like MBTiles and GEMF are source insensitive by default. But I didn't manage to do the same with `ZipFileArchive`, which apparently has to have the source name (e.g. 'Mapnik') as filename ('Mapnik.zip') in order to match the source and thus be shown. Btw there's a related issue with `ZipFileArchive` as it doesn't read direct MOBAC zip exports in general, because MOBAC doesn't use the source names as filenames anyways.

spyhunter99:
@secaw I have code for zip archives somewhere. Mobac uses whatever you enter for the layer name as the tile source.

secaw:
@spyhunter99 Unfortunately it's a bit more complicated with Mobac and osmdroid zip:

- osmdroid `ZipFileArchive` expects a zip structure like this: CurrentTileSourceName.zip / [zoomlevel folders]
- Mobac export as 'osmdroid zip' does this: AtlasName.zip / ShortMapSourceName / [zoomlevel folders]

So they are not only incompatible just from the filename, but from the whole structure!

Besides that, having `ZipFileArchive` be source aware doesn't make much sense: Imagine you want to have two offline zip files e.g. one of Papua New Guinea and one of Brooklyn. Both would have to be named 'Mapnik.zip' in order to appear on your openStreetMap, so they could not be used simultaniously. That's why I think `ZipFileArchive` should be source unaware by default just like MBTiles and GEMF file archives.

The way I think about the great offline map functionality of osmdroid is, I'd like to make it most easy for users to include their own maps - meaning that they export them from Mobac, drop them into osmdroid folder on their device and are ready to use them, no matter which (online) tile source they have set. This is now working nicely for me with my custom `DatabaseFileArchive`, GEMF and MBTiles formats. I don't support zip files right now because of the above described filename source awareness and general incompatibilities with Mobac.

[...]

PR #895, opened 2018-01-22
Let DatabaseFileArchive and ZipFileArchive ignore the TileSource

secaw:
As discussed in #688, it would be great to have all offline tile providers work in a way that tiles will be loaded regardless of their associated tile source name. `MBTilesFileArchive` and `GEMFFileArchive` have this behaviour by default.

This PR makes it possible for `DatabaseFileArchive` and `ZipFileArchive` to work the same way by introducing the new method `setIgnoreTileSource(boolean pIgnoreTileSource)` for both of them. When set true, the first available tile will be loaded regardless of the given tile source. (Default remains `false`)

### DIFF
--- a/osmdroid-android/src/main/java/org/osmdroid/tileprovider/modules/DatabaseFileArchive.java
+++ b/osmdroid-android/src/main/java/org/osmdroid/tileprovider/modules/DatabaseFileArchive.java
@@ -83,7 +83,6 @@ public class DatabaseFileArchive implements IArchiveFile {
 			Cursor cur;
 			if(!mIgnoreTileSource) {
 				cur = mDatabase.query(TABLE, tile, COLUMN_KEY+" = " + index + " and "+COLUMN_PROVIDER+" = '" + pTileSource.name() + "'", null, null, null, null);
-
 			} else {
 				cur = mDatabase.query(TABLE, tile, COLUMN_KEY+" = " + index, null, null, null, null);
 			}
--- a/osmdroid-android/src/main/java/org/osmdroid/tileprovider/modules/DatabaseFileArchive.java
+++ b/osmdroid-android/src/main/java/org/osmdroid/tileprovider/modules/DatabaseFileArchive.java
@@ -30,6 +30,7 @@ public class DatabaseFileArchive implements IArchiveFile {
 	public static final String COLUMN_KEY = "key";
 	static final String[] tile_column = {"tile"};
 	private SQLiteDatabase mDatabase;
+	private boolean mIgnoreTileSource = false;
 
 	public DatabaseFileArchive(){}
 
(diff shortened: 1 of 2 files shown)


spyhunter99:
how would one go about setting it to true? I think this probably needs some more changes but it's a good start. The alternative approach would be to extends both Zip and Database archive formats, set the default to true for ignoring the source, then register the new classes with the file archive factory. Thoughts?

secaw:
Hmm, in short I'm doing it like this:
```
List <IArchiveFile> archiveFiles  = new ArrayList();
ZipFileArchive archive = ZipFileArchive.getZipFileArchive(new File(archiveFileDir, archiveFilename));
archive.setIgnoreTileSource(true);
archiveFiles.add(archive);
IArchiveFile[] AarchiveFiles = archiveFiles.toArray(new IArchiveFile[archiveFiles.size()]);
fileArchiveProvider = new MapTileFileArchiveProvider(registerReceiver, archiveTileSource, AarchiveFiles);
```
Maybe also add a second getZipFileArchive() including a `pIgnoreTileSource`?

secaw:
spyhunter99 I get you're thinking about being able to setting this right from the 'default' configurations. Sure, then this here really is only a start, `IgnoreTileSource` must be passed probably beginning with `ArchiveFile.init() `all the way up to `MapTileProviderBasic`.

osmdroid-android/src/main/java/org/osmdroid/tileprovider/modules/DatabaseFileArchive.java
```diff
@@ -70,8 +79,14 @@ public void init(File pFile) throws Exception {
 			final long y = (long) pTile.getY();
 			final long z = (long) pTile.getZoomLevel();
 			final long index = ((z << z) + x << z) + y;
-			final Cursor cur = mDatabase.query(TABLE, tile, COLUMN_KEY+" = " + index + " and "+COLUMN_PROVIDER+" = '" + pTileSource.name() + "'", null, null, null, null);
-
+			
+			Cursor cur;
+			if(!mIgnoreTileSource) {
+				cur = mDatabase.query(TABLE, tile, COLUMN_KEY+" = " + index + " and "+COLUMN_PROVIDER+" = '" + pTileSource.name() + "'", null, null, null, null);
```
spyhunter99: hmm we should probably change this to use something closer to a prepared statement. this looks like sql injection (not related to this issue obviously)

spyhunter99:
i'm fine with merging this as is. Would you mind either adding to or changing the wiki to describe the how to? Or maybe adding a sample to the sample app?

secaw:
Great - Sure, I will think about taking it further or documenting the next days... thanks for your comments!

Issue #846, opened 2018-01-05
MapView enters an infinite redraw loop


miguel-porto:
## Issue Type

[x] Bug

## Description and/or steps/code to reproduce the problem

Sometimes, when pinch-zooming, the map enters an infinite redraw loop, where tiles are reloaded and removed indefinitely. Please see a video shot here: [infinite redraw bug](http://flora-on.pt/infinite-redraw.mp4)
I don't have any clue of what triggers this issue. It just happens "sometimes", not very frequently. If you zoom to other place, sometimes it goes back normal. Sometimes comes back again. You see the video!
When this happens, the app freezes on detach. No error messages are thrown.
Has anybody experienced this? Can you replicate?
Thanks!

### If it's a bug, version(s) of android this affects:

API 23

### Version of osmdroid the issue relates to:

6.0.0-SNAPSHOT

spyhunter99:
trippy!

spyhunter99:
#812 possibly related

miguel-porto:
I've been digging this issue by putting log messages throughout the redraw chain. It turns out that the [MapView's dispatchDraw](https://github.com/osmdroid/osmdroid/blob/5198ee03925d915c617b1b31517fdcb5d4b0f787/osmdroid-android/src/main/java/org/osmdroid/views/MapView.java#L1047), under some circumstances, starts getting called repeatedly, indefinitely, and this is causing the tiles continuously flicker.
So yes, it's an infinite redraw loop, the problem. But what's causing dispatchDraw to be called repeatedly? Maybe there is some misplaced invalidate()? Through where goes the feedback loop? I have no idea yet, and also no idea of what circumstances trigger this issue - but it happens only on zooming out, I think. I have no other overlays but the tile overlay.

spyhunter99:
can't help but wonder if it's related to #164. I'm trying to reproduce this on this end, not successful yet. There is a configuration setting which dictates the size of the in memory cache

secaw:
@miguel-porto Are you creating your own `MapTileProviderArray`? Then it's probably related to something I've been experiencing some time ago...

Quote from my own comment on 6 Sep 2017 in #646
> Since I did not use MapTileProviderBasic, but MapTileProviderArray directly, tile loading was stuck in an infinite loop with expired tiles and when offline, because 'isDowngradedMode()' always returned false. I created my own MapTileProviderArray now based on MapTileProviderBasic, which solved this problem and actually a few others.

[...]

PR #903, opened 2018-01-26
Displayed tiles are now safely stored in memory

monsieurtanuki:
We could even have a cache capacity of 0!

How do we do that?
* just before drawing the tiles, we list them
* using that list, we specifically ask the cache not to remove the LRU tile if it's among the displayed tiles

Isn't using the LRU enough?
* apparently not (cf. #846)
* neither theoretically (cf. `MapTileApproximater` potential side effects)

Now that we have the current list of all displayed tiles, we can do more refined cache as @shiroyagi suggested ("introducing some hysteresis to reduce disk/network access")

### DIFF
--- a/osmdroid-android/src/main/java/org/osmdroid/util/MapTileIndex.java
+++ b/osmdroid-android/src/main/java/org/osmdroid/util/MapTileIndex.java
@@ -1,0 +1,42 @@
+package org.osmdroid.util;
+
+import org.osmdroid.tileprovider.MapTile;
+
+import microsoft.mappoint.TileSystem;
+
+/**
+ * Computes a map tile index as `long` to/from zoom/x/y
+ * Algorithm unfortunately different from SqlTileWriter.getIndex for historical reasons.
+ * This version is better, because it's easy to get zoom, X and Y back from the index.
+ * This version is limited to zooms between 0 and 29, which should be enough.
+ * @since 6.0.0
+ * @author Fabrice Fontaine
+ */
+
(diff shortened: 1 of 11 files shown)


miguel-porto:
Haven't had time to further explore the issue since then, but this indeed seems to be the solution for the Xmas tree effect!! Thank you, I'll check it our as soon as possible.

osmdroid-android/src/main/java/org/osmdroid/util/MapTileIndex.java
```diff
@@ -0,0 +1,42 @@
...
+
+    private static int mMaxZoomLevel = TileSystem.primaryKeyMaxZoomLevel;
+    private static int mModulo = 1 << mMaxZoomLevel;
+
+    public static long getTileIndex(final MapTile pMapTile) {
+        return getTileIndex(pMapTile.getZoomLevel(), pMapTile.getX(), pMapTile.getY());
+    }
+
+    public static long getTileIndex(final int pZoom, final int pX, final int pY) {
+        return (((long)pZoom) << (mMaxZoomLevel * 2))
```
spyhunter99: i kind of feel that we should put this algorithm somewhere. it's used in several locations throughout the code base

monsieurtanuki: The algorithm is unfortunately slightly different in `SqlTileWriter.getIndex`. And mine is better ;)
If possible, we should reuse `MapTileIndex` now.

PR #907, opened 2018-01-28
Feature/#74

shiroyagi:
This implements the following:

1. 'pointer mode' to draw an arrow indicating the device orientation on the map instead of a conventional north/south needle (using a different bitmap to avoid any confusion)
2. option to draw the compass in the centre of the map
3. option to add a bearing offset for magnetic declination

I think this makes it much easier to navigate by following a route on the map.

I found trying to centre the compass by other means quite awkward, with dimensions not known in many methods, and seemingly arbitrary offsets for different devices used in the examples.

I couldn't bring myself to use tabs in the file, especially as the guidelines say not to, so the indentation may be broken depending on the tab width setting.

As in this screenshot (the crosshair on top of the compass is in a different overlay):
pointer


spyhunter99:
looks cool!

Issue #790, opened 2017-11-28
Get rid of `MapView` scale by using the floating point zoom


monsieurtanuki:
## Issue Type
[v] Improvement

## Description and/or steps/code to reproduce the problem
Now that we have floating point zooms (instead of integer zooms), we could get rid of the notion of `MapView` _scale_.
For the moment, the scale always equals 1.0f, except during pinch gesture where it reflects the delta zoom in/out. The idea would be to compute directly the next zoom and forget the notion of scale:
* The code would be simpler and easier to maintain.
* The behavior (especially the tiles display) would be the same during pinch gesture and when the gesture stops
* The best possible tiles would always be displayed, instead of a scaled version of the tiles that where on screen at pinch gesture start.

Please feel free to comment!

### Version of osmdroid the issue relates to:
6.0.0

monsieurtanuki:
Additional thoughts if we get rid of the notion of scale:
* we won't use it on Canvas and as a result we may avoid related problems about hardware acceleration
* I think there's a pending issue about "some layers being scaled when some others are not", which displays smaller/bigger marker pop-ups. We would solve that too

monsieurtanuki:
Could solve #595 too.

secaw:
Sounds great! Would probably reduce some CPU overhead too, right?

monsieurtanuki:
I guess so. And in the supposedly common case of a 0 degree map orientation, we could get rid of the whole canvas manipulation (save / rotate / scale / restore).

monsieurtanuki:
I can tell you, so far it looks great! Still working on it.

secaw:
Sounds promising - really looking forward to this!

spyhunter99:
merged

MKergall:
After all these changes related to "floating point map scale", the overall result is wonderful. 
Seeing for instance the marker clustering operating during the pinch zoom gesture (without any code change needed on my side), that's really great. 

@spyhunter99 & @monsieurtanuki Do you see something else absolutely required before releasing a new version? (could V6.0 land under the christmas tree?)

spyhunter99:
@MKergall that can be arranged! depends on what you want to do for the zoon button overlay

monsieurtanuki:
@MKergall I'm glad you enjoyed the new pinch gesture!
Regarding pending issues I have an interest in:
* #817 is minor but could be quickly fixed.
[...]

PR #908, opened 2018-01-29
Just removed `MapTile`, using `MapTileIndex` instead

monsieurtanuki:
Tons of classes are impacted.
This is only the first step in the fixing of the "slight pause".


spyhunter99:
umm wow.  Would it have made more sense to just merge your changes from MapTileIndex into the old MapTile class?

monsieurtanuki:
@spyhunter99 You're right, it looks a bit excessive. I just started by saying "in this particular class we don't need a `MapTile` object, just a `long` index", and ended up changing 52 classes...
Nevertheless:
* the `MapTile.expires` field was almost never used as is (only in the "Browse the cache" demo)
* we won't need to create an object and now just use a `long` index which is good news, memory- and performance-wise
* the only limit I see is zoom 29, which is more than enough in more than 99.99999% of the cases

Anyway the game is not over, I still have to fix the "slight pause effect" (which I don't really see on my smartphone btw, but I'm convinced it has to do with the sync part that calls the async.

spyhunter99:
expires is used, it's inserted in to the sqlite tile cache which is then used to prioritize deleting stuff from the cache

monsieurtanuki:
@spyhunter99 Oh sure there's an expiration field in the database.
But it is useless in `MapTile`. When we get a tile from the database, we can know if it's expired or not. Then we store it in memory with an up-to-date or expired flag put in `Drawable`.

spyhunter99:
it was used in the MapTile provider chain. Load from cache, if it's expired and we are online, download it again. Is that gone?

monsieurtanuki:
> Load from cache, if it's expired and we are online, download it again.

Correct. We query from the db both the bitmap and the expiration date. If the expiration date is in the past, the bitmap is flagged as expired. If not, it's flagged as up-to-date. And there's no `MapTile` involved in that process, just the primary key (source + xyz) for the `SELECT` db statement.

spyhunter99:
mm ok, what are your thoughts on postponing this pr and the slight pause ticket until after 6.0.0 is cut?

monsieurtanuki:
No objection.

monsieurtanuki:
Now, regarding memory cache, we have:
* a protection of the currently displayed tiles and the corresponding tiles for zoom levels +-1
* an async call for new tiles lookup

@miguel-porto Does it fix the slight pause effect?

miguel-porto:
I'm afraid, it doesn't fix... I've got a video showing how it behaves with the code from this branch: [slight-pause](http://flora-on.pt/slight-pause-2.mp4)
What is strange is that the slight pause now happens everytime I zoom in in the same place, whereas before it only happened on the 1st time I zoomed in (then the tiles were already in memory, so the next zoom-ins were smooth). So I think somewhere along this, there was some problem with the overshoot cache. BTW is the overshoot cache still used, with this new approach?

monsieurtanuki:
I think I know why there's the slight pause. Actually when you change "integer zoom levels" I perform a kind of garbage collection, but in a sync way. I should do this async, or maybe one tile at the time. It's not a problem of overshoot cache, which doesn't exist anymore (I mean, in this specific PR). Thanks for testing and sorry about not having fixed that yet. I'll work on that from Sunday evening (CET). As agreed with @spyhunter99 it's ok if it's not part of the "official" 6.0.0, but I'd like to get rid of this issue soon.

miguel-porto:
No problem, this is a minor issue. Thank you for the efforts.
I'm not sure if it is a good idea to get rid of the overshoot cache. I mean, it could live side by side with the tile protection. Instead of only protecting the 3 zoom levels always, depending on the overshoot cache setting we could have more tiles left in memory. The 3 zoom levels might not be enough for someone (like me) who's deep zooming in and out all the time in the same region. But with the overshoot, I could control that. The tile protection would be at least 3 zoom levels, but we could let it grow more than that. When, in a past version, I set the overshoot to the absurd number 300, then I got really smooth zooms all the way down & up!

miguel-porto:
Oh, I just found an exception happening sometimes using this PR code:
```
 java.util.ConcurrentModificationException
                                               at java.util.HashMap$HashIterator.nextEntry(HashMap.java:787)
                                               at java.util.HashMap$KeyIterator.next(HashMap.java:814)
                                               at org.osmdroid.tileprovider.MapTileCache.garbageCollection(MapTileCache.java:85)
                                               at org.osmdroid.views.overlay.TilesOverlay$CacheTileLooper.finaliseLoop(TilesOverlay.java:318)
                                               at org.osmdroid.util.TileLooper.loop(TileLooper.java:45)
                                               at org.osmdroid.views.overlay.TilesOverlay$CacheTileLooper.loop(TilesOverlay.java:303)
                                               at org.osmdroid.views.overlay.TilesOverlay.protectDisplayedTilesForCache(TilesOverlay.java:182)
                                               at org.osmdroid.views.overlay.DefaultOverlayManager.onDraw(DefaultOverlayManager.java:124)
                                               at org.osmdroid.views.MapView.dispatchDraw(MapView.java:1136)
more...
```

monsieurtanuki:
The overshoot cache can work only in a LRU like context (which tiles are we supposed to get rid of otherwise?). And I'm afraid the LRU mechanism may be skewed by things like `MapTileApproximater`.
That being said, we can ignore this PR and fix the slight pause effect by removing tiles from memory _in an async_ mode. Still a tricky part as concurrency exceptions may happen as you know!

Then, we may later create another issue - as an enhancement - more specific about how we would like the cache to work, and not about a bug to fix.

monsieurtanuki:
I think I'm close to something good, putting async here and there, for instance on `Bitmap` recycling. Still testing.

monsieurtanuki:
I'll push the (ultimate?) fix tomorrow.

monsieurtanuki:
Ok guys, this latest version might be the good one:
* stronger java concurrency compliance
* async bitmap recycling

Please test it and tell me what you think.

spyhunter99:
so i ran into a strange issue whereby the wrapped map did not display and momentarily flicker tiles, however it was a gray grid most of the time.
image
image
scrolling east (panning towards the left) caused all tiles to disappear

The minimap now flickers a lot, but the main map does seem improved, aside from the disappearing world

monsieurtanuki:
I just experienced the gray tiles, will fix that probably today.
Cannot reproduce the minimap flickering though.

monsieurtanuki:
I have good news: I managed to reproduce the flickering. On the main map though.
My educated guess: a 1<<zoom modulo to enforce in `MapTileIndex` for x and y, especially in multi world / intensively scrolled map.

monsieurtanuki:
Fixed: it was a world replication side effect.

miguel-porto:
This PR seems to fix the hang-on-detach issue! I need more tests, but so far so good.
However, the slight pause effect is still there...

spyhunter99:
confirmed that the world replication effect is fixed. i also took a quick video. On mapnik, it see a momentary full screen flicker when zooming across integer boundaries again. It doesn't appear very often on other tile sources. The minimap's flicker is also greatly reduced.

https://youtu.be/4gVVjjqyxJs

miguel-porto:
Talking about grey tiles, what do you think about these grey tiles here, that appear and disappear?
[grey-tiles](http://flora-on.pt/grey-tiles.mp4)
Note that this always happened, it's not new. Don't know in which situations it happens, but seems to be when there are no cached tiles for many zoom levels. Note that the usual grid is not displayed, so it's a different thing.

miguel-porto:
I mean, these don't seem to be "grey tiles", they seem to be portions of the screen that fail to render. I think they don't quite match the tiles.

spyhunter99:
@miguel-porto do you think this is tile source specific?

spyhunter99:
````
I/TestRunner(  596): java.lang.IllegalArgumentException: MapTileIndex: Y (4) is too big for zoom (2)

I/TestRunner(  596): 	at org.osmdroid.util.MapTileIndex.throwIllegalValue(MapTileIndex.java:69)

I/TestRunner(  596): 	at org.osmdroid.util.MapTileIndex.checkValues(MapTileIndex.java:61)

I/TestRunner(  596): 	at org.osmdroid.util.MapTileIndex.getTileIndex(MapTileIndex.java:20)

I/TestRunner(  596): 	at org.osmdroid.views.util.OpenStreetMapTileProviderDirectTest.test_getMapTile_not_found(OpenStreetMapTileProviderDirectTest.java:61)
````

**Comment by monsieurtanuki** [comment]
@spyhunter99 Thank you for extracting the error log. I just fixed it.

**Comment by monsieurtanuki** [comment]
@spyhunter99 I don't know how offensive your https://youtu.be/4gVVjjqyxJs is supposed to be, but Youtube rejects it and I cannot see it.

**Comment by miguel-porto** [comment]
@spyhunter99 I don't think it's source specific, but I can try other sources. However, I really think it is not a tile issue (those tiles are cached), but a redrawing issue, in which some portions of the screen are not invalidated? I suspect this because the shape of the grey areas is often like an L, irrespective of the region in which I'm zooming out.

**Comment by spyhunter99** [comment]
hmm what the hell.

**Comment by miguel-porto** [comment]
But this seems to happen only when you're offline and the only cached tiles you have are from very low resolutions and you're zooming out from a high zoom (so they're very amplified) -  I think you might be able to reproduce it. It's not a very big issue. Oh, and it only happens on zoom out, as if the app forgot to draw the marginal tiles.

**Comment by spyhunter99** [comment]
memory usage is slightly higher on this branch, but the performance increase is worth it IMO. It's running about 20MB higher than master. Master seems to be running at 100MB at the moment but stable

**Comment by monsieurtanuki** [comment]
@miguel-porto You should create a specific issue about your L-shaped gray tiles.

@spyhunter99 To be merged, then?

**Comment by miguel-porto** [comment]
@monsieurtanuki  Yep, will do it, and make a better video showing the problem, also.

**Comment by spyhunter99** [comment]
@monsieurtanuki i'm currently binary searching our commit history looking for when slight pausing first occurred. This won't help the christmas tree effect. I kind of feel like you did a lot of work that may not have been necessary.

here's what i have so far...
7eeeeb8 smooth floating zoom with approximateor
0990b70 zoom 29 smooth oct 1st
daa22b0 merge 834 jan 5 smooth
e3bf97 jan 10 smooth
50c275e jan 14 smooth
893dd5c jan 15 smoothish
8953af1 jan 17 pausing 
6e75e4a jan 21 reasonable
7738dff jan 21  pausing present 

i may have been at fault for this. still trying to narrow it down..

**Comment by miguel-porto** [comment]
But the christmas tree effect is fixed already (don't remember exactly at which commit).

**Comment by spyhunter99** [comment]
Yup the christmas tree effect is gone, however i'm now seeing flickering when crossing int zooms and slight pausing. It's the pausing that's bothering me.

**Comment by miguel-porto** [comment]
I think that's because the overshoot cache is not used any more (I mean the slight pause). I also feel that the code in this PR more frequently shows the slight pause effect than previous versions. I think that's because of the overshoot cache, which now seems to be ignored (@monsieurtanuki correct me if I'm wrong).
@monsieurtanuki what if this PR is split in two, one for fixing the hang-on-detach, and the the rest to be thought later?

**Comment by spyhunter99** [comment]
so the pausing i'm seeing is a slight jitter on zooming in. I've narrowed down the commit to 2f511b9 from jan 17, fix the zoom to bounding box demo. This has some minor changes in map view and projection. Perhaps they are related

**Comment by monsieurtanuki** [comment]
@spyhunter99 Hard to believe the bounding box limits - which are not even used in the default demo - could have an impact on the memory cache.

@miguel-porto Correct, the overshoot cache is not used anymore. As discussed in the "async pre-cache" issue, the solution could be to have a composite cache: partly LRU with overshoot, partly with zooms +-1 protection. But the `overshoot=300` 
statement stinks!

**Comment by miguel-porto** [comment]
But I'm telling you, those were the smoothest zooms ever! Everything was kept in memory! And as far as I remember, I had no out of memory problems during that time.

**Comment by monsieurtanuki** [comment]
@miguel-porto I'm currently coding something to take into account the overshoot. Should be OK by tomorrow.

**Comment by monsieurtanuki** [comment]
@miguel-porto Should be OK right now ;)

**Comment by spyhunter99** [comment]
Just tested it, I see a rather significant pause/lock up when using the built in zoom buttons, but only when zooming in. Zooming out is fine. Same goes for pinch zoom. Zoom in has around 1/4 sec pause. I can take a video if needed

**Comment by monsieurtanuki** [comment]
@spyhunter99 I can see a pause when zooming in, but once you're on a specific region there's no pause when you zoom in or out: do you see the same behavior?
1/4 sec? That's a lot! I'll try to use an emulator (supposedly slow) to profile and detect the performance bottleneck.

Thank you guys for running tests, and sorry about their frequency.

**Comment by spyhunter99** [comment]
That was a guestimate, it's quite noticable though. I'm seeing this on hardware on every zoom in

**Comment by spyhunter99** [comment]
profiles says there's a lot time spent on the ui thread. it looks like `saveTile` method of sqltilewriter. Looking at optimizing this now

**Comment by monsieurtanuki** [comment]
@spyhunter99 `SqlTileWriter.saveTile` obviously deserves some optimization:
````java
List<Byte> list = new ArrayList<Byte>();
//ByteArrayBuffer baf = new ByteArrayBuffer(500);
int current = 0;
while ((current = bis.read()) != -1) {
	list.add((byte) current);
}

byte[] bits = new byte[list.size()];
for (int i = 0; i < list.size(); i++) {
	bits[i] = list.get(i);
}
````
More specifically, the `ArrayList<Byte>` to be populated byte by byte up for bitmaps around 50Kb is probably not the best choice.
In another project I wrote a piece of code similar to `ByteArrayBuffer`; I'll see if it's worth integrating it.
And I noticed there are similar pseudo-ByteArrayBuffer codes to optimize.

But I don't think that should impact strongly the slight pause effect in all cases. If I keep zooming in and out on the same area, all the tiles are either in the database or in memory, therefore `SqlTileWriter.saveTile` isn't called anymore.

**Comment by InI4** [comment]
First loop only seems to be there to count the size? Could rather be a buffered copy into an 
https://docs.oracle.com/javase/7/docs/api/java/io/ByteArrayOutputStream.html
then `byte[] bits = ByteArrayOutputStream.toByteArray()`.

**Comment by monsieurtanuki** [comment]
@InI4 Yes, something like that, possibly with byte[ ] versions of read and write. Actually the first loop does not only count but also read each byte from the input stream.

**Comment by spyhunter99** [comment]
cool, on it

**Comment by monsieurtanuki** [comment]
Just wondering: there might be an improvement with HashMap<MapTileIndex, ...> with a custom hash code, something like zoom + x % 2^13 + y % 2^13, which would really discriminate tile indices.

**Comment by spyhunter99** [comment]
After making the change, the pausing is noticeable reduced.

I'm seeing lots of the following in the log
`02-10 08:45:06.713  1612  1677 W Bitmap  : Called reconfigure on a bitmap that is in use! This may cause graphical corruption!`

Also, probably a separate issue, when using mapnick and landing on zoom 20 or so, the tile provider chain loops forever trying to load the tile. I recall there being a ticket for this at one point

**Comment by spyhunter99** [comment]
optimizing the hash function is probably like shuffling the deck chairs on the titanic.

I'm going to start walking through diff's from this branch to the last known commit to master when the zooming was smooth. There's obviously something going on where. There were a few changes related to synchronization between background threads and the ui thread. I'm just guessing but it could be related to that

**Comment by spyhunter99** [comment]
so interesting performance comparison.
master

 - with scale tiles on, performance is reasonable
 - with scale tiles off, very slow zooming in

this branch

 - with scale tiles on, performance is reasonable
 - with scale tiles off, performance is reasonable

ya know what. i'm ok with the performance on this branch now. it's definitely better than master. I'll start going through the code changes now... (good work @monsieurtanuki!)

@monsieurtanuki any other changes that you want to make before this is merged?
@miguel-porto any other comments?

**Comment by monsieurtanuki** [comment]
> optimizing the hash function is probably like shuffling the deck chairs on the titanic.

@spyhunter99 You made my day!

Well, I started coding it anyway. Then it appeared that it was not only perhaps useless - as you wittily noticed - but also counter productive - as I would be forced to create object instances instead of relying on  the `Long` / `long` symbiosis. OK, you won.

Nevertheless, I think I know what could be improved. Again. But this time I think it's the good one!
The problem when crossing the integer zoom boundaries is that when we loop on all tiles to be displayed, none of them are in memory and therefore we need to use the providers in order to retrieve them. That's the "slightly slow" part, because either we run a sync code that is potentially a bit slow (and surely includes `synchronized` statements) or we run an async code in order to do that. For each tile. If we need 16 tiles, we sync'ly create 16 new `Runnable`s and run them async'ly.
My solution would be, in each tile display loop, to store the needed tiles, and to run at the end of the loop a unique `Runnable` that takes into account the 16 tiles (in the present case).

File: osmdroid-android/src/main/java/org/osmdroid/util/MapTileIndex.java [hunk 1]
```diff
@@ -15,14 +13,11 @@
 
 public class MapTileIndex {
 
-    private static int mMaxZoomLevel = TileSystem.primaryKeyMaxZoomLevel;
+    public static int mMaxZoomLevel = TileSystem.primaryKeyMaxZoomLevel;
```
spyhunter99: this whole class is very interesting. I like it!

spyhunter99:
@monsieurtanuki If i'm understanding your comments on the map tile index, does this mean that all Mobac and similar based sql archives are now no longer valid because of the index change? or does this only effect the "cache" database?

spyhunter99:
ha we might have to rename this project monsiuerdroid by the time this is done.

hmm regarding the threading...mmm. How did the previous setup work? I know the tiles overlay had a looper/handler setup that would use the tile providers to pull the needed tiles on demand from where ever and that the tile provider itself was multi threaded background tasks.

Can we somehow avoid synchronized calls on the tile looper?

monsieurtanuki:
@spyhunter99 The new index computation in `MapTileIndex` only affects the _memory_ cache.
And it also replaces `MapTile` every time it was used as a (zoom, x, y) container.
But nothing changed wherever a (zoom, x, y) combination was _stored_ (in database for instance): the old indices are still used, therefore everything should work the same.
We just have now two ways of computing (zoom, x, y) as a `long`: the new one that is obviously bijective, and the legacy one. 
Should we upgrade all legacy indices to the new version: maybe one day, if we drastically change something about the database anyway (e.g. one table per zoom level / per source). Do we have to? absolutely not. Though two different ways of computing a `long` index from (zoom, x, y) can be confusing.

monsieurtanuki:
> My solution would be, in each tile display loop, to store the needed tiles, and to run at the end of the loop a unique Runnable that takes into account the 16 tiles (in the present case).

The first results look OK (no slight pause anymore - AFAIK), but I need some time to get a clean version of this. 

@spyhunter99 What's the next move then? Merge what we already have, possibly in 6.0.0? Wait a couple of days until I'm done? Merge quietly in 6.1.0?

InI4:
Are you talking about the hashcode() in MapTile or alike? Hashcodes are another hobby of mine.

Both the existing one and the one suggested above by @monsieurtanuki  do not look optimal to me. The worst is, that they are symmetric in x and y. That sends tile(z,x+1,y) and tile(z,x,y+1) to the same slot. That is not good, cause those tiles are very like to be used on the same screen.

I'd suggest something more mixing with multiplicators: `zoom + 0x24A1*x + 0x18289*y `. It might be possible to get the multiplicants tuned even better, when one knows, how many tiles are usually/maximally displayed on one screen and in cache. But the above be better than everything I have seen so far.

monsieurtanuki:
@InI4 Actually I was in the metro when I typed the suggested hashcode - I had to type fast.
To be accurate, what I had in mind was `(zoom<<26) + (x% 2^13 <<13) + y % 2^13`, which I believe make a unique hashcode in the conditions we are using map tiles.
Anyway I was trapped into contradictions: either my HashMap key is a Long and I use its build-in neighborhood with long, or I need to create an instance of a new Object everytime I use a HashMap key. Which now that I'm thinking is probably the case, as Long object are _not_ created _only_ for a small range of values that we don't use.
Do you think it's possible, within a HashMap with Long key, to override the Long hashcode?
But as @spyhunter99 remarked, it's not where we should have significative performance gains.

spyhunter99:
if there's more work required, we can wait for inclusion with v6.0.0

InI4:
@monsieurtanuki that 2nd formula is of course much better! But, internally HashMaps in Java are restricted to arrays and Java arrays are restricted to int indexes. So, so for Java a feasible hashing should map to integers and not longs. This lets this 26/13 look too large. Also Java HashMaps do not use prime number sizes, but powers of 2, this puts the duty of mixing things over to the hashcode() method. Therefore, if you have time, you might still try to consider the formula I give, or augment your formula by a long->int mapping by a % reminder via a large, but int prime number ...

monsieurtanuki:
This version might be the good one!

monsieurtanuki:
> This lets this 26/13 look too large

Not at all! The max zoom is 29, which needs 5 bits. 26+5=31. 31 < 32.

For the moment there's no strong need anyway to change the hashcode as 1- we rely on `Long` that provides its own hashcode and 2- there's a fair chance it won't improve our performances that much regarding all the other things we're doing beside using `HashMap`s.

spyhunter99:
https://youtu.be/N9NktziTHs4

monsieurtanuki:
@spyhunter99 I think we're reaching the performance relevancy limits of async code called in UI thread, and that we shouldn't go further.

I believe the best way to get rid of the slight pause effect is to pre-cache (#930): pre-loading data into memory in order to get the best performances on the UI thread. That will be part of 6.1.0.

We should ignore my latest push and merge just before it, when you said you were OK with the performances.

Btw it looks like I accidentally removed one of your comments. I don't know which one it was and how to put it back. Sorry about that.

miguel-porto:
Sorry I couldn't keep up with these big changes, but I can't wait to test this code!

BTW, the hang-on-detach issue is fixed with this code (well, the two-days-ago-code), I now tested a lot and never happened, so consider the issue closed (it was never open anyway...)

miguel-porto:
There's something broken in this code... most of my cached tiles won't load any more. But if I switch back to a past version, they're there, cached.

spyhunter99:
@miguel-porto are you building this branch from source or using the snapshot on sonatype?

spyhunter99:
OK it's not the compass, but possibly the scale bar overlay, minimap, or mylocation

spyhunter99:
Alrighty, the problem with the StartMapFragment is...... (on master and on this branch)

`mMapView.setTilesScaledToDpi(false);`

That's causing the pausing. There are some minor issues with the compass overlay but those are easy to fix.

monsieurtanuki:
@spyhunter99 I'm not an expert, large trout wise, but I think what you see is a threshold effect.
When `mMapView.setTilesScaledToDpi(false);` the tiles are smaller, therefore it takes more tiles to fill the screen. Remember, the trouble with integer zoom boundaries (assuming the memory cache is empty for the new zoom level) is that we need to refresh the whole screen. We loop on each tile and for each tile we have no quick access to memory cache, therefore we need an additional step to retrieve the tile bitmaps. That's the tally of those additional steps that makes the slight pause possible. And there are more additional steps with smaller tiles.

secaw:
@spyhunter99 Yes, like @monsieurtanuki I also think that it's a matter of the number of tiles that have to be switched at zoom boundaries. IMO the mechanism finally just becomes visible with `mTilesScaledToDpi` set to `false`, because depending on the device there might actually be a multiple of tiles visible / to be switched around when zooming compared to `mTilesScaledToDpi = true`.

A little off topic, but shouldn't `mTilesScaledToDpi = true` be the default now? I remember when starting off with osmdroid, I thought it was unusable as everything was just too small, until I discovered `setTilesScaledToDpi`...

miguel-porto:
@spyhunter99 I'm building this branch from source, and I'm getting that creepy situation I described, lots of high-res tiles are not loaded at all. But I know I have them cached, cause switching back to an older version of this branch shows no problem.

monsieurtanuki:
@miguel-porto I didn't experience what you describe - which looks a bit like @spyhunter99 's video, but I must admit it looks bad.
I'm about to deliver some kind of rollback - probably today.

miguel-porto:
Will upload a video when possible, showing the behaviour of both versions.

monsieurtanuki:
I hope this version will suit you guys (no strange behavior, reasonable performances) and that we'll be able to merge.

miguel-porto:
The problem is gone (but the slight pause still there).
But are you sure you want to throw away the async loading? I didn't think much about it and didn't keep up with the recent changes, but it looks like a good solution to smooth things up. What I would imagine is, all the tiles load async from the cache DB, and while they're loading, what is displayed is the amplified tile from previous zoom (just like what happens when we're online). When a tile is finished loading, then it will replace the amplified previous zoom tile. We wouldn't need pre-cache this way, and the slight pause as well as the grey tiles would go away.

monsieurtanuki:
@miguel-porto Putting `MapTileApproximater` as the first provider in the list would partially do what you suggest. But the problem is that "having none of the tiles to be displayed in memory cache" causes the slight pause, regardless of the tile provider.
I'd prefer to merge what we have so far to 6.0.0, and deal with whatever pre-cache solution we invent in 6.1.0.

miguel-porto:
I agree, this PR is too large already. And as it is, it fixes the `onDetach()` issues, which is a priority IMHO.
Still hoping that the complete async loading will be possible, but that's eventually for another PR.

monsieurtanuki:
@spyhunter99 Ready to merge into 6.0.0?

spyhunter99:
let me test it out. if the studdering isn't fixed, i'm going to start rolling back commits on master until it's fixed, then that will become v6

spyhunter99:
@miguel-porto any idea what the actual fix was for the detach issue?

spyhunter99:
after testing, this is much improved on genymotion, still pausing on zoom in on hardware though

spyhunter99:
alrightly, after backtracking, the pausing looks like it was introduced with commit a622202, the merge of #790 which was merged on Dec 8th. Prior to that commit, the behavior during a pinch zoom, was to due the animation to scale on depend. On touch up, the animation ends, then the new required tiles are loaded.

At that commit, during the pinch animation, tiles are loaded on as they are needed, causing the pause in animation scaling.

That said, this whole pr does a good job of minimizing the pause, but not eliminating it. It's think it's probably the best compromise thus far.

monsieurtanuki:
@spyhunter99 The slight pause effect should be fixed by #930.

More memory is used now probably because the cache is bigger: the displayed tiles are protected now (in order to avoid the xmas tree effect) and so are the corresponding tiles of zoom levels +-1.

We'll introduce more flexibility on which tiles to protect in the memory cache (zoom levels +-2), or when to pre-cache tiles (triggered by a pinch-gesture in progress).

The animation scaling used to cause memory and network issues as it required unexpected far away tiles. And I'm sure the slight pause existed at that moment, but the UX wasn't impacted because the slight pause may have looked like a normal "rubber rollback" effect.

My opinion: explicit OK to merge.

Issue #776, opened 2017-11-16
LRUMapTileCache NoSuchElementException when clearing tiles


Maradox:
## Issue Type

[ ] Question
[X] Bug
[ ] Improvement
[ ] Build system related
[ ] Performance
[ ] Documentation

## Description and/or steps/code to reproduce the problem

Crashreporter:
```
java.util.LinkedHashMap$LinkedHashIterator.nextEntry (LinkedHashMap.java:349)
--
  | java.util.LinkedHashMap$KeyIterator.next (LinkedHashMap.java:366)
  | org.osmdroid.tileprovider.LRUMapTileCache.clear (SourceFile:61)
  | org.osmdroid.tileprovider.MapTileCache.clear (SourceFile:82)
  | org.osmdroid.tileprovider.MapTileProviderArray.detach (SourceFile:81)
  | org.osmdroid.views.overlay.TilesOverlay.onDetach (SourceFile:118)
  | org.osmdroid.views.overlay.DefaultOverlayManager.onDetach (SourceFile:137)
  | org.osmdroid.views.MapView.onDetach (SourceFile:889)
```

Corresponding code (from `LRUMapTileCache`):
```
@Override
public void clear() {
  // remove them all individually so that they get recycled
  while (!isEmpty()) {
    remove(keySet().iterator().next());
  }

  // and then clear
  super.clear();
}
```

### Version of osmdroid the issue relates to:

6.0.0

monsieurtanuki:
What exactly is the Exception you get? It looks like a java concurrency issue between the moment of the `!isEmpty()` and the moment of the `next()`.
I didn't manage to reproduce the bug on my smartphone.
My assumption is that something like that would solve the issue:
````java
public void clear() {
	// remove them all individually so that they get recycled
	while (!isEmpty()) {
		try {
			remove(keySet().iterator().next());
		} catch (NoSuchElementException nse) {
			// as a protection
		}
	}

	// and then clear
	super.clear();
}
````

spyhunter99:
good catch

spyhunter99:
You're positive it was published with a recent snapshot?

secaw:
Yes.
Could it be, that `mWorking` is corrupting the `TileCache.clear()` procedure theoretically? Then the proposed deletion could be it.

secaw:
Update - to be more specific:
In version before above fix, there were actual crashes with `NoSuchElementException`.
[...]

PR #728, opened 2017-10-05
added triming factor to setTilesScaledToDpi

roman-nevr:
added method setTilesScaledToDpi(float trimFactor)

PR #915, opened 2018-02-03
 Introducing an additional tiles scale factor and possible fix for #776

secaw:
Based on what @roman-nevr proposed in #728, it is now possible to set an additional tiles scale factor for map scaling. This now works both for ScaledToDpi and standard size display.

Also removed redundant `MapTileProviderBase.clearTileCache()` from `MapTileProviderArray.detach()` and `MapView.detach()` and moved it to `MapTileProviderBase.detach()` where it should belong. Might be a fix for #776.

### DIFF
--- a/osmdroid-android/src/main/java/org/osmdroid/tileprovider/MapTileProviderArray.java
+++ b/osmdroid-android/src/main/java/org/osmdroid/tileprovider/MapTileProviderArray.java
@@ -78,7 +78,6 @@ public class MapTileProviderArray extends MapTileProviderBase {
 			}
 		}
 
-		mTileCache.clear();
 		synchronized (mWorking) {
 			mWorking.clear();
 		}
--- a/osmdroid-android/src/main/java/org/osmdroid/tileprovider/MapTileProviderArray.java
+++ b/osmdroid-android/src/main/java/org/osmdroid/tileprovider/MapTileProviderArray.java
@@ -70,18 +70,15 @@ public class MapTileProviderArray extends MapTileProviderBase {
 
 	@Override
 	public void detach() {
-
 		synchronized (mTileProviderList) {
 			for (final MapTileModuleProviderBase tileProvider : mTileProviderList) {
 				tileProvider.detach();
 
 			}
 		}
-
 		synchronized (mWorking) {
 			mWorking.clear();
 		}
-		clearTileCache();
 		if (mRegisterReceiver!=null) {
 			mRegisterReceiver.destroy();
 			mRegisterReceiver = null;
--- a/osmdroid-android/src/main/java/org/osmdroid/tileprovider/MapTileProviderBase.java
+++ b/osmdroid-android/src/main/java/org/osmdroid/tileprovider/MapTileProviderBase.java
@@ -65,6 +65,7 @@ public abstract class MapTileProviderBase implements IMapTileProviderCallback {
 	 * Updated 5.2+
 	 */
 	public void detach(){
+		clearTileCache();
 		if (mTileNotFoundImage!=null){
 			// Only recycle if we are running on a project less than 2.3.3 Gingerbread.
 			if (Build.VERSION.SDK_INT < Build.VERSION_CODES.GINGERBREAD) {
--- a/osmdroid-android/src/main/java/org/osmdroid/views/MapView.java
+++ b/osmdroid-android/src/main/java/org/osmdroid/views/MapView.java
@@ -413,7 +413,7 @@ public class MapView extends ViewGroup implements IMapView,
 	/**
 	 * Setting an additional scale factor both for ScaledToDpi and standard size	 
 	 * > 1.0 enlarges map display, < 1.0 shrinks map display
-	 * @since 6.0.0
+	 * @since 6.1.0
 	 */	 
 	public void setTilesScaleFactor(float pTilesScaleFactor) {
 		mTilesScaleFactor = pTilesScaleFactor;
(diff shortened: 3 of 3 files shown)


monsieurtanuki:
@spyhunter99 Wouldn't that make sense to include it into 6.0.0 - I mean more specifically the cache part? Or is it too late?

miguel-porto:
Note that I'm facing frequent silent freezes when MapView.onDetach() is called. Maybe it's related to this.

monsieurtanuki:
@miguel-porto What you're experiencing is specifically with these commits, or mine in #908?
I just created #919 in order to improve memory cache management - and incidentally fix issues. And I don't think my own #908 goes 100% in the right direction.

miguel-porto:
I'm experiencing the freezes both with the master branch and with your PR #908. I didn't try this one here. Sorry I haven't been able to keep up so frequently in testing, as I have another project that is priority. Your #919, at first glance, seems to be the cure to my problems!

secaw:
@miguel-porto Would be absolutely great, if you find the time to test this PR!

@spyhunter99 Like @monsieurtanuki I would recommend to include at least the cache part of this PR in 6.0.0 (if it's still possible) as it's cleaning up the detach chain regarding `MapTileProviderBase.clearTileCache()` in a more logical and safer way. Note that I have this released for a few days now and #776 hasn't reoccured yet as well as some other ANRs seem to have disappeared which cause I wasn't able to identify. So I have good reason to believe that this actually fixes them!

spyhunter99:
If you all are reasonably confident that this solves the problem, then that's fine. I've never seen the slight pause on this end. If it can be replicated, the best thing to do is to use the android device monitor and run a profiling session to find out what methods are causing the pausing. It's really the only way

spyhunter99:
I'm on @secaw 's branch and i'm now seeing cases where by the entire screen flips grey the almost immediately back to tiles and i'm seeing pauses when crossing integer. But...after a short amount of time of pinch zooming and whatnot, whatever the problem was seemed to go away. I occasionally still see a brief instant where the screen is completely gray. Profiling the session gives the following

image

Just looking at the output, a surprising amount of time is spent loading tiles from the database. This could indicate that the in memory cache is too small or that we may have too many threads running that are utilizing the database. Note that the above was running on the genymotion vm.

Profiling by my s5, its quite a bit different. I never saw the flicker, but i did see the pausing while zooming.
image
although there's a lot more tile spent downloading tiles, which is normal. Again, most of the time was spent loading a tile from the sql cache.

Reducing the threads to 4 didn't seem to help the problem at all. It's possible we are just seeing the garbage collector running

spyhunter99:
I'm also seeing the hang on detach issue.  Ugh, gonna check master for the hang later

secaw:
@spyhunter99 That's strange, thanks for testing. I don't think, what you're seeing is caused by this PR though. Did you back check the master again?
This here is not about the slight pause thing and shouldn't have an influence on tile display / loading at all. I cannot reproduce any of that nor the hang on detach issue.
For me it seems to fix #776 though and possibly other related ANRs caused by concurrency issues before, that I believe now had the same source (basically `MapTileProviderBase.TileCache` was cleared too early in the detach chain while it was still open to be written to).
Just wondering, were you building your own `MapTileProviderArray` for the test above?

miguel-porto:
@secaw so far so good. This code here doesn't seem to hang on detach any more. But I'll keep on trying, since the hang problem was unpredictable, so it's better to keep on testing a bit more. I'll post here after I am more certain that this solves the issue.

miguel-porto:
Oh sorry speaked too early... yes the hang-on-detach problem still happens. I'm looking forward to #920...

miguel-porto:
Note that the hang-on-detach is silent, no exceptions are thrown, it gets completely blocked and the activity won't close. So it may be a different issue from #776.

spyhunter99:
Do we have a solid reproducible setup for the hang on detach issue? Is it only with certain overlays? when tiles are still in the download queue? maybe during a zoom animation?

miguel-porto:
I still find it unpredictable, but it seems to happen only after I have zoomed in & out a lot, so, after the memory cache being used a lot. It happens offline (not sure if online also). Note that I have nearly 1 GB of cached tiles. I'm only using Bing maps and SimpleFastPointOverlay (but as far as I remember, it also happens without any overlay).
It only happens when MapView.onDetach() is called. It never happened during the app use, even if I zoom like crazy. I'll post here if I have more clues. But can you reproduce it?

spyhunter99:
@miguel-porto i'm looking at the code for detach now... here's the basic order of operations

 - kill all overlays
 - destroy the tile provider (suspect)
 - purge the tile provider's memory cache (suspect)
 - destroy the tile request handler
 - destroy the projection

i'd suggest putting log statements in between each line to try and isolate the root cause.

spyhunter99:
do we have consensus on this yet? I'm ok with merging this

secaw:
@spyhunter99 Still no #776 or related ANRs since this PR included in published app, so I'm even more positive about this!

@miguel-porto Thanks for testing, too bad it doesn't solve your hang on detach also. Might be good to open a separate issue.
I can't reproduce the hang on detach at all, although I'm switching back and forth the map activity a lot. Maybe we should compare setups, e.g. I'm just using Itemized Overlays, Paths and an additional map overlay. No compass, minimap etc. Did you check deactivating such additional features, if present in your setup?

miguel-porto:
It seems that the hang-on-detach is fixed by #908 ! Still not fully sure, but I'm positive about it.
My setup is pretty simple, just Bing maps and SimpleFastPointOverlay on top of it. However, I have 1GB of cached tiles and work offline, it might have something to do with pushing memory to the limits. But no worries with this for now, because the last commits to #908 probably fixed it!

PR #932, opened 2018-02-10
#624 use network check to prevent the deletion of expired tiles when …

spyhunter99:
…offline (needs some testing)

for @secaw

### DIFF
--- a/OpenStreetMapViewer/src/main/java/org/osmdroid/debug/CacheAnalyzerActivity.java
+++ b/OpenStreetMapViewer/src/main/java/org/osmdroid/debug/CacheAnalyzerActivity.java
@@ -143,7 +143,7 @@ public class CacheAnalyzerActivity extends Activity implements AdapterView.OnIte
     }
 
     private void purgeCache() {
-        SqlTileWriter sqlTileWriter = new SqlTileWriter();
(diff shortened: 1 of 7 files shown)


secaw:
@spyhunter99 I'm not too sure of this one anymore. My 'proposal' in #624 was actually based on a misbehaviour of the cache and/or display system that's been fixed by now. Probably should have closed the issue. In general I really like the automatic cache size management.

In some cases it might be very useful to be able to 'freeze' the cache, which then probably would also need to take into account a 'setUseDataConnection(false)'. On the other hand this introduces the possibility that people actually run out of storage space. This will happen for example with this PR, if they don't update custom implementations to `SqlTileWriter(final Context pContext)`, because the cache will not be cleaned anymore, no matter if online.
Also e.g. the mapsforge provider writes to cache so that it can increase dramatically even when offline. So the topic of a possible 'cache freeze' is probably a bit more complex...

What do you think?

spyhunter99:
@secaw excellent point, i didn't even consider the mapsforge setup. And this is why we do pull requests...
i guess we'll just close this for now and the supporting ticket. no worries

secaw:
Alright, thanks anyways!

Issue #933, opened 2018-02-13
Polygon: Clip path to screen-rect


Maradox:
In order to improve the performance of the polygon, the path of the polygon should be clipped to the screen rect. Thus, only the visible path segments are drawn. 

### Version of osmdroid the issue relates to:

6.0.0

pasniak:
Yesterday I drew ~5000 small Polylines using Polygon.pointsAsCircle() and I get quite a pref hit even if only one  of circles was  on the screen... 

6.0.2-SNAPSHOT

pasniak:
If:
> The fix was to clip the Path into a relatively small square before displaying it: with this fix we could display Polygons with hardware acceleration on

How do I do that?  By overriding `draw`? Is there a code sample?

monsieurtanuki:
@pasniak Here it's @Maradox 's issue, and he PR'ed in #934.
When you say "_The fix was to clip the Path..._" the quote is mine, therefore we're talking about different issues, as there are many about performances with `Polygon`s:
* #934 is an open PR (issue is #933)
* #937 is an open PR (issues are #906 and #890)
* #857 is an open issue
* #820 is an open issue

What I can say now is that:
* "_The fix was to clip the Path..._" is currently implemented in `LinearRing` and `SegmentClipper` for `Polygon`s
* from what I read here and there, the `Path` are notoriously slow with HW acceleration on, cf. the simple and though sluggish "HeatMap" demo
* maybe you'll find good performances with the currently open PRs

monsieurtanuki:
cf. #934

Issue #857, opened 2018-01-09
Polygons' Path optimization


monsieurtanuki:
## Issue Type

[v] Improvement
[v] Performance

## Description and/or steps/code to reproduce the problem
@pasniak in #820 

> I have _hardware acceleration on_ because otherwise map seems sluggish (on tablet).
Recently I added bunch of irregular map areas as Polygons (they are concave and _most of the time they are outside the user's view_) and when I switch on the layer, the map becomes painfully slow.

The goal of this issue is to optimize the `Path` created for a `Polygon`, especially in the cases where the `Polygon`s are not even visible on the screen.
How?
* When hardware acceleration was on, very often Polygons' Path's weren't even displayed (because of Android side effect with hardware acceleration)
* The fix was to clip the Path into a relatively small square before displaying it: with this fix we could display Polygons with hardware acceleration on
* Unfortunately the Paths are not rendered quickly with the hardware acceleration on (plus the clipping process overhead). It's OK to display just a few Paths, but with not that many Paths the display is slow
* When `Path`s are clipped into a square, we must be careful so that the geometry of the Path _inside the clipping area_ respect the geometry of the same Path without clipping. The idea is that, while being careful, we dismiss Points that are useless for the Path display. Which could eventually lead to a 1 point Path, with an initial Point but no other Point deemed relevant display-wise (for instance if all the Points are on Japan and the map focuses on France)

### Version of osmdroid the issue relates to:

6.0.0

monsieurtanuki:
Solved by all the PR related to #1480.

PR #934, opened 2018-02-13
Clip path of polygon to screen rect

Maradox:
Clips the path of the polygon to the screen-rect. I couldn't find any negative side-effect of this approach. Note that it only uses the improvement for `>= KITKAT` due to `Path.Op`.

For my application, which uses two polygons spanning the whole map, it increases the performance from ~15fps to ~55fps.

Issue: https://github.com/osmdroid/osmdroid/issues/933

### DIFF
--- a/osmdroid-android/src/main/java/org/osmdroid/views/overlay/Polygon.java
+++ b/osmdroid-android/src/main/java/org/osmdroid/views/overlay/Polygon.java
@@ -5,8 +5,10 @@ import android.graphics.Canvas;
 import android.graphics.Color;
 import android.graphics.Paint;
 import android.graphics.Path;
+import android.graphics.Rect;
 import android.graphics.RectF;
 import android.graphics.Region;
+import android.os.Build;
 import android.view.MotionEvent;
 
 import org.osmdroid.api.IGeoPoint;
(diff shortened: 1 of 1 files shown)


monsieurtanuki:
Congratulations for improving the performances! I guess I could have used `Path.Op` in `SegmentClipper`. Perhaps to be used in #857.

My remarks regarding the code: I would instantiate `clipPath` once in order to avoid an `Object` creation in the UI thread (`draw`).

Could this work also with `Polyline`s?

Unfortunately we have no way to unit test `Polygon`s, and I'm not able to test this push.
My other concerns are that there might be side-effects:
* when map orientation is different from 0, as the clip rect you use always has smaller dimensions than the oriented version. A solution would be to consider the clip path to be a circle centered on the clip rect center, with a radius of max(width, height) / 2
* with the path outline border, for instance if the path outline border is 10px wide, don't you display it when the path is a rectangle [-101, -101, 601, 1101] and the clip rect is [0, 0, 500, 1000] (expected result: no path border displayed)

MKergall:
We already tried this Path.op clipping solution: see #353 full thread. 
We got a working solution, handling properly map rotation by clipping with: mapView.getIntrinsicScreenRect(mClipRect)

But I also got very (very) bad performance when using the emulator, so I gave up. 

Before pushing your commit in the main branch, I think it must be tested extensively (HW acc. on and off, devices and emulators, various API levels).

monsieurtanuki:
@Maradox Given the recent tremendous improvements of Polygon/Path drawing performances (that also include the notion of "clip rect"), the previous unsuccessful tests in #353 and the code conflicts with the up-to-date version, it makes sense to close this pull request, I'm afraid.

Please feel free to run performance tests on Polygons/Paths with the latest osmdroid code, and, if needed, to create a new pull request that would include code that uses the `Path.Op` you suggested, but only as an option (that wouldn't be the default mode).

Issue #906, opened 2018-01-28
Esri Shape file support


spyhunter99:
## Issue Type

[x] Improvement

## Description and/or steps/code to reproduce the problem

Support importing/convert an esri shape file to osm style overlays

spyhunter99:
I found an ASF compatible library and at least reads the important fields of shape files. I was able to get it imported and was able to convert a low res country border shape file.  This produced 250+ polygons, many with holes. There was a few rendering artifacts that @monsieurtanuki may be interested in, looks like it's isolated to polygons + date line rendering at a low zoom level.

Panning performance was also...poor to say the least. In genymotion it's rendering at about 1 fps. It's slightly better on real hardware but it's still rather poor.

related tickets 
#857 #820 #770 #890

I think this highlights that we could do a better job of optimizing what needs to be rendered and what doesn't and to run point reduction based on what's on screen + the zoom level.

spyhunter99:
https://www.youtube.com/watch?v=Ly7tZ_8LG1s

spyhunter99:
Sample file set is here: http://thematicmapping.org/downloads/TM_WORLD_BORDERS_SIMPL-0.3.zip

spyhunter99:
![image](https://user-images.githubusercontent.com/1975321/35485642-a6063f5a-0430-11e8-9e28-94321995701d.png)
![image](https://user-images.githubusercontent.com/1975321/35485650-b9946e02-0430-11e8-9ef7-8f580f576d95.png)
![image](https://user-images.githubusercontent.com/1975321/35485653-cb777bdc-0430-11e8-9243-2526d0b56eaa.png)
![image](https://user-images.githubusercontent.com/1975321/35485664-dbc72a50-0430-11e8-9abd-66b810a721f0.png)

after profiling, it looks like most of the time is spent in SegmentIntersection.divisionByZeroSideEffect

spyhunter99:
I also found a dbase reader so that we can attach the meta data with each shape. 

![image](https://user-images.githubusercontent.com/1975321/35486970-fb5e4fb6-0443-11e8-9a72-a292d253bdf8.png)

spyhunter99:
this issue depends on #910

monsieurtanuki:
> This produced 250+ polygons, many with holes. There was a few rendering artifacts that @monsieurtanuki may be interested in, looks like it's isolated to polygons + date line rendering at a low zoom level.

[...]

Issue #770, opened 2017-11-10
Add a getBoundingBox method in Overlay class


MKergall:
## Issue Type

[ ] Question
[ ] Bug
[X] Improvement
[ ] Build system related
[ ] Performance
[ ] Documentation

Purpose: allow generic objects to compute bounding boxes.
One key usage will be in FolderOverlay:

1. We will be able to implement an optimized FolderOverlay.draw method, handling culling based on bounding boxes. We can get massive performance improvement when displaying big structured overlays (a common use case when loading KML/GeoJson files).
2. The bounding box is needed to zoom on a specific overlay.

Implementation: Overlay.getBoundingBox could be either abstract (imposing all derived classes to implement it), or return a default bounding box planet-wide, or return null. Not sure yet what is the best choice.

spyhunter99:
there are some related algorithms in the geopackage source. Everytime something is added to a overlay that contains multiple items, it runs a function to expand the bounds of the parent overlay as needed. Fairly simply to port

MKergall:
The algorithm for FolderOverlay is very simple to implement. I already have something similar in my KML stuff. 
My request is just: can we add this getBoundingBox method at Overlay level, or not.

monsieurtanuki:
My two cents:
* I would implement `getBoundingBox` as null by default
* It might be tricky to dismiss the display of some `Overlay`s because of their "beyond the current map" bounding box, if you take the example of `Marker`s slightly outside of the `MapView` but to be displayed with a big icon that should appear (partly) on the map. Cf. a code I edited in `ItemizedOverlay.onDrawItem` that now involves `Rect.intersects`

spyhunter99:
@MKergall you don' t have to ask for permission. Of course we can. Are you writing this or is this just a good idea kind of thing and you're looking for someone else to write it?

spyhunter99:
@MKergall ping

spyhunter99:
working this now, it makes a huge difference!

PR #937, opened 2018-02-17
Feature/#906 ESRI ShapeFile reader + Bounds checks on drawing lines, polygons and markers

spyhunter99:
New module, ESRI Shape Reader. This converts shape file contents into polygons with the associated metadata attached to the infowindow popup. This is actually the smallest of the changes on this branch.

The large changes is optimizations to Polyline, Polygon, Marker, and Itemized Icon Overlay. As suggested by @MKergall in #770, this uses the bounding box check for those mentioned classes in the on draw method. If there is no overlap of the bounds of the overlay with the view's bounds, then skip the draw cycle.

Finally, assuming the line or polygon is within bounds, the point reduction algorithm is ran before drawing. At numerically low zoom levels, this usually reduces small scale polygons down to a triangle and it becomes more and more accurate the further in your zoom.

Impacts:
 - Bounds check: At numerically high zoom levels with a hundreds of geographically dispersed polygons, this has a huge performance boost (HUGE).
 - Point reduction: This has a noticeable impact at numerically low zoom levels
 - Polygons now have a "minimum zoom" level setting to prevent drawing a very low zoom levels. Default is 0 of course but it can be set to bypass drawing if needed for certain applications

Current open issues
 - Bounds check with info window. When the item scrolls out of view, the info window appears to cling to the edge of the screen. I did attempt to auto hide it but ran into some issues with it. I'm not sure what "right" looks like here.
 - When looking at the world at zoom <= 3, the bounds checks are skipped entirely due to world replication issues and no way to check if there are multiple worlds or not.

How to test
 - see #906, there's a download link for a shape file for country borders. Unzip it and copy the contents of the zip to somewhere on your device.
 - Open the app, more samples, Data, Shape File Import, hit the ... button, "Import a shape file"
 - then select the .shp file from the zip contents


spyhunter99:
yup forgot i used an unpublished github project and built it locally
https://github.com/iryndin/jdbf/issues/11

spyhunter99:
image

spyhunter99:
image

spyhunter99:
comparing the screen shots, looks like there are some polygons that are not rendered at the zoom level present. I'm thinking point reduction should be off by default

monsieurtanuki:
> point reduction should be off by default

@spyhunter99 Rather than setting point reduction off by default, wouldn't that make sense to:
* set it on by default
* in case the point reduction doesn't display anything, use a new alternate and very stupid downgraded mode - like a `Polyline` between topLeft and bottomRight?

MKergall:
For infowindow boundingbox check: 
A simple solution is to increase the  bounding box by a ratio in all directions. Maybe half the size: north += height*0.5

(More or less what is done with tiles)

MKergall:
This will also solve similar issues on polylines with huge width, and big marker icons.

spyhunter99:
@monsieurtanuki what i'm seeing is a complex polygon being reduced to just a very small triangle. Also point reduction is not applied to holes currently.

@MKergall good idea i'll see if i can work this in

monsieurtanuki:
@spyhunter99 My assumption was that users prefer a very lousy approximation (that displays something worth zooming in on) (which may look like an ink stain) than no polygon at all on some high zoom levels.
Regarding its accuracy, I agree that it's not very relevant, but we're talking about polygons to be drawn on a very small pixel rectangle, right? I can't say what's better, I'm not a regular polygon user.

spyhunter99:
more sample data https://catalog.data.gov/dataset/military-installations-ranges-and-training-areas

spyhunter99:
perhaps @monsieurtanuki can take a look at this if time allows. i think for polygons and polylines, we should run a point reducing algorithm against the points with zoom is less than 4 or so. using this branch, thew new sample and the linked shape file, draw speeds are less than 1 frame a second. there are tons of overlays and they are all hundreds of points each so naturally some performance issues are expected, however it can probably be better.

at one point, this branch had the point reducer setup built in, but since then theres been quite a bit of work has been done on polyline and polygon and this code is moot now. perhaps this would be better as a separate issue

monsieurtanuki:
@spyhunter99 I guess at least you should separate the ESRI Shapefile Reader code from the Polygon rendering part, and merge the ESRI part.

Then it would be time to open a new issue about Polygon rendering, which would include overlapping, point reduction, minimum zoom level, and a target (like "fast polygon rendering with those inputs data and this zoom level").

spyhunter99:
the biggest change here is that may impact other overlays and existing capabilities are the following

- polygon and polyline - with setPoints and addPoint will now auto set the bounding box the overlay
- folder overlay - with add/remove, the bounds of the overlay is calculated based all child overlays
- marker - bounds is now set to the point of the object
- itemized icon overlay - no changes
- default overlay manager - the bounds is compared with the current view before calling the draw method. it will only be called if the overlay is within, consumes, or overlaps the bounds.

non-impacts

- a new library for reading shape files. it can read plan shapes and metadata files. if metadata is available, all available data is appended to the title and/or description of the converted shape.
- not all shape file formats are supported, such as 3d shapes.
- the code in this library, and the dependency are asf/mit compatible. no gpl issues

since this branch was opened over a year ago, most of the pr is merged related

after merging, the wiki will be updated with further usage instructions

PR #943, opened 2018-02-18
#942 sample layer list

spyhunter99:


spyhunter99:
this change set probably highlights the need for a new method in the overlay class. String getTitle() perhaps.

PR #948, opened 2018-02-22
Fixed a side effect on double-to-long conversions in TileSystem

monsieurtanuki:
Impacted classes:
* `ProjectionTest`: added @Maradox's unit test method `test_conversionFromPixelsToPixels`; removed `delta` parameter in tests when possible
* `TileSystem`: fixed a side-effect in method `ClipToLong` ; deprecated now unused method `Clip`

### DIFF
--- a/osmdroid-android/src/main/java/org/osmdroid/util/TileSystem.java
+++ b/osmdroid-android/src/main/java/org/osmdroid/util/TileSystem.java
@@ -442,14 +442,22 @@ public final class TileSystem {
 	/**
 	 * @since 6.0.0
 	 */
-	public static long ClipToLong(final double value, final double max, boolean wrapEnabled) {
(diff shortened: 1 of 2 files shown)


spyhunter99:
the change looks fine, however i'm not familiar with how ClipToLong is used and what possible side affects this may cause

monsieurtanuki:
@spyhunter99 `ClipToLong` is only used to round mercator geo point coordinates (from `0` to `2^zoom`, as `double`) into pixel (from `0` to `2^zoom - 1`, as `long`). Nothing related to tiles.
Basically, we used to say `longValue = Math.floor(doubleValue - .5)`.
Now it's `longValue = Math.floor(doubleValue)`.
The impact on UI is that `Marker`s will be moved one pixel to the right in half of the cases, or something like that. And the rounding computation is now more accurate.

Issue #930, opened 2018-02-09
Async tile pre-caching


monsieurtanuki:
## Issue Type

- [x] Improvement
- [x] Performance

## Description and/or steps/code to reproduce the problem

Now that the zoom pinch gesture is smoother (no `View.scale` anymore but floating point zoom only), there are some side effects when zooming across integer zoom boundaries:
* @spyhunter99 mentioned "on mapnik a momentary full screen flicker" (#908)
* @miguel-porto mentioned a "slight pause" (#812)

That makes sense, because when we cross the integer zoom boundaries, we need to get at the same time (in the same `TilesOverlay` display loop, playing in the UI thread) a dozen of new tiles, that we need to extract either from the memory, from the database or from the network.

The screen flickering @spyhunter99 experienced is probably the case when all tiles belonging to the new zoom are in the database but not in the memory yet: before we put them from database to memory, we have nothing to display and therefore we display empty gray tiles.
Async'ly pre-caching tiles from database to memory would solve this issue, assuming the speed of access to memory is not the issue there.

The slight pause @miguel-porto experienced would probably be fixed if we implement the async pre-caching. In his case there seems to be an additional issue regarding access to memory, possibly because the tiles he uses are heavier, memory-wise, but the slight pause doesn't seem to appear when the tiles are already in memory, therefore it should be OK too.

What would be the scope of the pre-cache? In the version of `MapTileCache` implemented in #908, we cache the tiles currently displayed PLUS the equivalent tiles of zooms +-1. I guess we could use the same tile spans.

Then there's the question of how aggressive we are in pre-caching: should we only rely on stored data or get data from the network?

### Version of osmdroid the issue relates to:

6.0.0

miguel-porto:
> What would be the scope of the pre-cache?
This is a fundamental question. I'll also try to think about some proposal on this. I think the +-1 zooms should be combined with some kind of LRU (not strictly LRU, but something to think).
In an ideal scenario, one could have an option to choose the best approach for each use case (plain LRU, +-1 zoom cache or other possibility, like the frequency-based LRU).

monsieurtanuki:
@miguel-porto Fair enough. And this flexibility also applies to the memory cache tile list.

I guess we need a new `interface` (`MapTileListDestination`?) with a `MapTileList computeListFromSource(MapTileList source)` method where the list of tiles will be computed from the currently displayed tiles (as a source list).
[...]

Issue #919, opened 2018-02-05
Thread-safe, functionally relevant and async map tile memory cache


monsieurtanuki:
## Issue Type

[v] Improvement
[v] Performance

## Description and/or steps/code to reproduce the problem

As experienced in bugs #776 and #812, the map tile cache management could be improved:
* bullet-proof thread-safe access - in order to avoid #776
* async garbage collecting - in order to avoid #812, slow Activity detach or ANR
* fine grained functional relevancy of the tiles stored in the cache and possible deprecation of LRU - depending on each app's expressed needs

All the new stuff must be explicitly unit tested as it's a core feature.

### Version of osmdroid the issue relates to:

6.0.0

monsieurtanuki:
Following points were fixed by #908:
* bullet-proof thread-safe access - in order to avoid #776
* async garbage collecting - in order to avoid #812, slow Activity detach or ANR

Now the focus is on fine grained functional relevancy of the tiles stored in the cache.

pasniak:
I am on the master as of 2/25 and I observe the below code going into an infinite loop, when I switch between the tile sources:
```
    public void clear() {
        while(!this.isEmpty()) {
            try {
                this.remove(this.keySet().iterator().next());
            } catch (NoSuchElementException var2) {
                ;
            }
        }

        super.clear();
    }
```

![screenshot from 2018-02-25 21-05-32](https://user-images.githubusercontent.com/19778666/36650127-ab816072-1a6f-11e8-97e9-4c4d7423f7bb.png)

monsieurtanuki:
@pasniak In which class do you see this code, which is an old version? Obviously there's confusion: what I see in `MapTileCache` is
````java
public void clear() {
	// remove them all individually so that they get recycled
	final MapTileList list = new MapTileList(mCachedTiles.size());
	populateSyncCachedTiles(list);
	for (int i = 0; i < list.getSize() ; i ++) {
		final long index = list.get(i);
		remove(index);
	}

	// and then clear
	mCachedTiles.clear();
}
````

pasniak:
My code was for decompiled LRUMapTileCache.

monsieurtanuki:
@pasniak `LRUMapTileCache` is now deprecated and not used anymore. Its code is more or less embedded in `MapTileCache` now.

monsieurtanuki:
[...]

PR #956, opened 2018-02-25
Added flexibility when computing tiles to be kept in memory cache

monsieurtanuki:
Not to be merged; this only the first step, though the goal is not very far:
* having a consistent default mode, similar to what we have now (displayed tiles and zooms +-1)
* making it easy to override this mode for explicit needs - for the moment we need to explicitly override `MapTileCache`'s constructor, and I'd prefer something smoother

New interface:
* `MapTileListComputer`: compute a map tile list from a map tile list source

New classes:
* `MapTileListBorderComputer`: compute a map tile list from a map tile list source: its border
* `MapTileListZoomComputer`: compute a map tile list from a map tile list source, on another zoom level
* `MapTileListBorderComputerTest`: unit test class for `MapTileListBorderComputer`
* `MapTileListZoomComputerTest`: unit test class for `MapTileListZoomComputer`

Impacted classes:
* `MapTileCache`: impacted method `garbageCollection` with new classes
* `MapTileList`: moved code of method `populatedFrom` to new class `MapTileListZoomComputer`
* `MapTileListTest`: moved unit test code of method `testPopulatedFrom` to new class `MapTileListZoomComputerTest`

### DIFF
--- a/osmdroid-android/src/main/java/org/osmdroid/tileprovider/MapTileCache.java
+++ b/osmdroid-android/src/main/java/org/osmdroid/tileprovider/MapTileCache.java
@@ -51,7 +51,7 @@ public class MapTileCache {
 	 */
 	private final MapTileList mGC = new MapTileList();
 
-	private final List<MapTileListComputer> computers = new ArrayList<>();
+	private final List<MapTileListComputer> mComputers = new ArrayList<>();
 
 	private int mCapacity;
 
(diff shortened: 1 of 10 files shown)


monsieurtanuki:
Now we can fine tune the list of tiles to be protected in the memory cache, for better performances.

````java
mMapView = new MapView(inflater.getContext());

final List<MapTileListComputer> computers = mMapView.getTileProvider().getTileCache().getProtectedTileComputers();
computers.clear();
computers.add(new MapTileListZoomComputer(1)); // equivalent of displayed tiles at zoom +1
computers.add(new MapTileListZoomComputer(-1)); // equivalent of displayed tiles at zoom -1
computers.add(new MapTileListBorderComputer(1, false)); // 1-tile-wide border of displayed tiles
final MapTileList protect = new MapTileList(); // fixed list of tiles to protect. "fixed" meaning "not linked to displayed tiles"
protect.put(0); // whole world in zoom 0
protect.put(1); // whole world in zoom 1
protect.put(2); // whole world in zoom 2
protect.put(3, 4, 7, 4, 7); // north-east side of the world in zoom 3
computers.add(new MapTileListComputer() {
	@Override
	public MapTileList computeFromSource(MapTileList pSource, MapTileList pReuse) {
		return protect;
	}
});
````

@miguel-porto I know you'd prefer #930 to #919, but that's a first step: being able to keep a custom list of tiles in memory cache - assuming they've been loaded into memory first (as opposed to #930 where they would be automatically asynchronously pre-cached into memory).
Could you please have a go with this version:
* defining a list of tiles for your personal needs (anyway you would need that for #930)
* playing with the map, zooming in and out and panning in all directions
* checking that once you've displayed all tiles of a map, you never experience the slight pause again when going back to that map

Issue #954, opened 2018-02-23
ItemizedOverlay bounds check buffer


Maradox:
`ItemizedOverlay` has a performance feature, where items with the center outside the bounds are not drawn onto the canvas. The code for this check: `mapView.getBoundingBox().contains(item.getPoint()`.
The problem is that this check doesn't respect the width or height of the item. Thus, an item is not drawn when the center is outside the bounds, even if it is partly visible.

**Example**
The third image should still include a part of the icon but doesn't.

![screenshot_20180223-141237](https://user-images.githubusercontent.com/619172/36596121-fdc3e0c0-18a4-11e8-8f0a-1d1a6e487c69.png)
![screenshot_20180223-141247](https://user-images.githubusercontent.com/619172/36596124-001dc728-18a5-11e8-8511-8a54ce3a6672.png)
![screenshot_20180223-141255](https://user-images.githubusercontent.com/619172/36596132-0339739e-18a5-11e8-8ebd-45ee17eb33ea.png)

**Solution**
The check has to include the width and height of the item, for example by offsetting the size of the bounding box by 1/2 of the width and height (depending on the anchor point).

### Version of osmdroid the issue relates to:

6.0.1

Maradox:
I should have a PR by next week.

monsieurtanuki:
The good news is that there's a very simple fix. In `ItemizedOverlay.draw`, we are supposed to draw all items if they are visible.
The current code is:
````java
if (mapView.getBoundingBox().contains(item.getPoint()))
	mInternalItemDisplayedList[i] = onDrawItem(canvas,item, mCurScreenCoords, mapView);
````
The `if` test _must_ be removed.

@spyhunter99 Apparently you recently added the `if` test when trying to fix #182, but it was a wrong idea for several reasons:
1. the `if` clause doesn't take into account the fact that markers have a width and height - which motivated @Maradox to create this current issue because markers disappear when close to the borders
1. `onDrawItem` actually performs the check and bypass the `canvas.draw` step for items that are beyond the visible map (taking into account their sizes)
[...]

PR #957, opened 2018-02-26
Bug/#954 ItemizedOverlay bounds check buffer

Maradox:
ItemizedOverlay now respects the bounding rect of the item itself rather than only using the point coordinates, when deciding if the item should be drawn.

Issue: https://github.com/osmdroid/osmdroid/issues/954


spyhunter99:
well that looked like a painful set of commits. does this work with screen rotation?

monsieurtanuki:
My two cents about screen rotation: the trick is not to consider a width x height display rectangle, but a square whose side is sqrt(width*width + height*height). That's a slightly too large but easy to implement approximation of a rotation circle centered on the map view center, that works for all rotation angles.

spyhunter99:
@monsieurtanuki so your verdict on this? I'm fine with merging as is. If you feel it still needs further changes, speak up

Issue #965, opened 2018-03-13
performance of SqlTileWriter.runCleanupOperation()


shiroyagi:
## Issue Type

[ *] Bug

## Description and/or steps/code to reproduce the problem

When the maximum cache size is exceeded and runCleanupOperation() is called, it can cause the device to freeze for several minutes.  Presumably this is a very cpu and disk intensive operation and can overwhelm the device to such an extent that it stops responding and I can't close the application.  I do have cleanOnStartup=false to avoid purging the cache if I don't have network connection.

Possible ideas for improvement:

1. create an index for COLUMN_EXPIRES
2. perform 'DELETE FROM' in several chunks (eg. 10) rather than one go (perhaps not without an index?)
3. split the database into several (eg. 5) files and just delete the oldest one when the cache size is exceeded (may slow tile search?)
4. don't batch delete tiles at all, just delete/replace individual tiles with newly downloaded ones when full

Maybe some of these have been tried before, or are known not to work?

### If it's a bug, version(s) of android this affects:

5.1

### Version of osmdroid the issue relates to:

6.0.1

monsieurtanuki:
`SqlTileWriter.runCleanupOperation` is commented: _this could be a long running operation, don't run on the UI thread unless necessary_.

It's currently called in 3 places :
* in the `cleanOnStartup` mode, it's called async'ly in a `Thread` with low priority
* when it's called in `saveFile`, it's called sync'ly in two cases: beyond the functional space threshold, or when there's really no space left.

The code itself is:
* check database size
* if it's not too big (by default, under 600Mb), return
* pre-computation with unnecessary approximation, and then a big ineffective SQL `DELETE` statement

In your case, the problems I see are:
* it's called sync'ly
* it's called with a very bad timing - as you're currently saving a new tile into the database
* it's _very_ slow

What I suggest in order to fix it:
* call it async'ly (instead of sync'ly) when you exceed the threshold (when it's first called in `saveFile`), possibly using a singleton in order to avoid running tons of `Thread`s
* as you suggested, split the big `DELETE` statement into a loop of single row `DELETE` statements - something like "while we are beyond the threshold, get the eldest row, delete the row, continue if needed". No more bulk `DELETE`, just a string of harmless single row `DELETE` statements

To be confirmed: I guess the main problem with the current big `DELETE` statement is the eldest tiles selection as there's no index: it must perform a full scan of the SQL table.
[...]

PR #973, opened 2018-03-26
Smooth tile garbage collection, with new SQL index

monsieurtanuki:
New classes:
* `GarbageCollector`: running a task smoothly and asynchronously, only one execution at the same time
* `GarbageCollectorTest`: unit test class related to `GarbageCollector`

Impacted class:
* `SqlTileWriter`: included index creation inside `runCleanupOperation` and one-by-one tile removal; replaced `runCleanupOperation` calls to garbage collection calls

### DIFF
--- a/osmdroid-android/src/main/java/org/osmdroid/config/DefaultConfigurationProvider.java
+++ b/osmdroid-android/src/main/java/org/osmdroid/config/DefaultConfigurationProvider.java
@@ -56,6 +56,9 @@ public class DefaultConfigurationProvider implements IConfigurationProvider {
     protected int animationSpeedShort =500;
     protected boolean mapViewRecycler=true;
     protected short cacheTileOvershoot=0;
+    protected long mTileGCFrequencyInMillis = 300000;
+    protected int mTileGCBulkSize = 20;
+    protected long mTileGCBulkPauseInMillis = 500;
 
     public DefaultConfigurationProvider(){
 
@@ -498,5 +501,35 @@ public class DefaultConfigurationProvider implements IConfigurationProvider {
     @Override
     public short getCacheMapTileOvershoot() {
         return cacheTileOvershoot;
+    }
+
+    @Override
+    public long getTileGCFrequencyInMillis() {
+        return mTileGCFrequencyInMillis;
+    }
+
+    @Override
+    public void setTileGCFrequencyInMillis(final long pMillis) {
+        mTileGCFrequencyInMillis = pMillis;
+    }
+
+    @Override
+    public int getTileGCBulkSize() {
+        return mTileGCBulkSize;
+    }
+
+    @Override
+    public void setTileGCBulkSize(final int pSize) {
+        mTileGCBulkSize = pSize;
+    }
+
+    @Override
+    public long getTileGCBulkPauseInMillis() {
+        return mTileGCBulkPauseInMillis;
+    }
+
+    @Override
+    public void setTileGCBulkPauseInMillis(final long pMillis) {
+        mTileGCBulkPauseInMillis = pMillis;
     }
 }
--- a/osmdroid-android/src/main/java/org/osmdroid/config/IConfigurationProvider.java
+++ b/osmdroid-android/src/main/java/org/osmdroid/config/IConfigurationProvider.java
@@ -393,4 +393,37 @@ public interface IConfigurationProvider {
      * @return
      */
     short getCacheMapTileOvershoot();
+
+    /**
+     * Delay between tile garbage collection calls
+     * @since 6.0.2
+     */
+    long getTileGCFrequencyInMillis();
+
+    /**
+     * @since 6.0.2
+     */
+    void setTileGCFrequencyInMillis(final long pMillis);
+
+    /**
+     * Tile garbage collection bulk size
+     * @since 6.0.2
+     */
+    int getTileGCBulkSize();
+
+    /**
+     * @since 6.0.2
+     */
+    void setTileGCBulkSize(final int pSize);
+
+    /**
+     * Pause during tile garbage collection bulk deletions
+     * @since 6.0.2
+     */
+    long getTileGCBulkPauseInMillis();
+
+    /**
+     * @since 6.0.2
+     */
+    void setTileGCBulkPauseInMillis(final long pMillis);
 }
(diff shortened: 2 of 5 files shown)


InI4:
Questions:
1. What is in the expiry column? If it is a timestamp, shouldn't this be sorted ascending instead of descending?
2. Couldn't at least the select for tiles be batched in the sense, that several are selected at once? The deletion itself could still be one by one ..

monsieurtanuki:
Both remarks are correct, though the impact is light:
* the "ORDER BY" should be "ASC" instead of "DESC"
* we could retrieve the N eldest records (like, 10) and perform the same kind of loop

I'll fix that today.
Beyond that, do you agree with the mechanism?

InI4:
1. Well, the order by would have deleted the freshest tiles, wouldn't it? This is more important.
2. Mechanism is cool, I am looking very much forward to it and hope, index creation will not kill us.
3. When I think of to my own little app, I'd actually even give a time limit, for example 4 weeks, and would delete tiles which are older than 4 weeks without the need of a storage bottleneck .. But this is a new feature ;-)

monsieurtanuki:
> Well, the order by would have deleted the freshest tiles, wouldn't it? This is more important.

It's not that important as in the `SELECT` I already restricted to the _expired_ tiles: I wasn't deleting the youngest tiles, but the youngest _expired_ tiles. Anyway, I'm fixing it.

> Mechanism is cool, I am looking very much forward to it and hope, index creation will not kill us.

In the worst case, the index creation will happen only once. But we should put the index creation not only in the `gc` (for the old apps) but also just after the table creation (for newcomers).

monsieurtanuki:
It should be better now.

About creating the index at table creation too, I don't see any elegant way to do it. And the index is meaningful only in the context of `gc` anyway. I guess we can survive without that feature.

InI4:
Index creation table creation would have one advantage: It would be very fast, since the table is empty.

monsieurtanuki:
The index creation will definitely be faster at table creation time.
Today I'll add a lousy "if count(*)=0 create index" statement just after the "create table if exists".
In the meanwhile any test of the tile garbage collection is welcomed.

monsieurtanuki:
Just found and PR'ed an elegant SQL index creation timing.

monsieurtanuki:
@spyhunter99 Help needed. Doesn't seem to be able to compile, fails with that kind of log:
````
* What went wrong:
Could not resolve all dependencies for configuration ':osmdroid-server-jdk:compileClasspath'.
> Could not determine artifacts for org.eclipse.jetty.orbit:javax.servlet:3.0.0.v201112011016
   > Could not get resource 'https://jcenter.bintray.com/org/eclipse/jetty/orbit/javax.servlet/3.0.0.v201112011016/javax.servlet-3.0.0.v201112011016.orbit'.
      > Could not HEAD 'https://jcenter.bintray.com/org/eclipse/jetty/orbit/javax.servlet/3.0.0.v201112011016/javax.servlet-3.0.0.v201112011016.orbit'. Received status code 502 from server: Bad Gateway
````
Is the answer: "it's just a bad luck error, try later" or is there something we can fix on the osmdroid side?

shiroyagi:
Looks nice, particularly your SQL.  A few thoughts:

1. Would it be worth putting a sleep (eg. 200ms) in your runCleanupOperation() loops?  This will reduce the load on the system and give other higher priority tasks a chance to run (eg. loading a tile from cache).
2. I would still reduce `lastSizeCheck + 300000`: small and often is definitely better (or at least make it tunable)
3. Is it worth checking `db_file.length()` in the loop in case additional tiles are added concurrently?
4. Possibly have the loop always running whilst the map is visible with a longer sleep (eg. 5s) to increase the separation of the cleanup from user interaction (less sure about this).

monsieurtanuki:
@shiroyagi I totally agree with 1 and 2.

About 3: for different reasons we cannot rely 100% on db length anyway (the db file doesn't shrink automatically, if we delete a table record the corresponding index record are deleted as well, and the table record is slightly bigger than just the tile blob). Therefore I think we should apply the Pareto principle and admit the current algorithm is good enough.

About 4: I find elegant to run the gc only when we save tiles, and in offline mode I think it makes sense to keep expired tiles rather than deleting them and displaying nothing. And that offline specificity would have do be dealt with, which would make the code less readable.

About 5: if for some reason the db file shrinks very slowly we will delete much more records than needed, including not expired ones.

I'll work on 1 and 2 by the end of the week.

shiroyagi:
Given the uncertain size on disk have you thought about `SQLiteDatabase.setMaximumSize()` and calling gc() in response to `SQLiteFullException`?

spyhunter99:
I'll have a look at it

monsieurtanuki:
@shiroyagi What should be done when `SQLiteFullException` is caught? I don't know, there's nothing functionally obvious.
For instance, imagine the user-defined database limit is 600Mb, and so far there are 400Mb in the db. And 500Mb left on the disk.
And then I download 500Mb of MP3 on my disk: the disk is full.
What could we expect of osmdroid:
* removal of expired and unexpired tiles - down to which Mb limit? why would osmdroid tiles be considered as less important than the recent MP3 files?
* display of a `Toast`?
* display of a "Cleanup Yes/No" alert dialog?

I think that whatever is related to `SQLiteFullException` should be part of a new issue: the current issue only deals with on-the-fly `gc`.
I guess we should let the developer catch `SQLiteFullException` and decide what to do with it in the context of his own app.
What I can do in the current issue is adding parameters to the tile removal process:
* total tile size to be deleted (now: db file length - max expected db size)
* bulk size (now: 20)
* sleep duration between bulk actions (now: 0)
* including unexpired tiles or not (now: false)

shiroyagi:
Yes, that all that sounds good and anything more can be done elsewhere.

My expectation was that the Android API would provide a better way to understand the size the database when some of it may be cached in memory.  I would hope `SQLiteFullException` relates to `setMaximumSize()` and that something else filling the disk would result in `SQLiteDiskIOException` or some other exception but this would need testing.

Even if it behaves as I hope `SQLiteFullException` is not ideal as by this point new rows are being rejected (a soft limit or `getSize()` would be better).  Still it may be the best we can do and after catching one I would add the rejected new tile transaction to an asynchronous queue to be completed after something else is removed (gc to the trim size):
`SQLiteDatabase.setMaximumSize(Configuration.getInstance().getTileFileSystemCacheMaxBytes())`

monsieurtanuki:
@spyhunter99 Seems to be OK now.

monsieurtanuki:
Just PR'ed a version of `runCleanupOperation` with more flexibility and configuration default parameters.

monsieurtanuki:
@shiroyagi Does this PR eventually solve #965? Can we merge it?

PR #983, opened 2018-04-05
Feature/#977: Sometimes CacheManager#downloadAreaAsync too slow

Tester2012:
Do not calculate all the points for a bounding box in the CacheManagerTask constructor but rather retrieve them one by one.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -43,7 +43,7 @@ gradleFury.version=1.0.14
 android-plugin.version=2.1.0
 android-support.version=23.1.1
 junit.version=4.12
-robolectric.version=3.8
+robolectric.version=3.0
 
 
 # Maven Repository (i.e. Sonatype Nexus Repository Manager) Configuration --------------------------
--- a/gradle.properties
+++ b/gradle.properties
@@ -43,6 +43,7 @@ gradleFury.version=1.0.14
 android-plugin.version=2.1.0
 android-support.version=23.1.1
 junit.version=4.12
+robolectric.version=3.8
 
 
 # Maven Repository (i.e. Sonatype Nexus Repository Manager) Configuration --------------------------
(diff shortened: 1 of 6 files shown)


Tester2012:
@monsieurtanuki , thanks for the review! I have a couple of remarks regarding your comment:

1. > I would make mTiles an IterableWithSize<T>, in order to get rid of the lousy instanceof inside onPreExecute

    How will it work for `CacheManagerTask(final CacheManager pManager, final CacheManagerAction pAction,
                                final List<Long> pTiles,
                                final int pZoomMin, final int pZoomMax)`? `final List<Long> pTiles` cannot be converted to `final IterableWithSize<Long> pTiles`   

2. > consider that your implementation with IterableWithSize<Long> is the only way to deal with BoundingBox+zoom tile sets 

Just confirm - in `public static List<Long> getTilesCoverage(final BoundingBox pBB,
                                              final int pZoomMin, final int pZoomMax)` you're suggesting to use another implementation of `IterableWithSize<Long>` rather than `public static Collection<Long> getTilesCoverage(final BoundingBox pBB, final int pZoomLevel)` To do so `IterableWithSize<Long>` should extend `Collection<Long>` for compatibility.

monsieurtanuki:
@Tester2012 You're right, I was a bit optimistic regarding Java casting.
Nevertheless, the solution is to say that in `CacheManagerTask`- and specifically in `getTilesCoverage` - we eventually don't want to use a `List<Long>` but an `IterableWithSize<Long>` instead:
* if they are populated with your BoundingBox+zoom fix, you use your own implementation of `IterableWithSize<Long>`
* if they are populated with a `List<Long>`, you embed it into something like that:
````java
class MyList<T> implements IterableWithSize<T> {
	private final List<T> mList;

	public MyList(final List<T> pList) {
		mList = pList;
	}

	@Override
	public Iterator<T> iterator() {
		return mList.iterator();
	}

	@Override
	public int size() {
		return mList.size();
	}
}
````

Do you like this idea?

Tester2012:
@monsieurtanuki Your idea looks very good. I'll update my PR according to your suggestions.

Tester2012:
@monsieurtanuki done

Issue #182, opened 2015-10-30
Performance Issue with a lot of Overlays


Maradox:
When using ItemizedIconOverlays with a lot of points (altogether ~1500) the performance decreases significantly (below 60fps on a Nexus 5). Via the profiler I can see that most of the time is required for drawing onto the canvas.

Is that behavior to be expected with that amount of points? Is there something that could be done to increase performance (beside using an extra TileOverlay instead where the points are drawn on)? Skipping items which are currently not visible is not a real solution in my case, since sometimes all Points are visible.

spyhunter99:
have you tried the clustering mechanism osmbonus pack has?

Maradox:
I wanted to avoid clustering. My plan is to rescale the drawables of the OverlayItems according to the zoom-Level where a higher zoom-level increases the size of the OverlayItems. This approach already looks and works very well and has no performance penalty during scrolling on its own.

If there is no solution for the performance problem I may implement a custom TilesOverlay which draws the items on tiles which can be cached (in this case creating the Tiles may be a performance-problem, but I will have to check). Or would something speak against this approach?

Otherwise I may reconsider clustering.

spyhunter99:
I just added an example that lets you add points 500 at a time. on my s5, it runs pretty good up until 3k icons. at 6k icons it's noticeably slow. what kind of device are you running on?

Maradox:
I'm using a Nexus 5. 
I tried your example. With 500 icons it 's still smooth. 1500 is already a bit laggy. 3000 is very laggy.

spyhunter99:
Sounds like your use case uses data that doesn't change very frequently or not at all. In this case, painting the icons right on to the tiles overlay could be a feasible solution. You may also want to look at MapForge, which renders tiles on the fly using OSM vector data. There's a plugin in OSMBonus pack for it.

do you by chance have hardware acceleration turned off?

spyhunter99:
Also, are your icon all the same or are they unique? you can probably save memory usage by using a single icon and reusing it. Using a smaller icon should help reduce the lag too. Another strategy is to remove icons from the overlay when they are off screen and re-add when they become on screen. One can accomplish this by attaching a map listener for zoom and movement events and then handle accordingly. Reducing the amount of stuff in the osmdroid overlays will reduce draw times, even if the item is currently off screen

Maradox:
The overlays don't change frequently but can be changed in the settings. I use about 35 different icons.

[...]

PR #984, opened 2018-04-05
Feature #182: Fastest marker display ever ;)

monsieurtanuki:
The general idea was to get rid of most of the canvas manipulations:
* the canvas scale part became useless when we started using double zooms instead months ago (scale now always equals to 1)
* the canvas rotation part can be prevented most of the time. Before we decided to display, we used to check if the rotated canvas (rotation using save, rotate, restore) matched the marker area. Now we check if the canvas matches the _rotated_ marker area (the rotated `Rect` is computed with ad hoc optimized code).

To my standards the performances are very good.
And I believe I also fixed a "click of the marker" bug in the process.

Demo "More Samples / Data / Itemized overlay at Scale" now has 5K markers (instead of 500) and rotation gesture, for performance test purpose.
One remark: the info window follows the map orientation, but the icon does not. What is the expected behavior?

Demo "OSMapView with ItemizedOverlay" now has rotation gesture, for display and click test purpose.

Potential error: we use class members in order to avoid recreating `Rect`s and `Point`s, and they are used for the marker display AND for the click detection as well. Do you guys think we should play it safe - I mean, we're only talking about a couple of `Rect`s and a `Point` to be created for each display?


OpenStreetMapViewer/src/main/java/org/osmdroid/samplefragments/data/SampleWithMinimapItemizedOverlayWithScale.java
```diff
@@ -62,7 +62,7 @@ protected void addOverlays() {
 		{
 			/* Create a static ItemizedOverlay showing some Markers on various cities. */
 			final ArrayList<OverlayItem> items = new ArrayList<>();
-               for (int i=0; i < 500; i++){
+               for (int i=0; i < 5000; i++){
```
Tester2012: What are you trying to test it with 5000 ?

monsieurtanuki: I test with 5k in order to check that we get the best performances even with a large crowd of markers. The point being to dismiss the markers that won't be displayed.
I assumed it was more or less the purpose of this demo.

Tester2012: You're right, it's written in your commit message.

PR #995, opened 2018-04-14
Bug/#992: Refactored code and cleaned javadoc in class Projection

monsieurtanuki:
Impacted classes:
* `MyLocationNewOverlay`: replaced deprecated `Rect` and `Point` method calls from `Projection` by the `RectL` and `PointL` version
* `Projection`: fixed some javadoc comments; deprecated the version of `toProjectedPixels` still using `1E6 int` coordinates; added an explicit javadoc `@deprecated` comment for every `@Deprecated` annotation; removed useless private methods `getPixelXFromMercator`, `getPixelYFromMercator` and `getPixelFromMercator`; removed int/long duplicate code in the deprecated methods
* `RectL`: added methods `offset` and `union`, inspired from `Rect`'s, in order to be able to deprecate some `Projection` methods

### DIFF
--- a/osmdroid-android/src/main/java/org/osmdroid/util/RectL.java
+++ b/osmdroid-android/src/main/java/org/osmdroid/util/RectL.java
@@ -316,4 +316,40 @@ public class RectL {
                                    final double pCos, final double pSin) {
         return pCenterY + Math.round((pX - pCenterX) * pSin + (pY - pCenterY) * pCos);
     }
+
+    /**
+     * @since 6.0.2
+     */
+    public void offset(final long pDx, final long pDy) {
+        left += pDx;
+        top += pDy;
+        right += pDx;
+        bottom += pDy;
+    }
+
+    /**
+     * @since 6.0.2
+     */
+    public void union(final long pLeft, final long pTop, final long pRight, final long pBottom) {
+        if ((pLeft < pRight) && (pTop < pBottom)) {
+            if ((left < right) && (top < bottom)) {
+                if (left > pLeft) left = pLeft;
+                if (top > pTop) top = pTop;
+                if (right < pRight) right = pRight;
+                if (bottom < pBottom) bottom = pBottom;
+            } else {
+                left = pLeft;
+                top = pTop;
+                right = pRight;
+                bottom = pBottom;
+            }
+        }
+    }
+
+    /**
+     * @since 6.0.2
+     */
+    public void union(final RectL pRect) {
+        union(pRect.left, pRect.top, pRect.right, pRect.bottom);
+    }
 }
--- a/osmdroid-android/src/main/java/org/osmdroid/views/Projection.java
+++ b/osmdroid-android/src/main/java/org/osmdroid/views/Projection.java
@@ -206,7 +206,7 @@ public class Projection implements IProjection {
 	}
 
 	/**
-	 * A wrapper for {@link #toProjectedPixels(long, long, PointL)}
+	 * A wrapper for {@link #toProjectedPixels(double, double, PointL)}
 	 */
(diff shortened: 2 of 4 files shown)


osmdroid-android/src/main/java/org/osmdroid/util/RectL.java
```diff
@@ -316,4 +316,40 @@ public static long getRotatedY(final long pX, final long pY,
...
+        bottom += pDy;
+    }
+
+    /**
+     * @since 6.0.2
+     */
+    public void union(final long pLeft, final long pTop, final long pRight, final long pBottom) {
+        if ((pLeft < pRight) && (pTop < pBottom)) {
+            if ((left < right) && (top < bottom)) {
+                if (left > pLeft) left = pLeft;
```
Tester2012: You could also write it like `top = Math.max(top, pTop);`

osmdroid-android/src/main/java/org/osmdroid/util/RectL.java
```diff
@@ -316,4 +316,40 @@ public static long getRotatedY(final long pX, final long pY,
...
+        left += pDx;
+        top += pDy;
+        right += pDx;
+        bottom += pDy;
+    }
+
+    /**
+     * @since 6.0.2
+     */
+    public void union(final long pLeft, final long pTop, final long pRight, final long pBottom) {
```
Tester2012: You could also write a unit test for this method.

osmdroid-android/src/main/java/org/osmdroid/util/RectL.java
```diff
@@ -316,4 +316,40 @@ public static long getRotatedY(final long pX, final long pY,
...
+        if ((pLeft < pRight) && (pTop < pBottom)) {
+            if ((left < right) && (top < bottom)) {
```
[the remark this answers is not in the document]
monsieurtanuki: Oh, I just copied the code from `android.graphics.Rect` with `long` instead of `int`.
> Shouldn't a check like this `if ((left < right) && (top < bottom))` be done beforehand, when RectL is created?

There's no such thing in the original `Rect` code. When I created the `RectL` class I had to deal with all the osmdroid code history with `Rect`, and it was too dangerous to change the behavior. And I'm not against a `left > right` possibility in order to handle the date line cases, but it's a matter of taste.

osmdroid-android/src/main/java/org/osmdroid/util/RectL.java
monsieurtanuki: I assume Android's code for `Rect` is tested ;)

osmdroid-android/src/main/java/org/osmdroid/util/RectL.java
monsieurtanuki: Just a simple copy of the `Rect.union` code. And I guess the code will be identically easy to read and quick to run in the `Rect` version and in your suggestion.

osmdroid-android/src/main/java/org/osmdroid/util/RectL.java
Tester2012: :)

osmdroid-android/src/main/java/org/osmdroid/util/RectL.java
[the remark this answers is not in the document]
Tester2012: If a Rect with `left >= right`or `top >= bottom` are valid usage cases, then OK.

osmdroid-android/src/main/java/org/osmdroid/util/RectL.java
Tester2012: Sure, that's why it's  a comment and not a task:)

PR #996, opened 2018-04-15
Feature/#974: faster database upsert for tiles

monsieurtanuki:
Impacted class:
* `SqlTileWriter`: replaced the calls to `delete` and `insert` by a single call to `SQLiteDatabase.replace`


spyhunter99:
looks good to me

PR #997, opened 2018-04-16
Fix creating tilepackage from temp directory

kampkb:
Bug fix: runDownloading should not be called if packaging from a temp directory
(serverURL==null && tempFolder.exists())


Tester2012:
@monsieurtanuki can your suggestion be done in a separate PR? Otherwise this PR may stay like this for a long time..

monsieurtanuki:
@Tester2012 Sure.

PR #1002, opened 2018-04-18
bug/#989 - adding an increasing delay in tile fetch repetition

monsieurtanuki:
When for some reason we could not get an up-to-date version of a tile, the tile fetch process went in an endless loop in the search of this up-to-date version.
The endless loop is not good itself, and it's even worse when there's a provider that uses the network: tons of useless network tries.
The fix is to consider that if we can't get the up-to-date version, we wait before trying again. And the delay is increasing.

New classes:
* `Delay`: successive delays container
* `DelayTest`: the unit test class around `Delay`

Impacted classes:
* `MapTileDownloader`: unrelated optimization regarding the instantiation of `TileLoader`; considered `UnknownHostException` as similar to other `IOException`s
* `MapTileProviderArray`: added member `mTileDelays` as a `Map<tile, Delay>` and an exponential backoff delay array `mExponentialBackoffDurationInMillis`; added a check in `getMapTile` in order to delay the next tile fetch; adapted `mapTileRequest*` methods in order to affect the tile delay; created methods `shouldWait`, `nextDelay` and `removeDelay`

### DIFF
--- a/osmdroid-android/src/test/java/org/osmdroid/util/DelayTest.java
+++ b/osmdroid-android/src/test/java/org/osmdroid/util/DelayTest.java
@@ -1,0 +1,56 @@
+package org.osmdroid.util;
+
+import org.junit.Assert;
+import org.junit.Test;
+
+import java.util.Random;
+
+/**
+ * @since 6.0.2
+ * @author Fabrice Fontaine
+ */
+public class DelayTest {
+
+    @Test
+    public void testDelayOne() {
+        final long millis = 500;
+        final Delay delay = new Delay(millis);
+        for (int i = 0 ; i < 5 ; i ++) {
+            check(delay, millis);
+            final long next = delay.next();
+            Assert.assertEquals(millis, next);
+        }
+    }
+
+    @Test
+    public void testDelayMulti() {
+        final long[] millis = new long[] {500, 600, 800, 1000};
+        final long lastDuration = millis[millis.length - 1];
+        long next;
+        final Delay delay = new Delay(millis);
+        for (int i = 0 ; i < millis.length ; i ++) {
+            check(delay, millis[i]);
+            next = delay.next();
+            Assert.assertEquals(i < millis.length - 1 ? millis[i + 1] : lastDuration, next);
+        }
+        check(delay, lastDuration);
+        next = delay.next();
+        Assert.assertEquals(lastDuration, next);
+        check(delay, lastDuration);
+    }
+
+    private void check(final Delay pDelay, final long pMillis) {
+        sleep((pMillis * 3) / 4);
+        Assert.assertTrue(pDelay.shouldWait());
+        sleep(pMillis / 2);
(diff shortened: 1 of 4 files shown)


osmdroid-android/src/test/java/org/osmdroid/util/DelayTest.java
```diff
@@ -0,0 +1,56 @@
...
+        Assert.assertTrue(pDelay.shouldWait());
+        sleep(pMillis / 2);
+        Assert.assertFalse(pDelay.shouldWait());
+    }
+
+    private void sleep(final long pMillis) {
+        try {
+            Thread.sleep(pMillis, 0);
+        } catch (InterruptedException e) {
+            //
```
jayaddison, line 53: Hi @monsieurtanuki - I'm wondering whether this line might relate to some [flaky test results](https://github.com/osmdroid/osmdroid/actions/runs/3548810663/jobs/5960434258#step:5:1354) that appear for [this test assertion](https://github.com/osmdroid/osmdroid/blob/713f680f8e7c56418331415f7dd8adb9b6f43375/osmdroid-android/src/test/java/org/osmdroid/util/DelayTest.java#L42).

From [what I've read](https://medium.com/my-random-thoughts-in-software-engineering/get-rid-of-thread-sleep-e2abf3cadb6), there may be more convenient and reliable patterns to adjust a sleep like this (although some of those introduce additional library dependencies).

Alternatively, I wonder whether a simple approach here (in terms of fewer dependency/logical changes required in the code) might be to wrap the `try...catch` in a a `while` loop that exits successfully if the current time exceeds the `start + pMillis` time.

monsieurtanuki: Hi @jayaddison!
I don't code for osmdroid anymore, but if you PR a fix for a bug you've encountered, I may be able to review it.

jayaddison: Thanks @monsieurtanuki - I don't have a fix for this yet, as I haven't configured a build environment for it, but may do in future.  It mostly depends on how long I decide to keep #1877 open -- because it would be good for CI to pass (reliably) before that could be merged.

jayaddison: I think that I'd misunderstood the problem here:

The flaky test is due to an `assertTrue(shouldWait)` call that fails.  So, according to the `Delay` object, the thread has already waited for `pMillis` or longer: adding an additional `while` loop to add further sleeps will not help that.

My guess at the moment is that this could occur when the test thread is interrupted, and resumes, but perhaps is resumed after a duration of time exceeding expected wake-up time.

jayaddison: Potential theory behind the cause, and corresponding fix, opened as #1879 for review.

Issue #998, opened 2018-04-16
Crash in demos around Bookmarks


monsieurtanuki:
## Issue Type

- [x] Bug

## Description and/or steps/code to reproduce the problem

* go to demo "More Samples / Location / Heading/Compass up"
* go back
* go again to demo "More Samples / Location / Heading/Compass up"
* then it crashes with that kind of log:
````java
E/StrictMode: A resource was acquired at attached stack trace but never released. See java.io.Closeable for information on avoiding resource leaks.
              java.lang.Throwable: Explicit termination method 'close' not called
                  at dalvik.system.CloseGuard.open(CloseGuard.java:184)
                  at android.database.sqlite.SQLiteDatabase.openInner(SQLiteDatabase.java:909)
                  at android.database.sqlite.SQLiteDatabase.open(SQLiteDatabase.java:878)
                  at android.database.sqlite.SQLiteDatabase.openDatabase(SQLiteDatabase.java:699)
                  at android.database.sqlite.SQLiteDatabase.openOrCreateDatabase(SQLiteDatabase.java:714)
                  at android.database.sqlite.SQLiteDatabase.openOrCreateDatabase(SQLiteDatabase.java:707)
                  at org.osmdroid.samplefragments.bookmarks.BookmarkDatastore.<init>(BookmarkDatastore.java:45)
                  at org.osmdroid.samplefragments.bookmarks.BookmarkSample.<init>(BookmarkSample.java:26)
                  at java.lang.reflect.Constructor.newInstance(Native Method)
                  at java.lang.Class.newInstance(Class.java:1656)
                  at org.osmdroid.samplefragments.SampleFactory.getSample(SampleFactory.java:304)
````

### Version of osmdroid the issue relates to:

6.0.2

spyhunter99:
hmm that's strange since there's definitely a close call in onDestroy

monsieurtanuki:
@spyhunter99 It's even stranger because I'm not even supposed to use bookmarks in this demo.

spyhunter99:
with your procedure, did you ever load the bookmarks sample?

spyhunter99:
oh i see the problem, never mind

spyhunter99:
@monsieurtanuki want to retest?

monsieurtanuki:
@spyhunter99 As written in my comment about your commit, it fixed the issue.
Please PR and merge.

PR #1008, opened 2018-04-21
bug/#1005 - SqlTileWriter now uses a static database, in order to avoid locks

monsieurtanuki:
A direct impact is that the app must be restarted in order to have a database path change taken into account.

Impacted class:
* `SqlTileWriter`: `db_file` and `db` are now `static`; added a `static` lock used in the constructor; removed everything in `onDetach` as nobody should close a `static` database

### DIFF
--- a/OpenStreetMapViewer/src/main/java/org/osmdroid/debug/model/SqlTileWriterExt.java
+++ b/OpenStreetMapViewer/src/main/java/org/osmdroid/debug/model/SqlTileWriterExt.java
@@ -1,6 +1,7 @@
 package org.osmdroid.debug.model;
 
 import android.database.Cursor;
+import android.database.sqlite.SQLiteDatabase;
 import android.util.Log;
 
 import org.osmdroid.api.IMapView;
@@ -27,6 +28,7 @@ import static org.osmdroid.tileprovider.modules.DatabaseFileArchive.TABLE;
 public class SqlTileWriterExt extends SqlTileWriter {
 
     public Cursor select(int rows, int offset) {
+        final SQLiteDatabase db = getDb();
         if (db != null)
             return db.rawQuery("select " + COLUMN_KEY + "," + COLUMN_EXPIRES + "," + COLUMN_PROVIDER + " from " + TABLE + " limit ? offset ?", new String[]{rows + "", offset + ""});
         return null;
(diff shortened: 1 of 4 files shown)


monsieurtanuki:
@spyhunter99 The idea of a `static` database could help fix the bookmark crashes #998

spyhunter99:
the danger with this is permissions or if the storage location is external storage and it's unmounted. There won't be a way to refresh the database connection

monsieurtanuki:
> There won't be a way to refresh the database connection

A solution might be to be more aggressive about `Exception`s inside `SqlTileWriter`: if we catch something that says "OK you tried to insert a record but you don't even have a usable database", then we can reset the `db` to `null`.
And if instead of a direct access to `db` inside `SqlTileWriter` we use some kind of `getDb` method that checks first if `db` is null and sets it (like in the `synchronized(mLock)` part) if necessary, we'll have the best of both worlds: a database without locks, and a database that can be refreshed at any time (even more flexible that before this PR).
@spyhunter99 What do you think of this?

monsieurtanuki:
Just implemented in `SqlTileWriter`:
* a `getDb` method to be used systematically in order to get the database (and to reconnect if necessary)
* a `refreshDb` method in order to explicitly ask for a database reconnection
* more subtile `catch`es of `SQLiteException` in order to automatically refresh the database when relevant

@spyhunter99 Does it cover all your requirements?

secaw:
@monsieurtanuki Without testing yet - looks like the perfect solution. What if we extend `SqlTileWriter`, will it also use the single instance?

monsieurtanuki:
@secaw I'm not sure it will work if you extend `SqlTileWriter`; perhaps by just overriding `getDb`. I'm not a _design pattern_ expert, but maybe using a _factory_ would help.
Do you currently extend `SqlTileWriter`? Because there are currently 10 places in the code where `new SqlTileWriter()` is used: how do you manage that?

secaw:
Ok, thanks - I just use an extended `SqlTileWriter` for custom cache scanning and cleaning. Had no conflicts so far, so probably doesn't matter anyways...

monsieurtanuki:
@secaw Thinking again about it that should be OK.
What is important is the singleton'ed database, not the `SqlTileWriter` instances - there are many, but with the same database. It should be the same for the classes that extend `SqlTileWriter`, as long as they use `SqlTileWriter.getDb`.

secaw:
@monsieurtanuki Thanks for clarification, that's what I reckoned too.

monsieurtanuki:
@InI4 @secaw During splash screen you can now run:
````java
new DuringSplashScreen().runDuringSplashScreen();
````

spyhunter99:
I'm OK with this

PR #1029, opened 2018-05-12
New projection systems: class TileSystem is now abstract and can be extended

monsieurtanuki:
New class:
* `TileSystemWebMercator`: the current implementation of a tile system in a Web Mercator projection

Impacted classes:
* `BookmarkSample`: now uses getters for min/max latitude/longitude
* `BoundBoxTest`: now uses getters for min/max latitude/longitude
* `BoundingBox`: now uses getters for min/max latitude/longitude
* `DomParserWms111`: now uses getters for min/max latitude/longitude
* `GeoPackageMapTileModuleProvider`: now uses getters for min/max latitude/longitude
* `GeoPointTest`: now uses getters for min/max latitude/longitude and other non static `TileSystem` methods
* `IconPlottingOverlay`: now uses getters for min/max latitude/longitude
* `LinearRing`: impacted the `getCenter` method
* `LinearRingTest`: now uses getters for min/max latitude/longitude
* `MapController`: now uses getters for min/max latitude/longitude
* `MapView`: created a `TileSystem` member with getter and setter, and a default `new TileSystemWebMercator()` value
* `MilStdPointPlottingOverlay`: now uses getters for min/max latitude/longitude
* `OpenStreetMapViewTest`: now uses getters for min/max latitude/longitude and other non static `TileSystem` methods
* `Overlay`: now uses getters for min/max latitude/longitude
* `PathProjection`: now uses non static `TileSystem` methods
* `Projection`: added a `TileSystem` in the constructor and as a member; now uses non static `TileSystem` methods
* `ProjectionTest`: impacted the new `TileSystem` parameter in the `Projection` constructor; now uses non static `TileSystem` methods
* `SampleAnimateTo`: now uses non static `TileSystem` methods
* `SampleAssetsOnlyRepetitionModes`: now uses getters for min/max latitude/longitude
* `SampleZoomToBounding`: now uses getters for min/max latitude/longitude and other non static `TileSystem` methods
* `microsoft.mappoint.TileSystem`: now uses non static `TileSystem` methods
* `org.osmdroid.util.TileSystem`: made it abstract, with 4 lat/lon to/from X01/Y01 methods and 4 min/max lat/lon getters; deprecated the min/max lat/lon members; removed the `static` keyword as much as possible in order to make the class extendable; renamed method `getX01FromLongitudeWithoutWrap` as `getX01FromLongitude` (same for `getY01FromLatitude`)
* `TileSystemTest`: now uses non static `TileSystem` methods of a `TileSystemWebMercator` instance

### DIFF
--- a/OpenStreetMapViewer/src/main/java/org/osmdroid/samplefragments/bookmarks/BookmarkSample.java
+++ b/OpenStreetMapViewer/src/main/java/org/osmdroid/samplefragments/bookmarks/BookmarkSample.java
@@ -89,9 +89,9 @@ public class BookmarkSample extends BaseSampleFragment {
                             valid = false;
                         }
 
-                        if (latD > mMapView.getTileSystem().getMaxLatitude() || latD < mMapView.getTileSystem().getMinLatitude())
+                        if (!mMapView.getTileSystem().isValidLatitude(latD))
                             valid = false;
-                        if (lonD > mMapView.getTileSystem().getMaxLongitude() || lonD < mMapView.getTileSystem().getMinLongitude())
+                        if (!mMapView.getTileSystem().isValidLongitude(lonD))
                             valid = false;
 
                         if (valid) {
--- a/OpenStreetMapViewer/src/main/java/org/osmdroid/samplefragments/bookmarks/BookmarkSample.java
+++ b/OpenStreetMapViewer/src/main/java/org/osmdroid/samplefragments/bookmarks/BookmarkSample.java
@@ -9,7 +9,6 @@ import org.osmdroid.R;
 import org.osmdroid.events.MapEventsReceiver;
 import org.osmdroid.samplefragments.BaseSampleFragment;
 import org.osmdroid.util.GeoPoint;
-import org.osmdroid.util.TileSystem;
 import org.osmdroid.views.overlay.MapEventsOverlay;
 import org.osmdroid.views.overlay.Marker;
 
@@ -90,9 +89,9 @@ public class BookmarkSample extends BaseSampleFragment {
                             valid = false;
                         }
 
-                        if (latD > TileSystem.MaxLatitude || latD < TileSystem.MinLatitude)
+                        if (latD > mMapView.getTileSystem().getMaxLatitude() || latD < mMapView.getTileSystem().getMinLatitude())
                             valid = false;
-                        if (lonD > TileSystem.MaxLongitude || lonD < TileSystem.MinLongitude)
+                        if (lonD > mMapView.getTileSystem().getMaxLongitude() || lonD < mMapView.getTileSystem().getMinLongitude())
                             valid = false;
 
                         if (valid) {
--- a/OpenStreetMapViewer/src/main/java/org/osmdroid/samplefragments/drawing/IconPlottingOverlay.java
+++ b/OpenStreetMapViewer/src/main/java/org/osmdroid/samplefragments/drawing/IconPlottingOverlay.java
@@ -5,7 +5,6 @@ import android.graphics.drawable.Drawable;
 import android.view.MotionEvent;
 
 import org.osmdroid.util.GeoPoint;
-import org.osmdroid.util.TileSystem;
 import org.osmdroid.views.MapView;
 import org.osmdroid.views.overlay.Marker;
 import org.osmdroid.views.overlay.Overlay;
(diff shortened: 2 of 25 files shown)


spyhunter99:
wow

monsieurtanuki:
@janekp Just PR'ed:
* on point 1, I fixed the "cleanLongitude on latitude" bug
* on point 2, I'm a bit confused with the tiles being fine and the geo points _not_ being fine at the same time. Would having something like a `getX01FromCoordinates(double latitude, double longitude)` help?
* on point 3, I think I fixed the bug by creating a default constructor for `BoundingBox` with no checks

It's about a map of Estonia, right? I ran the Tallinn marathon last year - and even finished ;)

monsieurtanuki:
@janekp On point 3
> N:1064500.0
Caused by: java.lang.IllegalArgumentException: north must be in [5993000.0,7017000.0]

Couldn't it be caused by a confusion (allegedly) on your side, because I can find 1064500 as the east limit in your configuration:
````java
MAAAMET = new Proj4TileSystem(
        "EPSG:3301",
...
        new RectF(40500, 5993000, ____1064500____, 7017000));
````

**Comment by janekp** [comment]
Found it.

```
Caused by: java.lang.IllegalArgumentException: north must be in [5993000.0,7017000.0]
        at org.osmdroid.util.BoundingBox.set(BoundingBox.java:68)
        at org.osmdroid.util.BoundingBox.<init>(BoundingBox.java:45)
        at org.osmdroid.views.overlay.Overlay.<init>(Overlay.java:56)
        at org.osmdroid.views.overlay.OverlayWithIW.<init>(OverlayWithIW.java:38)
        at org.osmdroid.views.overlay.Marker.<init>(Marker.java:103)
        at org.osmdroid.views.overlay.Marker.<init>(Marker.java:99)
```

**Comment by janekp** [comment]
```
private static final TileSystem tileSystem = new TileSystemWebMercator(); // used only for the default bounding box
	protected BoundingBox mBounds = new BoundingBox(tileSystem.getMaxLatitude(), tileSystem.getMaxLongitude(),tileSystem.getMinLatitude(),tileSystem.getMinLongitude());
```

janekp:
North was actually 85.05112877980659. Not 1064500.0. Wrong log. My bad.

janekp:
3. works, but has few issues with hardcoded BoundingBox / Overlay TileSystem constants - its possible to workaround these with Java Reflection APIs, but that seems hackish. 2. would indeed work if getX01 accept both latitude and longitude as inputs. That's how I did in the initial PoC.

There are strong arguments for keeping the tile system (web mercator) and coordinates (wgs84) separate. Many existing APIs depend on it. In my case non-wgs84 is ok and actually easier to manage, because coordinate conversions can be avoided.

monsieurtanuki:
@janekp I'm not an expert in GIS, therefore I cannot debate about Web Mercator tile system and WGS84 coordinates.

Regarding only the code, as I assume the coordinate conversions are bijective and can be performed in a very elegant and painless way (such as extending `GeoPoint`), I strongly prefer that the conversions are done on your side.
The idea is to preserve the osmdroid basics (Web Mercator / WGS84), and potentially provide users with simple steps on how to use a different tile/projection system.

An alternate solution would be to agree on some kind of `Projection` and `TileSystem` `interface`s, to build your own `Projection` and `TileSystem` classes, and to make it possible for a `MapView` to embed whatever `Projection` instance. Instead of bringing confusion to the current Web Mercator / WGS84 code by warping it.

janekp:
Re: conversions. These are done on application side at the moment and overall don't seem to be a bottleneck.
I think I can make it work even with the current PR ( Java Reflection for BoundingBox.tileSystem :) ).

Looking at alternative/longterm solution, 2 interfaces would indeed work (1 for tiles (x/y <-> lng/lat, web mercator by default) and 1 for coordinate system magic sauce (lng/lat, wgs84 by default)

monsieurtanuki:
For a short term solution, the only thing that bothers you now is `BoundingBox.tileSystem` being a `TileSystemWebMercator`, right?
What about a `static` setter for this member? To be called in `MapView.setTileSystem`?

janekp:
Yes. A static setter would be great for now. That would solve issues with BoundingBox/Overlay.

monsieurtanuki:
Just PR'ed: `static` `MapView.getTileSystem` is now the reference for all the classes. Value can be changed with `MapView.setTileSystem`. @janekp Tell me if it actually helps.

janekp:
It appears to work fine!

PR #1031, opened 2018-05-17
Removed the systematic and problematic "is SD card available?" check

monsieurtanuki:
If ever a map tile provider deserves to fail because it needs the SD card and the SD card is not available, it will fail anyway.
And if it doesn't need the SD card, it won't be prevented from working correctly anymore by this "is SD card available?" check.

Impacted classes:
* `MapTileFileArchiveProvider`: removed calls to `isSdCardAvailable`
* `MapTileFileStorageProviderBase`: removed static member `mSdCardAvailable` and methods `checkSdCard` and `isSdCardAvailable`
* `MapTileFilesystemProvider`: removed call to `isSdCardAvailable`
* `MapTileSqlCacheProvider`: removed call to `isSdCardAvailable`


spyhunter99:
looks good to me

Issue #690, opened 2017-08-28
Secondary tile overlay does not always display tiles


nijs9:
I'm using osmdoird 5.6.5 in my Android application. I'm using a custom baselayer, and I add a second layer above it in this way:

baselayer:

````
final MapTileProviderBasic tileProvider = new MapTileProviderBasic(getApplicationContext());
        final ITileSource tileSource = new XYTileSource(getResources().getString(R.string.rp_cachemap), 6, 17, 256, ".png", tileserver);
        tileProvider.setTileSource(tileSource);
        final TilesOverlay tilesOverlay = new TilesOverlay(tileProvider, this.getBaseContext());
        tilesOverlay.setLoadingBackgroundColor(Color.TRANSPARENT);
````

overlay:

````
final MapTileProviderBasic anotherTileProvider = new MapTileProviderBasic(getApplicationContext());
        final ITileSource tileSource2 = new XYTileSource("tile_overlay_3",6, 17, 256, ".png", new String[]{"https://geoserver2.nodemapp.com/geoserver/"})
        {
            @Override
            public String getTileURLString(MapTile aTile) {
                String hello= getBaseUrl() + "gwc/service/gmaps?layers=nodemapp-public-mobile-with-cache&zoom=" + aTile.getZoomLevel() + "&x=" + aTile.getX() + "&y=" + aTile.getY() + "&format=image/png&viewparams=kind:hike;groupid:1&version=1";
                System.out.println(hello);
                return hello;
            }
        };
        anotherTileProvider.setTileSource(tileSource2);
        final TilesOverlay secondTilesOverlay = new TilesOverlay(anotherTileProvider, this.getBaseContext());
        secondTilesOverlay.setLoadingBackgroundColor(Color.TRANSPARENT);
````

set tiles on map:

````
osmMap.setTileSource(tileSource);
osmMap.getOverlays().add(secondTilesOverlay);
osmMap.invalidate();
````

Eveything is working fine, but when scrolling the map, sometimes there are some tiles not showing properly in the second overlay. When scrolling a little bit further, they are loaded. Is this a known bug?

spyhunter99:
I have seen this issue with secondary tile overlays but I'm not entirely sure why. It's even present on the demo app. I think the sample has mapnik + some kind of road overlay (in blue). It's probably some kind of threading or concurrency issue. I'll make it as a bug and hopefully someone will look into further

nijs9:
[...]

PR #1037, opened 2018-05-19
bug/#690 - secondary tile overlay does not refresh when tiles download

monsieurtanuki:
I explained how to make things work in my comments of issue #690, and I fixed the demo.

Impacted class:
* `SampleWithTilesOverlayAndCustomTileSource`: added an explicit call to `setTileRequestCompleteHandler`


spyhunter99:
that's a brilliantly simple fix

monsieurtanuki:
The way osmdroid works is actually a bit counterintuitive:
* we don't display what is in the providers but what is in the memory cache (which is eventually the output of the providers)
* by default we don't refresh the display when we successfully download new tiles (which makes sense if we think about `CacheManager` download feature)

Issue #1012, opened 2018-04-24
Very rare occurences of NegativeArraySizeException in MapTileCache.clear


secaw:
reported in the Developer Console.
I have no idea how to reproduce this, but maybe we should just check against or catch it...

## Issue Type

[X] Bug

## Description and/or steps/code to reproduce the problem

```
java.lang.NegativeArraySizeException:  
  at org.osmdroid.util.MapTileList.<init> (SourceFile:25) 
  at org.osmdroid.tileprovider.MapTileCache.clear (SourceFile:149) 
  at org.osmdroid.tileprovider.MapTileProviderBase.clearTileCache (SourceFile:269) 
  at org.osmdroid.tileprovider.MapTileProviderBase.getMapTile (SourceFile:71)
                                                   detach
```

### If it's a bug, version(s) of android this affects:

All

### Version of osmdroid the issue relates to:

6.0.0

monsieurtanuki:
@secaw Very strange...
We're talking about
````java
final MapTileList list = new MapTileList(mCachedTiles.size());
````
that calls
````java
public MapTileList(final int initialCapacity) {
    mTileIndices = new long[initialCapacity];
}
````
How can we have a `HashMap` (`mCachedTiles`) with a negative `size()`? Maybe it's a concurrency side effect.

As you don't know how to reproduce this issue, the fix would be something like
````java
public MapTileList(final int initialCapacity) {
    mTileIndices = new long[Math.max(0, initialCapacity)];
}
````
with a bit of `synchronized (mCachedTiles)` in `MapTileCache.clear()`

secaw:
@monsieurtanuki Yes, it's weird. Might be related to Overlay also using `MapTileProviderBase`, but still... - shouldn't it have it's own `MapTileCache`? Solution sounds reasonable anyways.

spyhunter99:
Definitely a concurrency issue

monsieurtanuki:
Allegedly fixed in #1018 (could not reproduce the issue before).
@secaw As it looks like a very rare bug I presume you cannot actually test the fix. Therefore I close the issue; feel free to reopen it.

secaw:
@monsieurtanuki Great, thanks!

lfielke:
I've noticed instances of this crash even after the pull request merged in 5264f19ff2b5c36c0899c1d2c23f796b4785caba

[...]

PR #1041, opened 2018-05-24
bug/#1012 Add synchronization in MapTileCache.remove

lfielke:
Fix a race in `MapTileCache` (#1012).

This synchronizes a call to `mCachedTiles.remove`, the same way other calls are guarded.

As discussed on: https://github.com/osmdroid/osmdroid/issues/1012#issuecomment-391466140


lfielke:
I've added the `testConcurrency` test. You can easily test the problem by commenting out the synchronized block added in my first commit.

PR #1062, opened 2018-06-12
feature/#1044 : more flexibility for tile expiration

monsieurtanuki:
New method `MapTileDownloader.computeExpirationTime(Long pHttpExpiresTime)` can be overridden in order to set specific expiration times.

Impacted class:
* `MapTileDownloader`: added methods `private long computeExpirationTime(String pHttpExpiresHeader)` and overridable `protected long computeExpirationTime(Long pHttpExpiresTime)` in order to add flexibility; gently refactored


spyhunter99:
looks easily overridable. great. I'm afraid i jacked up your pr.

PR #1063, opened 2018-06-18
feature/#1004 - offline first

monsieurtanuki:
New class:
* `SampleLieFi`: new demo in "More Samples / Tilesources / Lie-fi - slow online source" - we emulate a slow online source in order to show the offline first behavior

Impacted classes:
* `MapTileDownloader`: created method `getLieFiLag` in order to emulate a lie-fi lag - of course it must ALWAYS be set to 0, except during lie-fi tests; used new method `getLieFiLag` accordingly in method `TileLoader.downloadTile`
* `MapTileProviderBasic`: moved the online provider at the end of the provider list
* `SampleFactory`: added new class `SampleLieFi`


monsieurtanuki:
@spyhunter99 I added a "lie-fi lag" feature in `MapTileDownloader`, for "offline first" testing purposes. Maybe it'd be better to create a dedicated `MapTileLieFiDownloader` private class instead, in order not to add confusion to `MapTileDownloader`. What do you think about it?

spyhunter99:
probably a better idea than adding an additional method. Does bumping the order to tile provides negative affect performance?

monsieurtanuki:
@spyhunter99 OK, I'll create a private class dedicated to unit test.
Regarding the performances, for the moment I'm still fine tuning the display - which is not 100% working. And then performances will matter.

PR #1081, opened 2018-07-05
bug/#1076 - rolling back to MapView in xml

monsieurtanuki:
Impacted file:
* `sample_cachemgr.xml`: rolled back to `MapView` instead of the recent `LinearLayout` change, as the corresponding code didn't take the same path and that caused an ANR (casting a `LinearLayout` as a `MapView`)


cbalster:
Hey, I just pushed PR #1084 which fixes this issue.

PR #1086, opened 2018-07-08
#1065 Fix handling of reusable bitmaps

### LINKED ISSUE #1065: Error decoding map tiles
## Issue Type

[x] Bug

## Description and/or steps/code to reproduce the problem

On API 16, 17 and 18 emulators I see an intermittent problem in `OpenStreetMapTileProviderDirectTest > test_getMapTile_found()`:
`org.osmdroid.views.util.OpenStreetMapTileProviderDirectTest > test_getMapTile_found[test-api18(AVD) - 4.3.1] [31mFAILED [0m
junit.framework.AssertionFailedError: Expect tile to be not null from path /data/data/org.osmdroid/files/osmdroid/OpenStreetMapTileProviderTest.png` 

Corresponding logcat is:
`D/skia    ( 2011): --- decoder->decode returned false`
`E/BitmapFactory( 2011): Unable to decode stream: java.lang.IllegalArgumentException: Problem decoding into existing bitmap`
`D/OsmDroid( 2011): /data/data/org.osmdroid/files/osmdroid/OpenStreetMapTileProviderTest.png is an invalid image file, deleting...`

Full logcat: https://api.travis-ci.org/v3/job/396125120/log.txt

Currently I assume this is a problem with the handling of reusable bitmaps on different API levels and the implementation of `org.osmdroid.tileprovider.BitmapPool` and related classes. I didn't investigate further though yet.

### If it's a bug, version(s) of android this affects:

Currently observed in 16, 17, 18

6.0.2

cbalster:
closes #1065 

So the problem was that on 10 < API < 19 the BitmapPool returned bitmaps for reuse without checking if size is identical, which is a hard requirement before API 19. This was unlikely to show up in production apps since most bitmaps are tiles which have identical sizes anyway.

With this change, we have to trust that the tilesize in the different `ITileSource`s is correct. Determining the size from the file or inputstream would add significant overhead.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -54,10 +54,10 @@ env:
     #- API=14 ANDROID_ABI=armeabi-v7a ANDROID_TAG=default         # emulator consistently times out
     #- API=15 ANDROID_ABI=armeabi-v7a ANDROID_TAG=default         # https://github.com/osmdroid/osmdroid/issues/1066
     #- API=15 ANDROID_ABI=armeabi-v7a ANDROID_TAG=google_apis     # currently not needed
-    #- API=16 ANDROID_ABI=armeabi-v7a ANDROID_TAG=default         # https://github.com/osmdroid/osmdroid/issues/1065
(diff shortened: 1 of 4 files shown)


spyhunter99:
awesome work

PR #1090, opened 2018-07-11
bug/#1040_2 - calls to listeners for ScrollEvents in MapView.setExpectedCenter

monsieurtanuki:
Impacted class: `MapView`

### DIFF
--- a/osmdroid-android/src/main/java/org/osmdroid/views/MapController.java
+++ b/osmdroid-android/src/main/java/org/osmdroid/views/MapController.java
@@ -203,10 +203,6 @@ public class MapController implements IMapController, OnFirstLayoutListener {
             return;
         }
         mMapView.setExpectedCenter(point);
-        ScrollEvent event = null;
-        for (MapListener mapListener: mMapView.mListners) {
-            mapListener.onScroll(event != null ? event : (event = new ScrollEvent(mMapView, 0, 0)));
-        }
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


monsieurtanuki:
@InI4 
> Looks reasonable. I will try to dig for an old branch of my app to test this later today.

Don't bother, if the test is more or less "I want to have several `ZoomEvent`s and `ScrollEvent`s during `animateTo`" I can do that.

> Its probably cheap and usually only one listener anyway, but I still feel like the <new ScrollEvent()> could be outside the loop?!

Yeah, I thought the same, and then imagined that you would create a `ScrollEvent` object for all the cases, even when you don't use at least one listener.
Now that you mention it again I found the solution:
````java
ZoomEvent event = null;
for (MapListener mapListener: mListners) {
	mapListener.onZoom(event != null ? event : (event = new ZoomEvent(this, newZoomLevel)));
}
````

monsieurtanuki:
There are some side-effects, because in `MapView.setZoomLevel` you call `MapView.setExpectedCenter`.
Therefore if you animate on both zooms and centers, you may end up calling twice as often `ScrollEvent`s (that happen during zoom change and center change) than `ZoomEvent`s (that happen only during zoom change).

InI4:
I can confirm your change would have saved the old version of my app which relied on a `(Delayed)MapListener` which probably also means it had no problems with doubled amounts of events. 

Doing no `ZoomEvents` in case of no listeners? This is memory optimal!

monsieurtanuki:
> Doing no ZoomEvents in case of no listeners? This is memory optimal!

It was already the case before, as we never entered the listener loop.

PR #1091, opened 2018-07-11
feature/#111 - getting rid of old classes

monsieurtanuki:
Deleted classes:
* `microsoft.mappoint.TileSystem`
* `org.osmdroid.TileSystemMathTest`


monsieurtanuki:
@spyhunter99 Shall I merge now, or should it be dealt with during the "official" release of 6.0.2? I mean, some apps may use that code, and it would be more acceptable with a release note.

PR #1105, opened 2018-07-21
feature/#1104 - more explicit Marker.setIcon methods and anchor setting fix

monsieurtanuki:
Impacted classes:
* `Marker`: deprecated static member `ENABLE_TEXT_LABELS_WHEN_NO_IMAGE`, member `mEnableTextLabelsWhenNoImage` and method `setIcon`; created methods `setTitleIcon`, `setTextIcon`, `setDrawableIcon` and `setDefaultIcon` that have more explicit names; fixed unexpected anchor setting; gently refactored
* `IconPlottingOverlay`, `IISTrackerBase`, `LatLonGridlineOverlay`, `LatLonGridlineOverlay2`, `LayerManager`, `MilStdMultipointOverlay`, `MilStdPointPlottingOverlay`, `OsmMapShapeConverter`, `Plotter`, `SampleCustomMyLocation`, `SampleMarker`, `SampleMilitaryIconsMarker`, `SampleOsmPath`: used `setTitleIcon` and `setDrawableIcon` instead of deprecated method `setIcon`; removed calls to `ENABLE_TEXT_LABELS_WHEN_NO_IMAGE` and `mEnableTextLabelsWhenNoImage`

### DIFF
--- a/OpenStreetMapViewer/src/main/java/org/osmdroid/samplefragments/data/IISTrackerBase.java
+++ b/OpenStreetMapViewer/src/main/java/org/osmdroid/samplefragments/data/IISTrackerBase.java
@@ -64,7 +64,7 @@ public abstract class IISTrackerBase extends SampleGridlines {
 
         marker = new Marker(mMapView);
         marker.setImage(image);
-        marker.setIcon(icon);
+        marker.setDrawableIcon(icon);
         marker.setTitle("International Space Station");
 
 
@@ -110,7 +110,7 @@ public abstract class IISTrackerBase extends SampleGridlines {
 
                                         marker = new Marker(mMapView);
                                         marker.setImage(image);
-                                        marker.setIcon(icon);
+                                        marker.setDrawableIcon(icon);
                                         marker.setTitle("International Space Station");
                                         marker.setPosition(location);
                                         mMapView.getController().setCenter(location);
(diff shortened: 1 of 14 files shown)


MKergall:
Oh my God! You removed setIcon! 
setIcon is "core" API of Marker, reproducing directly Google Maps API equivalent. 
We have to maintain a reasonable compatibility. 

You can provide additional methods for specific needs (for instance, a setIconAsLabel, or setTextIcon if you prefer), but you cannot remove this one.

spyhunter99:
i'll second @MKergall's remarks

monsieurtanuki:
@MKergall @spyhunter99 I did NOT remove `setIcon`: I added alternate explicit and non-ambiguous versions, and I deprecated the original `setIcon` method (_deprecated_, not removed).
I heard your sentimental and legacy arguments. If I remove the infamous deprecation tags, would that be OK for you guys?

monsieurtanuki:
Just removed the deprecation tags.

MKergall:
Yes :-)

Some remarks: 
1) A small bug: setTextIcon(final String pText) is not using pText as text, but marker.getTitle. 

2) There could be an extended setTextIcon variant, with parameters for most its hard coded settings (usual things should be simple, complex things should be possible). 

3) Small is beautiful: I think setTitleIcon and setDrawableIcon are useless. 

4) And as now people can use setTextIcon, the mess inside setIcon about enabling text labels should be removed.

monsieurtanuki:
@MKergall Just PR'ed
1. Fixed - `setTextIcon` now uses its parameter `pText`
1. I finally don't add another version of `setTextIcon` for several reasons:
	1. There are already available parameters like `mTextLabelBackgroundColor`, `mTextLabelFontSize` and `mTextLabelForegroundColor`, which gives room for customization
	1. Where would the customization end? `Paint` parameters? Border size parameters?
	1. The "complex" way is still possible, if the user builds its own icon
	1. Actually the cleanest way to achieve that might be to have a dedicated class that transforms texts into `Drawable`s and to replace `Marker.setTextIcon` with `Marker.setIcon(TransformFromTextIntoDrawable(String))`, but that's another story
1. Removed `setTitleIcon` and `setDrawableIcon`
1. Removed the lousy members

PR #1111, opened 2018-07-25
Some minor documentation fixes

oliverlockwood:
I noticed these when upgrading from `5.6.5` to `6.0.2` and making use of some new functionality.


oliverlockwood:
@monsieurtanuki fair point; comments addressed.

Issue #1098, opened 2018-07-16
Update to target api 26 to comply with playstore rules


spyhunter99:
Please note: issues related to or caused by osmbonuspack will be closed with a
suggestion to open an issue at https://github.com/MKergall/osmbonuspack

(feel free to delete whatever doesn't apply)

## Issue Type

[ ] Question
[ ] Bug
[ ] Improvement
[ x] Build system related
[ ] Performance
[ ] Documentation

cbalster:
Since this is becoming a problem for the demo app by the end of the month, I was wondering if there is any progress on this?
If I remember correctly the main problem here is gradlefury not working with gradle => 3.4 which prevents updating the support libs/build tools since they at some point require higher versions of gradle. Any chance of making gradlefury work with newer gradle versions?

spyhunter99:
Damn. I'm super busy at the moment. Maybe this weekend I can look at it.
The android gradle changes prevent pom generation for some reason. I can
try to manually generate it with strings, but that's no fun.

spyhunter99:
looking into it now. just did a test build. sonatype fails us on both javadocs and sources jar missing on the jre tile packager.

spyhunter99:
success, wasn't too bad actually

PR #1204, opened 2018-11-24
Bug/#1098 - Bump support lib to 26

cbalster:
The recent commits concerning #1098 only covered part of the problem: Having compileSDK on a different version than the support libs is guaranteed to cause issues.
This PR brings the support lib up to version to 26.x. Please note that I just made sure this builds an the tests run - there still might be some issues due to the rather large jump and the corresponding behavior changes in various components of the support lib. Ideally someone should go through https://developer.android.com/topic/libraries/support-library/revisions and check the notes for possible issues. 

Most importantly this PR also raises the minSDK for the demo app to 14, since the 26.0 support libs require it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,6 @@ android:
     #
     
     - android-26
-    - android-23
     #- addon-google_apis-google-23
     
     # Additional components - currently not required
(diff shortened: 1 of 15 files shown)


spyhunter99:
cool, thanks. building it now as a release build + sonatype push, then i'll drop the sonatype repo when done. it's really the only way to test this to make sure we can still release

spyhunter99:
this is good to go, the only issue i ran into was related to my gpg key, everything else checks out

PR #1208, opened 2018-11-27
BoundingBox: Rewrite increaseByScale()

schildbach:
* to take into account the dateline
* clip out-of-bounds latitude
* wrap-around out-of-bounds longitude


schildbach:
Thanks for your review! I didn't know about the assertEquals() variant :laughing:  Removed the test dep.

I'm also using a single TOLERANCE constant now; however I had to adjust it down from 1E10 to 1E5 because otherwise my tests would fail. I'm not sure if it's worth trying to increase precision.

About the overload, I can change it so that the old method stays in place if you prefer. In general I think the fact that the class didn't take into account the date line should be considered as a bug that should be fixed. And as such it doesn't require any of the new methods.

PR #1210, opened 2018-11-27
BoundingBox: Fix wrapping of longitude.

### LINKED ISSUE #1207: BoundingBox: 360 degrees vs 180 degrees?
https://github.com/osmdroid/osmdroid/blob/e1ab630826f7b360d39a81ab96f57328f5e6bb51/osmdroid-android/src/main/java/org/osmdroid/util/BoundingBox.java#L223

Maybe I'm wrong, but a couple of times in this class only 180° is used as a "wrap around delta" for longitude, rather than 360°. Can this be right?

schildbach:
Fixes #1207.


monsieurtanuki:
@schildbach What about `increaseByScale`?

PR #1212, opened 2018-11-28
bug/#1209 - using a resource for direction icon on each dpi folder

monsieurtanuki:
Removed resource:
* `drawable.direction_arrow.png` (only 1 version)

New resource:
* `drawable.round_navigation_white_48.png` (5 versions from `mdpi` to `xxxhdpi`)

Impacted classes:
* `CustomPaintingSurface`, `DirectedLocationOverlay`, `MyLocationNewOverlay`: used drawable `round_navigation_white_48` instead of `direction_arrow`


spyhunter99:
looks ok to me

PR #1227, opened 2018-12-10
bug/#994_2 - fix of irrelevant drawing of focused base for ItemizedOverlayWithFocus

monsieurtanuki:
Impacted class:
* `ItemizedOverlayWithFocus`: fixed in method `draw` the drawing of the marker focused base


monsieurtanuki:
@spyhunter99 Can't run integration checks. It looks like there a gradle config issue. Is that a "don't worry, just try later" case or is it more serious, I don't know.

monsieurtanuki:
What if I merge this PR: is that going to fail for the same reason, or is there a possibility that it works as we are talking about different check/integration processes?

cbalster:
As it is right now it will fail. The ci tests running on PRs and master are
the same. I will try to find a solution the next days.

spyhunter99:
https://www.reddit.com/r/androiddev/comments/a4wpuk/for_anyone_wondering_about_todays_failed_builds/

monsieurtanuki:
Now it's OK.

PR #1253, opened 2019-01-11
feature/#990_2 - introducing the concept of MapSnapshot

monsieurtanuki:
A `MapSnapshot` makes it easy to
* create a bitmap from map parameters
* in the background
* potentially waiting for the latest versions of the tiles to be downloaded

There are also 2 new items in the `TilesOverlay` menu (cf. "More Samples" in the demo app):
* "States", which `Toast`s the current state of the display tiles
* "Snapshot", which creates a bitmap out of the current `MapView`

New classes:
* `MapSnapshot`: creates a bitmap in the background from `MapView`-like data but without a `MapView`
* `MapSnapshotHandler`: custom-made `Handler` for `MapSnapshot`

Related changes:
* `InfoWindow`: handled the case when the `InfoWindow` is to be displayed but not on a `MapView` (solution: we don't draw but we don't crash)
* `MapTileProviderBase`: replaced the unique `Handler` by a `Collection<>`, deprecated method `setTileRequestCompleteHandler` in favour of `getTileRequestCompleteHandlers`
* `MapView`: now using `MapTileProviderBase.getTileRequestCompleteHandlers`
* `MinimapOverlay`: now using `MapTileProviderBase.getTileRequestCompleteHandlers`
* `TileStates`: replaced the unique `Runnable` by a `Collection<>`, deprecated method `setRunAfter` in favour of `getRunAfters`
* `TilesOverlay`: enriched the menu with 2 options - "States" that displays the current states of tiles as a `Toast`, and "Snapshot" that export a snapshot of the current map into a file ("snapshot.png" in the osmdroid directory); handled the "snapshot preparation" case when we want to loop through the tiles but with a null `Canvas`
* `strings.xml`: added the 2 new `TilesOverlay` menu options
* `SampleTileStates`: now using `TileStates.getRunAfters`
* `SampleWithTilesOverlayAndCustomTileSource`: now using `MapTileProviderBase.getTileRequestCompleteHandlers`; gently refactored

Unrelated refactoring:
* `IConfigurationProvider`: fixed typo; removed reference to now deprecated class `OsmPath`
* `MapController`: now using the remaining `MyMath` class
* `Mercator`: removed the class, as it was already deprecated 5 years ago...
* `org.osmdroid.util.MyMath`: moved code of method `getNextSquareNumberAbove` from the other `MyMath` class
* `org.osmdroid.views.util.MyMath`: deprecated the whole class and moved its only method into the other `MyMath` class
* `PathProjection`: deprecated the whole class in favour of `Polygon` or `Polyline`
* `OsmBitmapShader`: deprecated the whole class
* `OsmPath`: deprecated the whole class in favour of `Polygon` or `Polyline`


monsieurtanuki:
Please feel free to test the new `MapSnapshot` feature, to try to find bugs or possible improvements, and to tell me what you think.

pasniak:
How do I drive this through Samples's menu?  There are 2 entries:
image

pasniak:
The duplication in menu is maybe some bug?  Anyway I found the sample:
```
else if (menuId == MENU_SNAPSHOT) {
			final MapSnapshot mapSnapshot = new MapSnapshot(new MapSnapshot.MapSnapshotable() {
				@Override
				public void callback(final MapSnapshot pMapSketch) {
					if (pMapSketch.getStatus() != MapSnapshot.Status.CANVAS_OK) {
						return;
					}
					final File file = new File(Configuration.getInstance().getOsmdroidBasePath(), "snapshot.png");
					pMapSketch.save(file);
					pMapSketch.onDetach();
				}
			}, MapSnapshot.INCLUDE_FLAG_UPTODATE, pMapView);
			new Thread(mapSnapshot).start();
			return true;
		}
```

My comment : maybe a constructor with 11 args is an overkill? No way to have a  builder pattern?  I am referring to:

```
    public MapSnapshot(final MapSnapshotable pMapSnapshotable,
                       final int pIncludeFlags,
                       final MapTileProviderBase pTileProvider,
                       final List<Overlay> pOverlays,
                       final int pWidth, final int pHeight,
                       final double pZoomLevel, final IGeoPoint pCenter,
                       final float pOrientation,
                       final boolean pIsHorizontalWrapEnabled, final boolean pIsVerticalWrapEnabled) 
```
Something that would allow simple usage:
MapshotBuilder(pMapSnapshotable).withFlags(f).withOverlay(o).withSize(w,h).atZoom(l).atCenter(c)

What are the flags for?
```
    public static final int INCLUDE_FLAG_UPTODATE   = 1;
    public static final int INCLUDE_FLAG_EXPIRED    = 2;
    public static final int INCLUDE_FLAG_SCALED     = 4;
    public static final int INCLUDE_FLAG_NOTFOUND   = 8;
    public static final int INCLUDE_FLAGS_ALL =
            INCLUDE_FLAG_UPTODATE + INCLUDE_FLAG_EXPIRED + INCLUDE_FLAG_SCALED + INCLUDE_FLAG_NOTFOUND;
```

monsieurtanuki:
> 
> 
> How do I drive this through Samples's menu? There are 2 entries:
> image

I don't think this duplicate is specific to the new feature: in your screenshot "Map Source" and "Set to offline mode" is duplicated too.
And I cannot see that duplicate on my smartphone.
It looks like in `DefaultOverlayManager.onCreateOptionsMenu`, `TilesOverlay.onCreateOptionsMenu` may be called twice, which would explain the duplicates.
Could be specific to a specific `Overlay` combination. @pasniak In which demo did you find the duplicates?

IMHO the menu code shouldn't be embedded into class `TilesOverlay`. `TilesOverlay` is something useful to every osmdroid user, and the menu is useful only for the demo app.

monsieurtanuki:
I'm not sure if the builder pattern is possible as we definitely need all parameters, and preferably as `final`.

There are parameters with potential default values:
* `int pIncludeFlags` would default to `INCLUDE_FLAGS_ALL`
* `List<Overlay> pOverlays` would default to `null`
* `float pOrientation` would default to `0f`
* `boolean pIsHorizontalWrapEnabled` and `boolean pIsVerticalWrapEnabled` would default to `true`

What can we do with those defaultable parameters? Remove them from the constructor, and implement a sort of "set only once, and before calling `run`"?

I like the idea of a constructor with all the parameters, because you explicitly see what you need and what you can do.

Another solution would be a new constructor with a `Projection pProjection` parameter replacing the 7 parameters `int pWidth, int pHeight, double pZoomLevel, IGeoPoint pCenter, float pOrientation, boolean pIsHorizontalWrapEnabled, boolean pIsVerticalWrapEnabled`.
But that would mean that in order to create a snapshot you need to understand how to "construct" a `Projection`: what to put in `Rect pScreenRect, pScrollX, long pScrollY, TileSystem pTileSystem`. And that's not the easiest part.

Anyway I assume most users will use the snapshot using the `MapView` constructor.

I can add some comments in the code. For instance, explaining the "include flags": which tile states I'm accepting in my snapshot.

What do you think folks?

monsieurtanuki:
@pasniak I cannot even put aside the 5 parameters with default values as 4 of them are deduced from the `MapView`. That would cause confusion, something like "you can call `MapSnapshot.setOrientation` only if you used the constructor without the `MapView` parameter".

Perhaps, a solution could be to create a simpler version of the `Projection` constructor, and use it in the `MapSnapshot` constructor instead of the big list of parameters?
````java
public Projection(
		double pZoomLevel, int pWidth, int pHeight,
		GeoPoint pCenter,
		float pOrientation,
		boolean pHorizontalWrapEnabled, boolean pVerticalWrapEnabled)
````

````java
public MapSnapshot(MapSnapshotable pMapSnapshotable,
				   int pIncludeFlags,
				   MapTileProviderBase pTileProvider,
				   List<Overlay> pOverlays,
				   Projection pProjection)
````

monsieurtanuki:
@pasniak Happier with the new `MapSnapshot` constructor?

PR #1274, opened 2019-02-19
bug/#1247 - new `MapView.setDestroyMode` feature

monsieurtanuki:
So far `MapView.onDetach` was systematically called in `MapView.onDetachedFromWindow`, which could cause problems when the view was detached from the window but not to be destroyed (in rare cases).
Now we have a `boolean` that says if we should call it. The default value is `true` for code legacy reasons.

The correct way is now something like:
````java
@Override
public View onCreateView(LayoutInflater inflater, ViewGroup container, Bundle savedInstanceState) {

	// ...
	mMapView.setDestroyMode(false);
}

@Override
public void onDestroyView() {
	super.onDestroyView();
	mMapView.onDetach();
}
````

Impacted classes:
* `MapView`: new `boolean` member `mDestroyModeOnDetach` with a setter; impacted method `onDetachedFromWindow` accordingly; minor refactoring
* `StarterMapFragment`: added an explicit `mMapView.setDestroyMode(false)` on the `onCreateView` method; minor refactoring


pasniak:
Great! I tried your branch and I get no more exceptions.

PR #1275, opened 2019-02-20
bug/#1265 - switching to mapsforge 0.10.0 (from 0.8.0)

monsieurtanuki:
Impacted files:
* `MapsForgeTileSource.java`: removed the call to `ReadBuffer.setMaximumBufferSize` in `createInstance`
* `osmdroid-mapsforge/build.gradle`: now using version `0.10.0` of mapsforge

### DIFF
--- a/osmdroid-mapsforge/build.gradle
+++ b/osmdroid-mapsforge/build.gradle
@@ -37,9 +37,9 @@ dependencies {
     testCompile "junit:junit:${project.property('junit.version')}"
 
     //Mapsforge rendering and database support, which is LGPL
-    compile 'org.mapsforge:mapsforge-map-android:0.8.0'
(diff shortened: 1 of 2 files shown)


TomasJohansson:
If you look at how mapsforge changed their code when they removed the method,
they instead introduced a new class "Parameters".
https://github.com/mapsforge/mapsforge/commit/c755db6f8f316260a52cd4815c1b677845282349#diff-77b9f5256e13803ec9ab28654633fd5a
Instead of using ReadBuffer.MAXIMUM_BUFFER_SIZE they use Parameters.MAXIMUM_BUFFER_SIZE.
So, maybe that Parameters.MAXIMUM_BUFFER_SIZE should now be set instead of just removing the setting of ReadBuffer.MAXIMUM_BUFFER_SIZE (through the now removed method setMaximumBufferSize) ?

PR #1285, opened 2019-03-02
bug/#1280 - possibility to draw Polylines using Path

monsieurtanuki:
New classes:
* `PolyOverlayWithIW`: extending `OverlayWithIW`, extended by both `Polyline` and `Polygon` given the number of common methods
* `SampleDrawPolylineAsPath`: sample similar to `SampleDrawPolyline`, but where `Polyline`s are drawing in the "path" mode (and not in the "lines" mode)

Impacted classes:
* `CustomPaintingSurface`: created mode `PolylineAsPath`; put alpha in the color in order to see the vertices for `Polyline`s
* `LinearRing`: added the `boolean` member `mClosed` in order to make both `Polyline` and `Polygon` draw using a `Path`; new constructor for `Path` with a `boolean pClosed` parameter; taking into account the `mClosed` member when building the `Path`
* `SampleFactory`: added `SampleDrawPolylineAsPath` just after `SampleDrawPolyline`
* `Polygon`: now extends new class `PolyOverlayWithIW`
* `Polyline`: now extends new class `PolyOverlayWithIW`

Impacted classes - slightly unrelated refactoring regarding `Paint`s:
* `Bug382Crash`
* `HeatMap`
* `LatLonGridlineOverlay`
* `LatLonGridlineOverlay2`
* `MarkerDrag`
* `MilStdMultipointOverlay`
* `OsmMapShapeConverter`
* `SampleMilestonesNonRepetitive`
* `SampleOsmPath`
* `SampleRace`
* `SampleZoomToBounding`

### DIFF
--- a/osmdroid-android/src/main/java/org/osmdroid/views/overlay/LinearRing.java
+++ b/osmdroid-android/src/main/java/org/osmdroid/views/overlay/LinearRing.java
@@ -60,11 +60,15 @@ class LinearRing{
 	private boolean mGeodesic = false;
 
 	/**
+	 * @since 6.1.0
+	 */
+	private final boolean mClosed;
+
+	/**
 	 * Dedicated to `Path`
 	 */
 	public LinearRing(final Path pPath) {
-		mPath = pPath;
-		mPointAccepter = new PathBuilder(pPath);
+		this(pPath, true);
 	}
 
 	/**
@@ -74,6 +78,16 @@ class LinearRing{
 	public LinearRing(final LineBuilder pLineBuilder) {
 		mPath = null;
 		mPointAccepter = pLineBuilder;
+		mClosed = false;
+	}
+
+	/**
+	 * @since 6.1.0
+	 */
+	public LinearRing(final Path pPath, final boolean pClosed) {
+		mPath = pPath;
+		mPointAccepter = new PathBuilder(pPath);
+		mClosed = pClosed;
 	}
 
 	void clearPath() {
@@ -194,9 +208,11 @@ class LinearRing{
 			getBestOffset(pProjection, offset);
 		}
 		mSegmentClipper.init();
-		clipAndStore(pProjection, offset, true, pStorePoints, mSegmentClipper);
+		clipAndStore(pProjection, offset, mClosed, pStorePoints, mSegmentClipper);
 		mSegmentClipper.end();
-		mPath.close();
+		if (mClosed) {
+		    mPath.close();
+        }
 		return offset;
 	}
 
--- a/osmdroid-android/src/main/java/org/osmdroid/views/overlay/PolyOverlayWithIW.java
+++ b/osmdroid-android/src/main/java/org/osmdroid/views/overlay/PolyOverlayWithIW.java
@@ -1,0 +1,233 @@
+package org.osmdroid.views.overlay;
+
+import android.graphics.Canvas;
+import android.graphics.Paint;
+import android.graphics.Path;
+
+import org.osmdroid.util.GeoPoint;
+import org.osmdroid.util.PointL;
+import org.osmdroid.views.MapView;
+import org.osmdroid.views.Projection;
+import org.osmdroid.views.overlay.infowindow.InfoWindow;
+import org.osmdroid.views.overlay.milestones.MilestoneManager;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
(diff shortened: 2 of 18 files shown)


conej730:
Thanks @monsieurtanuki this is great.

osmdroid-android/src/main/java/org/osmdroid/views/overlay/LinearRing.java
```diff
@@ -194,9 +208,11 @@ PointL buildPathPortion(final Projection pProjection,
 			getBestOffset(pProjection, offset);
 		}
 		mSegmentClipper.init();
-		clipAndStore(pProjection, offset, true, pStorePoints, mSegmentClipper);
+		clipAndStore(pProjection, offset, mClosed, pStorePoints, mSegmentClipper);
 		mSegmentClipper.end();
-		mPath.close();
+		if (mClosed) {
+		    mPath.close();
+        }
```
conej730: tab issue

osmdroid-android/src/main/java/org/osmdroid/views/overlay/PolyOverlayWithIW.java
```diff
@@ -0,0 +1,233 @@
...
+		}
+		if (pUsePath) {
+			mPath = new Path();
+			mLineDrawer = null;
+			mOutline = new LinearRing(mPath, pClosePath);
+		} else {
+			mPath = null;
+			mLineDrawer = new LineDrawer(256);
+			mOutline = new LinearRing(mLineDrawer);
+			////mOutline.clearPath();
```
conej730: I think this comment could be removed?

osmdroid-android/src/main/java/org/osmdroid/views/overlay/PolyOverlayWithIW.java
```diff
@@ -0,0 +1,233 @@
...
+			milestoneManager.init();
+			milestoneManager.setDistances(mOutline.getDistances());
+			for (final PointL point : mOutline.getPointsForMilestones()) {
+				milestoneManager.add(point.x, point.y);
+			}
+			milestoneManager.end();
+		}
+
+		for (final MilestoneManager milestoneManager : mMilestoneManagers) {
+			milestoneManager.draw(pCanvas);
```
conej730: Can the draw be done in the previous loop after the milestone manager is populated, or does it cause delayed draw behavior that way?

spyhunter99:
let's hold off on merging this until the rest of the bugs are ironed out.

osmdroid-android/src/main/java/org/osmdroid/views/overlay/PolyOverlayWithIW.java
monsieurtanuki: We need these two steps.
The first step computes the actual "milestones", which can be something like a milestone every km, or 5 equidistant milestones for the whole length of the polyline. Therefore we need to parse everything first. Then draw, in the second step.

osmdroid-android/src/main/java/org/osmdroid/views/overlay/PolyOverlayWithIW.java
monsieurtanuki: Fair enough, `mOutline.clearPath();` should be executed and not commented.

osmdroid-android/src/main/java/org/osmdroid/views/overlay/LinearRing.java
monsieurtanuki: OK

conej730:
We've been testing with this branch for a bit now.  So far no issues have arisen.

Ruediger34:
I am totally fine with this implementation. 
Tested with no performance related impacts or other issues in my use case. 

Would be nice to this in the next release. 

Thank you @monsieurtanuki

spyhunter99:
roger that. i want to cut another version before merging this, but it will be first on the list

monsieurtanuki:
@spyhunter99 Any progress on a next version?

spyhunter99:
i can cut it today but we still have that one bug with the stack overflow issue

monsieurtanuki:
> i can cut it today but we still have that one bug with the stack overflow issue

Do you mean only #1299?

spyhunter99:
yup

spyhunter99:
merge conflicts

monsieurtanuki:
solved conflicts

PR #1291, opened 2019-03-06
bug/#1287 - Android 9 - handling "http only" (not "httpS") tile providers

monsieurtanuki:
Impacted files:
* `OpenStreetMapViewer:AndroidManifest.xml`: added a reference to new config file `res/xml/network_security_config.xml`
* `OpenStreetMapViewer:res/xml/network_security_config.xml`: whitelisting of "http only" (and not "httpS") tile providers
* `osmdroid:TileSourceFactory`: whenever possible, switched the "http" tile providers into "https"


spyhunter99:
looks good to me

InI4:
I did not check all map sources, but is good here.

PR #1293, opened 2019-03-08
feature/#1289_2 - normalizing user agent and disabling preventive downloads

monsieurtanuki:
… for MAPNIK

Around `TileSourcePolicy`:
* "normalized" user agent (package name + "/" + version)
* disabling preventive download

Impacted classes:
* `DefaultConfigurationProvider`: new member `String mNormalizedUserAgent` and its getter; new method `computeNormalizedUserAgent(Context)` to compute its value
* `IConfigurationProvider`: added method `String getNormalizedUserAgent()`
* `MapTilePreCache`: added a check in method `search` - if the tile source policy says "we don't try to preventively download tiles", there won't be any download attempt
* `MapTileProviderBasic`: added the `MapTileDownloader` in the pre-cache tile providers
* `TileDownloader`: in `downloadTile`, uses the normalized user agent if the tile source policy says so
* `TileSourceFactory`: edited the `TileSourcePolicy` related to MAPNIK (new parameters)
* `TileSourcePolicy`: now using `int` flags instead of `boolean`s (therefore new constructor); new methods `acceptsPreventive` and `normalizesUserAgent`

### DIFF
--- a/osmdroid-android/src/main/java/org/osmdroid/config/DefaultConfigurationProvider.java
+++ b/osmdroid-android/src/main/java/org/osmdroid/config/DefaultConfigurationProvider.java
@@ -2,6 +2,8 @@ package org.osmdroid.config;
 
 import android.content.Context;
 import android.content.SharedPreferences;
+import android.content.pm.PackageInfo;
+import android.content.pm.PackageManager;
 import android.os.Build;
 import android.util.Log;
 
@@ -61,6 +63,11 @@ public class DefaultConfigurationProvider implements IConfigurationProvider {
     protected int mTileGCBulkSize = 20;
     protected long mTileGCBulkPauseInMillis = 500;
     protected boolean mTileDownloaderFollowRedirects = true;
+
+    /**
+     * @since 6.1.0
+     */
+    private String mNormalizedUserAgent;
 
     public DefaultConfigurationProvider(){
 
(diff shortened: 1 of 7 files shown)


akoua:
Please @monsieurtanuki how can I modify these classes in my project ? (I ask this question because I know that it's impossible to modify a class of OSMdroid in an android project)

monsieurtanuki:
@akoua
* if you have a problem with the maps not being displayed, the fix is probably just to set a relevant user agent, using something like this: `Configuration.getInstance().setUserAgentValue(context.getPackageName());`
* if you really want the latest version of those classes, you can get the full sources and include them in your code (which makes obvious the fact that I'm not a `git` expert)
* if you want to modify those classes in order to contribute to osmdroid, welcome!

I hope this helps.

ambrose-at-cordic:
I have an app which was being blocked as it was using a unique string rather than the package name for the user-agent value, 

Once I used: OpenStreetMapTileProviderConstants.setUserAgentValue(getApplicationContext().getPackageName()): instead, that resolved the problem. The problem is, when our tested tried reporduce the problem on the same device with version of the app without the aforemenioned change, it wasn't blocked. Why would this happen? shouldn't it have blocked it once it had seen the invalid user-agent value?

PR #1306, opened 2019-03-30
feature/#1294 - additional offset parameters for IMapController.animateTo

monsieurtanuki:
New class:
* `SampleAnimateToOffset`: demo about the new offset parameters of `IMapController.animateTo` - demo that can be reached at "More Samples / Events / Animate To with Offsets"

Impacted classes:
* `IMapController`: additional version of method `animateTo`, with the extra int offset parameters
* `MapController`: additional version of method `animateTo`, with the extra int offset parameters, with its impacts on inner class `MapAnimatorListener`, `ReplayController` and `ReplayClass`; gentle refactoring
* `SampleFactory`: added the reference to the new `SampleAnimateToOffset` demo


pasniak:
It would be better to have more points to get the idea across.  I added below code but the effect is not what I want.

```
        final GeoPoint geoPoint = mList.get(pIndex % mList.size());
        final Double bearing = mList.get(prevIndex % mList.size()).bearingTo(geoPoint);
        final Float orientation = bearing.floatValue();
```

See https://youtu.be/5TyETWpzJPw?t=696 for the effect.

monsieurtanuki:
Closing this PR as the solution does not seem to be related to `animateTo`, but intrinsicly to the `MapView` instead: "I want that each time I use `MapView.setExpectedCenter`, I go to the same screen pixel _and_ it can be something different from the screen center (which is the default behavior)".

Issue #1299, opened 2019-03-25
Zoom Button Crash in 6.1.0


InI4:
## Issue Type

[X] Bug

## Description and/or steps/code to reproduce the problem

Plugging 6.1.0 into gradle to replace 6.0.3 in my App, all things so far seem fine. 
All but the (standard) zoom buttons: These bottons relyably crash 100% of the time after app start. With plenty of stack trace which all seem repeations of this (one click creates hundreds and so far I only checked manually):

```
2019-03-25 09:26:07.704 ? E/AndroidRuntime:     at org.osmdroid.views.MapController$MapAnimatorListener.onAnimationUpdate(MapController.java:520)
        at android.animation.ValueAnimator.animateValue(ValueAnimator.java:1547)
        at android.animation.ValueAnimator.end(ValueAnimator.java:1133)
        at org.osmdroid.views.MapController.animateTo(MapController.java:167)
        at org.osmdroid.views.MapController.animateTo(MapController.java:137)
        at org.osmdroid.views.MapController.animateTo(MapController.java:183)
        at org.osmdroid.views.MapController.animateTo(MapController.java:129)
        at org.osmdroid.views.MapView.setZoomLevel(MapView.java:492)
        at org.osmdroid.views.MapController$MapAnimatorListener.onAnimationUpdate(MapController.java:520)
        at android.animation.ValueAnimator.animateValue(ValueAnimator.java:1547)
        at android.animation.ValueAnimator.end(ValueAnimator.java:1133)
        at org.osmdroid.views.MapController.animateTo(MapController.java:167)
        at org.osmdroid.views.MapController.animateTo(MapController.java:137)
        at org.osmdroid.views.MapController.animateTo(MapController.java:183)
        at org.osmdroid.views.MapController.animateTo(MapController.java:129)
        at org.osmdroid.views.MapView.setZoomLevel(MapView.java:492)
        at org.osmdroid.views.MapController$MapAnimatorListener.onAnimationUpdate(MapController.java:520)
        at android.animation.ValueAnimator.animateValue(ValueAnimator.java:1547)
        at android.animation.ValueAnimator.end(ValueAnimator.java:1133)
        at org.osmdroid.views.MapController.animateTo(MapController.java:167)
        at org.osmdroid.views.MapController.animateTo(MapController.java:137)
        at org.osmdroid.views.MapController.animateTo(MapController.java:183)
        at org.osmdroid.views.MapController.animateTo(MapController.java:129)
[...]

PR #1307, opened 2019-03-30
bug #1299: Zoom Button Crash in 6.1.0

shorti1996:
Fix for https://github.com/osmdroid/osmdroid/issues/1299

fix StackOverflow, because when zooming in with zoom buttons,
osm tries to keep the map centered to the current user location.
See: org.osmdroid.views.overlay.mylocation.MyLocationNewOverlay.onSnapToItem

### DIFF
--- a/osmdroid-android/src/main/java/org/osmdroid/views/MapController.java
+++ b/osmdroid-android/src/main/java/org/osmdroid/views/MapController.java
@@ -164,7 +164,7 @@ public class MapController implements IMapController, OnFirstLayoutListener {
             }
 
             if (mCurrentAnimator != null) {
-                mCurrentAnimator.end();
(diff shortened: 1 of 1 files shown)


JiKanghui:
It's finally settled. Thank you.

InI4:
Just wanted to reiterate, that my app is now running 3 months on the corresponding snapshot and I have no issues at this end!

monsieurtanuki:
Tested with the new "Snappable" dedicated demo (#1371): no more stack overflow, and a behavior that looks reasonable.

@spyhunter99 Adding the correct demo behavior to @InI4's very positive feedback, I see no reason to delay the merging of this PR as it fixes #1299. And it looks like it could be time for another version of osmdroid.

umartinez22:
Thanks for fixing this problem.

PR #1353, opened 2019-06-05
feature/#1294_2 - new MapView.setMapCenterOffset method

monsieurtanuki:
The map center used to be projected into the screen center.
Now we have a possible offset from the screen center; default offset is [0, 0].

New class:
* `SampleMapCenterOffset`: demo about the new `MapView.setMapCenterOffset` method - demo that can be reached at "More Samples / Events / Animate To with Map Center Offset"

Impacted classes:
* `MapView`: created `int` members `mMapCenterOffsetX` and `mMapCenterOffsetY` with getters and a common setter; used it when constructing a new related `Projection`
* `Projection`: created `final int` members `mMapCenterOffsetX` and `mMapCenterOffsetY`; added these parameters to constructors; impacted accordingly methods `getScreenCenterX` and `getScreenCenterY`; gently refactored
* `SampleFactory`: added the new `SampleMapCenterOffset` demo
* `SampleMapSnapshot`: impacted the use of a `Projection` constructor

### DIFF
--- a/OpenStreetMapViewer/src/main/java/org/osmdroid/samplefragments/SampleFactory.java
+++ b/OpenStreetMapViewer/src/main/java/org/osmdroid/samplefragments/SampleFactory.java
@@ -37,6 +37,7 @@ import org.osmdroid.samplefragments.data.SampleRace;
 import org.osmdroid.samplefragments.data.SampleSimpleFastPointOverlay;
 import org.osmdroid.samplefragments.data.SampleSimpleLocation;
 import org.osmdroid.samplefragments.events.SampleAnimateToWithOrientation;
+import org.osmdroid.samplefragments.events.SampleMapCenterOffset;
 import org.osmdroid.samplefragments.tileproviders.SampleTileStates;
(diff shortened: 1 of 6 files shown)


PaoloDozio:
When do you think this added feature will be available in the baseline?

pasniak:
ProjectionTest.java is missing offsets 0, 0

pasniak:
After I fixed the build, the change works great, I tested on emu and tablet with a compass-driven-orientation.  Haven't implemented yet switching between offset and no offset.

monsieurtanuki:
Oh, ProjectionTest, of course.

You didn't mention it before: what do you expect as a transition between different offsets? I guess something like should work:
````java
// from (offsetX, offsetY) to (0, 0)
GeoPoint g = mapview.getCenter;
mapview.setMapCenterOffset(0, 0); // no invalidate
mapview.setExpectedCenter(g, -offsetX, -offsetY); // supposed a transparent transition
mapview.animateTo(g); // smooth animation
````

spyhunter99:
this looks good to me, just fix the merge conflict.

PR #1358, opened 2019-06-13
feature/#1347 - SpeechBalloonOverlay

monsieurtanuki:
New classes:
* `SampleSpeechBalloon`: Demo around the new `SpeechBalloonOverlay` feature in "More Samples / Data / Speech Balloon"
* `SpeechBalloonHelper`: Helper class for `SpeechBalloonOverlay` that includes method `compute` which computes the intersection between a rectangle and the triangle that starts from a Point and goes to a circle centered on the rectangle's center
* `SpeechBalloonHelperTest`: Unit test class for `SpeechBalloonHelper`
* `SpeechBalloonOverlay`: Overlay that displays a title in a "speech balloon"

Impacted classes:
* `MyMath`: added methods `computeAngle` and `computeCirclePoint`
* `MyMathTest`: added methods `testComputeAngle` and `testComputeCirclePoint`
* `RectL`: added methods `centerX` and `centerY`
* `SampleFactory`: added a reference to new demo `SampleSpeechBalloon`

### DIFF
--- a/osmdroid-android/src/main/java/org/osmdroid/util/SpeechBalloonHelper.java
+++ b/osmdroid-android/src/main/java/org/osmdroid/util/SpeechBalloonHelper.java
@@ -1,0 +1,87 @@
+package org.osmdroid.util;
+
+import org.osmdroid.views.overlay.SpeechBalloonOverlay;
+
+/**
+ * Helper class for {@link SpeechBalloonOverlay}
+ * whose main interest is method {@link #compute(RectL, PointL, double, PointL, PointL)}
+ * @since 6.1.1
+ * @author Fabrice Fontaine
+ */
+public class SpeechBalloonHelper {
+
+    public static final int CORNER_INSIDE  = -1;
+    public static final int CORNER_NONE    = 0;
+    public static final int CORNER_LEFT    = 1;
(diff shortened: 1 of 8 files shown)


osmdroid-android/src/main/java/org/osmdroid/util/SpeechBalloonHelper.java
```diff
@@ -0,0 +1,87 @@
...
+
+import org.osmdroid.views.overlay.SpeechBalloonOverlay;
+
+/**
+ * Helper class for {@link SpeechBalloonOverlay}
+ * whose main interest is method {@link #compute(RectL, PointL, double, PointL, PointL)}
+ * @since 6.1.1
+ * @author Fabrice Fontaine
+ */
+public class SpeechBalloonHelper {
```
spyhunter99: should this class be all static functions and drop the private variables? it seems like it's more of a utility class than anything else

monsieurtanuki: It's definitely a utility class. I don't have strong views on Singleton vs. static classes.
In that case, the key is the optimization: not recreating the `PointL` for every `draw`.

Issue #1385, opened 2019-08-27
Disabling/Enabling RotationOvelay causes rotation jumps


worldsnas:
## Issue Type

[ ] Question
[x] Bug
[ ] Improvement
[ ] Build system related
[ ] Performance
[ ] Documentation

## Description of the problem
if we add the rotation overlay but disable the overlay and set map rotation manually when we enable it again it jumps to the last rotation that was set by the user and wont follow the fixed rotation we applied on map and causes a UI jump

### If it's a bug, version(s) of android this affects:

All

### Version of osmdroid the issue relates to:

6.0.2

there is a very easy way to fix this and that is by passing mEnabled paramter to gestureDetector and let capture the user events but block publishing it (PR on it's way!).

spyhunter99:
merged

spyhunter99:
https://github.com/osmdroid/osmdroid/pull/1386

PR #1386, opened 2019-08-27
#1385 added new field to rotation gesture detector to watch overlay disable…

worldsnas:
fixed UI jumps caused by gesture detector when it's rotation overlay is disabled and re-enabled


spyhunter99:
do you think we need a getter for isEnabled?

PR #1408, opened 2019-09-29
bug/#1407 - simplified "speech balloon" demo

monsieurtanuki:
Impacted classes:
* `MyMath`: unrelated new precision in the javadoc of method `computeCirclePoint`
* `SampleSpeechBalloon`: removed all the `Timer` feature, which was irrelevant for the very demo and caused crashes in the demo app; minor refactoring

### DIFF
--- a/OpenStreetMapViewer/src/main/java/org/osmdroid/samplefragments/data/SampleSpeechBalloon.java
+++ b/OpenStreetMapViewer/src/main/java/org/osmdroid/samplefragments/data/SampleSpeechBalloon.java
@@ -14,8 +14,6 @@ import org.osmdroid.views.overlay.SpeechBalloonOverlay;
 
 import java.util.ArrayList;
 import java.util.List;
-import java.util.Timer;
-import java.util.TimerTask;
 
 /**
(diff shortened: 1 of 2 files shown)


spyhunter99:
i rather liked the animation but removing it is fine. i think all that was needed was some null checks and try/catch. no big deal

monsieurtanuki:
> i rather liked the animation but removing it is fine. i think all that was needed was some null checks and try/catch. no big deal

Initially that's how I planned to fix the code. But it didn't work. Then I removed the Timer altogether. It didn't work either and I got suspicious. It turned out Android Studio didn't compile my freshest code. Then I got busy with something else, turned off my computer, retried hours later and as the Timer-less version worked, I couldn't find the courage or the time to fiddle once again with `onDestroy` / `try` / `catch`.

monsieurtanuki:
@spyhunter99 Btw it's good to have Travis work again, thank you for that!

spyhunter99:
the avd emulator tests... i.e. connectedCheck...are disabled. i am going to
build a jenkins cluster eventually along with a collection of physical
devices. should be more reliable that travis

PR #1414, opened 2019-10-06
Avoid Zoom Buttons to Cancel Follow Location State.

InI4:
As a user of my own app I find it completely annoying, that everytime I use a (permanent) **zoom** button the state of **followLocation** is canceled. 

This is a slightly hacky solution to solve this issue: `CustomZoomButtonsController` and `MyLocationNewOverlay` interact now.

Note:
- I think in long term (code separation, etc.) it might be nicer to have zoom buttons an overlay itself, but I just got my issue solved. 
- I thought, there was a bug for it, but I failed to find it ...

Developed and tested on todays (6.1.3) snapshot.

### DIFF
--- a/osmdroid-android/src/main/java/org/osmdroid/views/overlay/mylocation/MyLocationNewOverlay.java
+++ b/osmdroid-android/src/main/java/org/osmdroid/views/overlay/mylocation/MyLocationNewOverlay.java
@@ -26,7 +26,6 @@ import org.osmdroid.views.Projection;
 import org.osmdroid.views.overlay.IOverlayMenuProvider;
 import org.osmdroid.views.overlay.Overlay;
 import org.osmdroid.views.overlay.Overlay.Snappable;
-import org.osmdroid.views.CustomZoomButtonsController;
 
 import java.util.LinkedList;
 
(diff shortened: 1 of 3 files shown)


osmdroid-android/src/main/java/org/osmdroid/views/overlay/mylocation/MyLocationNewOverlay.java
```diff
@@ -295,7 +296,10 @@ public boolean getEnableAutoStop(){
 	@Override
 	public boolean onTouchEvent(final MotionEvent event, final MapView mapView) {
 		if (event.getAction() == MotionEvent.ACTION_DOWN && enableAutoStop) {
-			this.disableFollowLocation();
+      CustomZoomButtonsController zoomController = mapView.getZoomController();
+      if ( zoomController != null && !zoomController.isHandling(event) ) {
+        this.disableFollowLocation();
+      }
```
monsieurtanuki: Too intrusive!

I think the solution would be something like adding those lines in `MapView.dispatchTouchEvent`, just after the "old zoom controller touch test":
````java
if (mZoomController != null && mZoomController.isTouched(event)) {
	return true;
}
````
And additionally make `CustomZoomButtonsController.isTouched` `public` and making it an always non-rotated version.

Could you have a go with that? My first test looks OK, but as the "follow" feature doesn't work on my old smartphone I cannot test it.

InI4: Thx @monsieurtanuki. 

I was looking for an less intrusive place, but didn't find it. I will check your suggestion! And come back later.

The `CustomZoomButtonsController.isTouched()` thing did not work! As `isTouched()` right now also immediately executes the touch, this is why I introduced that `isHandling()` thing doing the mere check without side effect. Might of course be unnecessary, when I can get the events really early. 

I will look into both.

InI4:
Your suggested change is much less cluttered and the UI is feelable more responsive. Waiting for double tap is not good for UIs.

However we have one minor "feature" here now. A long touch now creates a double zoom. 
I actually like it.

osmdroid-android/src/main/java/org/osmdroid/views/overlay/mylocation/MyLocationNewOverlay.java
InI4: Hmm, am I supposed to close this change discussion?

monsieurtanuki:
Eventually solved by #1419, but the initial bug detection and fix solution was here in #1414.

PR #1415, opened 2019-10-06
Fixed fling: The velocity vector needed some (un-)rotation.

InI4:
Just cause it came along while I was doing that scroll button thing. This fixes it for my device: When you fling the motion stays all in one direction, regardless of the actual rotation of the map.

### DIFF
--- a/osmdroid-android/src/main/java/org/osmdroid/views/MapView.java
+++ b/osmdroid-android/src/main/java/org/osmdroid/views/MapView.java
@@ -1548,8 +1548,10 @@ public class MapView extends ViewGroup implements IMapView,
 			}
 			mIsFlinging = true;
 			if (mScroller!=null) {  //fix for edit mode in the IDE
-				mScroller.fling((int) getMapScrollX(), (int) getMapScrollY(), (int) -velocityX,
-						(int) -velocityY, Integer.MIN_VALUE, Integer.MAX_VALUE, Integer.MIN_VALUE, Integer.MAX_VALUE);
+        Point p = getProjection().unrotateAndScalePoint((int)velocityX, (int)velocityY, null);
+				mScroller.fling((int) getMapScrollX(), (int) getMapScrollY(),
+            (int) -p.x, (int) -p.y,
+            Integer.MIN_VALUE, Integer.MAX_VALUE, Integer.MIN_VALUE, Integer.MAX_VALUE);
 			}
 			return true;
 		}
--- a/osmdroid-android/src/main/java/org/osmdroid/views/MapView.java
+++ b/osmdroid-android/src/main/java/org/osmdroid/views/MapView.java
@@ -1548,9 +1548,13 @@ public class MapView extends ViewGroup implements IMapView,
 			}
 			mIsFlinging = true;
 			if (mScroller!=null) {  //fix for edit mode in the IDE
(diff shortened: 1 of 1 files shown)


InI4:
Wait a minute! 

The patch is fine on my new phone, but testing it on my old one only now, flinging is now wrong.

So it seems like the orientation of velocity in `OnGestureListener.onFling()` or in `Scroller.fling()` is different between the two devices. Could be an Android version thing (28 vs 17 here), could be something different. So far I failed to google it. 

Do you know something?
Is a PR feasible which is good for new and bad for old devices? 
Is it something else?

InI4:
I just updated the PR with a version with an explicit check for Android build 28. 
This number has been determined by tests with emulators here. 
Good news: It works in here in all cases now.
Bad news: No reference where the observed behaviour comes from.

InI4:
Oops, I apologize I missed your comment about the point object. Of course I can do this with a statically allocated "flingPoint" or see, if there are some other temp Points free in that class ...

monsieurtanuki:
@InI4 Oh if it's used only once in a while and not for every pixel step of the fling, we can afford the current Point creation in the method. Can you check that?

InI4:
OK, now here are holiday obligations. I will try to check later this year ..

InI4:
@monsieurtanuki I am messed up here, please advise!

- I did the tests: The call is rare and the point probably well spent.
- However I cannot test flings in an emulator and my phone got updated to API 29 these days.
  -The other fling testresults are confusing:
    - Phone API 28, local built Osmdroid 6.1.3 does not work.
    - Phone API 28, local build Osmdroid 6.1.3 with patch works.
    - Phone API 29, local build 6.1.6 with patch does not work.
    - Phone API 29, local build 6.1.6 with disabled patch does work!
  - Possible explanations: 
    - API change mattered something on my phone (but shouldn't other apps be affected then).
    - 6.1.6 does something to flings elsewhere.

Right now, I am not sure about the patch anymore. More testers with API 28/29 might help. I cannot downgrade my phone to redo 6.1.6. tests.

CC: @jlimek can you add observations with different 6.1.6 versions?

monsieurtanuki:
@InI4 If I sum up, you created a patch that finally may happen to be relevant only in limited cases (API 28), that may not be relevant any more (API 29) and for which we couldn't find bug/fix siblings on the net.

Of course further tests would be welcomed.

But in the meanwhile, removing your patch would make sense.
An alternate solution: create a public method in `MapView` called `setBehavior(long)` with flags, and `public static final long BEHAVIOR_FLAG_FLING_ROTATION = 1`. The default action for the fling being the code before your patch, and if the flag is set (which you can do on demand with `setBehavior`, possibly when API == 28), then your patch code is used. What do you think of that?

InI4:
@monsieurtanuki a extra flag would be an approach.  Though right now I could also go from `>= 28` to `== 28` in the patch.

More tests would be good, I will ask around!

InI4:
I got some feedback from my userbase. For many users with API 28 the patch works, no feedback on the API 29 level. So maybe the modification should be rather == 28 than 29. This needs a new pull request, doesn't it?

monsieurtanuki:
@InI4 Indeed. If I understand well the PR will be `if (Build.VERSION.SDK_INT == 28)` instead of `if (Build.VERSION.SDK_INT >= 28)`, am I correct? If so, please send the PR.

PR #1419, opened 2019-10-09
feature/#1418 - custom zoom refactoring while removing the old standard zoom controllers

monsieurtanuki:
Impacted classes:
* `AsyncTaskDemoFragment`, `CustomRecycler`, `SampleFollowMe`, `SampleHeadingCompassUp`, `SampleSplitScreen`, `WeathForceActivity`: removed a useless call to now deprecated method `MapView.setBuiltInZoomControls`, which used to be the default behavior anyway
* `MapView`: removed all the references on the deprecated Android zoom controls; moved the calls to zoom control test in `dispatchTouchEvent`; deprecated method `setBuiltInZoomControls`; unrelated minor refactoring
* `CustomZoomButtonsController`: made `public` method `isTouched` which uses new method `CustomZoomButtonsDisplay.isTouched`; deprecated method `onSingleTapConfirmed` and `onLongPress`
* `CustomZoomButtonsDisplay`: deprecated method `isTouchedRotated` as we now check the touch event before applying the rotation; created method `isTouched(MotionEvent,boolean)` for the same reasons; unrelated minor refactoring

### DIFF
--- a/OpenStreetMapViewer/src/main/java/org/osmdroid/bugtestfragments/WeathForceActivity.java
+++ b/OpenStreetMapViewer/src/main/java/org/osmdroid/bugtestfragments/WeathForceActivity.java
@@ -95,7 +95,6 @@ public class WeathForceActivity extends BaseActivity implements LocationListener
         mLocationOverlay.enableFollowLocation();
         mLocationOverlay.enableMyLocation();
         this.mMapView.getOverlayManager().add(mLocationOverlay);
-        mMapView.setBuiltInZoomControls(true);
         mMapView.setMultiTouchControls(true);
         mMapView.setTilesScaledToDpi(true);
     }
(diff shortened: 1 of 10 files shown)


InI4:
Works for me. I.e. I do not fully understand it, but it solves the issues solved by #1414 .

Remark: I needed to run the build with 
```
gradlew -x testReleaseUnitTest :osmdroid-android:publishToMavenLocal
```
to sidestep failing tests.

```
org.osmdroid.util.GarbageCollectorTest > testSecond FAILED
    junit.framework.AssertionFailedError at GarbageCollectorTest.java:51
```

monsieurtanuki:
@InI4 Actually it works because it's just what you coded in #1414 ("moving the zoom ontouch detection test"), but with more impact checks and refactoring.
I thought it made no sense to accept #1414 as it was (didn't work with orientated maps for instance) nor to reject it again and again with new justifications ("oh you should also do this and that"). My point was not to take the credits of this fix from you, it was to fix comprehensively the custom zoom controller that I painfully coded.

That being said, I've had a look at the (unrelated) crashing unit test.
It works OK for me, and it looks like "bad luck crash" in your case.
There are threads involved in this unit test, one being with low priority, and the unit test is about checking if the thread is over after n milliseconds. In your case I guess that the thread started later (because of the low priority), therefore is not over after n milliseconds of the main thread. Thence the crash.
I'm working on it.

monsieurtanuki:
@InI4 I've just fixed the unit test. Can you please try again?

InI4:
I can confirm, the unit test is now also running smoothly here.

PR #1421, opened 2019-10-10
Add optional language parameter to MapsForgeTileSource

### LINKED ISSUE #1420: Mapsforge adapter: multi-language support
## Issue Type

[x] Question

## Description

Mapsforge supports multi-language maps, whcih allows to set the map language (see [here](https://github.com/mapsforge/mapsforge/blob/c8294c7ae27b2ce4a6af56516216af8c6d7cb0b3/mapsforge-map-reader/src/main/java/org/mapsforge/map/reader/MapFile.java#L217) and [here](https://github.com/mapsforge/mapsforge/blob/c8294c7ae27b2ce4a6af56516216af8c6d7cb0b3/mapsforge-samples-android/src/main/java/org/mapsforge/samples/android/MultiLingualMapViewer.java)). Is it possible to achieve this using the Mapsforge adapter ?

6.1.2

sgelb:
Adds an optional language paramter to MapsForgeTileSource. This allows to set a prefered language for labels, which in most cases default to the map location's local language.

This sets one language for all map sources. Theoretically it is possible to set a prefered language per map source. I left this out for 1) simplicity and 2) because I don't see much use for this.

Fixes #1420

### DIFF
--- a/osmdroid-mapsforge/src/main/java/org/osmdroid/mapsforge/MapsForgeTileSource.java
+++ b/osmdroid-mapsforge/src/main/java/org/osmdroid/mapsforge/MapsForgeTileSource.java
@@ -56,15 +56,16 @@ public class MapsForgeTileSource extends BitmapTileSourceBase {
      * @param maxZoom
      * @param tileSizePixels
      * @param file
+     * @param language
      * @param xmlRenderTheme the theme to render tiles with
      * @param hillsRenderConfig the hillshading setup to be used (can be null)
      */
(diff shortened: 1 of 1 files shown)


sgelb:
Thank you for your feedback, those are all valid points. I'll push an updated PR later this week.

sgelb:
In the developer guidelines it says to use 100 character line width and to prefix method arguments with p. Most parts of `MapsForgeTileSource.java` don't follow those rules. How should I proceed?

spyhunter99:
don't worry about it

sgelb:
No problem, I'll update the PR with that refactoring.

monsieurtanuki:
@sgelb Only if that makes sense to you and if you have time. I don't mind if you don't edit it - and that would also save me the time to re-approve it ;)

sgelb:
@monsieurtanuki Okay, if that is fine, then I'm done with it. Thanks!

PR #1424, opened 2019-10-12
feature/#1412: Implement setPadding for CustomZoomButtonsDisplay

wharnisch:
feature/#1412

### DIFF
--- a/osmdroid-android/src/main/java/org/osmdroid/views/CustomZoomButtonsDisplay.java
+++ b/osmdroid-android/src/main/java/org/osmdroid/views/CustomZoomButtonsDisplay.java
@@ -32,6 +32,10 @@ public class CustomZoomButtonsDisplay {
 	private boolean mHorizontalOrVertical;
 	private float mMargin; // as fraction of the bitmap size
 	private float mPadding; // as fraction of the bitmap size
+	private int mPaddingLeft = 0; // in pixels
+	private int mPaddingTop = 0; // in pixels
+	private int mPaddingRight = 0; // in pixels
+	private int mPaddingBottom = 0;	// in pixels
 
 	public CustomZoomButtonsDisplay(final MapView pMapView) {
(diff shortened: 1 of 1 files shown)


wharnisch:
Class _GoogleMaps_ in Google Maps API offers a method _setPadding_ with exactly the same parameters and functionality as the method I suggested. The 'object' is the whole view of CustomZoomButtonsDisplay, so the padding is inside the object. Therefore I thought _setPadding_ would be a suitable name for the method. Nevertheless I will try to change things like you commented.

monsieurtanuki:
Let me have a look at the GoogleMaps' `setPadding` method.
That would make sense to match GoogleMaps method name. Assuming they're right ;)
But anyway in the current code the margin and the padding have another meaning, and we need to have consistent names. I'll get back to you.

monsieurtanuki:
I confirm I'm right: in Google Maps, [`setPadding`](https://developers.google.com/android/reference/com/google/android/gms/maps/GoogleMap#setPadding(int,%20int,%20int,%20int)) applies to a **map** and indicates the "authorized rectangle" of the map. And then, all the controls need to be in this zone, which would in your case move the zoom controls. That's what you initially wanted to do, moving the zoom controls, fair enough.

But in osmdroid, we're dealing with a method for the **zoom controls**, not for the **map** in general. Therefore in our case it's more `zoomDisplay.setMargins`.
It would be `mapView.setPaddings` if we implemented in the same spirit as GoogleMaps ("which is the authorized zone for the controls of the map?"). That would make sense too, but it's not what you've been implementing.

wharnisch:
Sorry, did not manage to change pull request, created a new pull request (Implement setAdditionalPixelMargins for CustomZoomButtonsDisplay #1426)

monsieurtanuki:
@wharnisch No problem, I know how git can be confusing and stressful. I learned it the hard way. Still learning btw ;)

PR #1436, opened 2019-11-15
Bing map upgrade authentication and tile URL to https

jpstotz:
The Bing maps authentication URL that is used for sending the application's Bing Maps Key to Microsoft was still using unencrypted HTTP, instead of HTTPS. 

I have verified that the URL via HTTPS works as expected, it was correctly returning a valid response using my personal Bing Maps key for the imagery-set "Road".


jpstotz:
>All checks have failed
Looks like the osmdroid ci build system has some serious problems:

It tries to load gradle from `http://services.gradle.org/distributions/gradle-2.14.1-bin.zip` -> HTTP 403

The gradle project seems to disallows downloads via unprotected http (totally makes sense to me especially in the context of a ci build system). Change the URL to https:// and the download works.

jpstotz:
Once my other PR #1437 has been applied the CI system will be working again and the "All checks have failed" flag should vanish.

PR #1440, opened 2019-11-20
Advanced Polyline drawing (feature/#1378)

mat8854:
Hello, these are my changes for the feature to draw poylines with borders and color mappings.

# Changes
* I put the color mapping code into a separate sub package `advancedpolyline`
* Color mappings are:
1. `ColorMappingPlain`
2. `ColorMappingCycle`
3. `ColorMappingRanges`
4. `ColorMappingVariationHue`, `ColorMappingVariationSaturation` and `ColorMappingVariationLuminance` (using HSL http://hslpicker.com/)
* I had to change the `PointAccepter` with a third parameter to preserve index. In the previous code at some point the information about the current position of a line point was just dropped.
* Changes to `Polyline` and `PolyOverlayWithIW` with added functions to add points with a scalar for color mapping.
* Updated `LineBuilder` and `LineDrawer` with logic for indexes and to check for new polyline style set.

# Usage
Sample code below adds five points. Color mapping will vary hue from red to green.
```
val line1 = Polyline()
line1.width = 20.0f
val style1 = PolylineStyle(ColorMappingVariationHue(0.0f, 50.0f, 0.0f, 120.0f, 1.0f, 0.5f), true)
style1.setBorder(25.0f, Color.BLACK)
line1.style = style1
line1.addPoint(GeoPoint(34.0, -118.0), 0.0f)
line1.addPoint(GeoPoint(35.0, -119.0), 10.0f)
line1.addPoint(GeoPoint(36.0, -118.0), 35.0f)
line1.addPoint(GeoPoint(37.0, -119.0), 40.0f)
line1.addPoint(GeoPoint(38.0, -118.0), 23.0f)
map.overlayManager.add(line1)
```

# Known issues
* The style should be set before adding any points. With a set style the user should always call the add point functions with an additional scalar.


monsieurtanuki:
@mat8854 Great new feature! Unfortunately I have currently a very limited access to internet that prevents me from testing it in the following weeks.
What about the demo? I think I'll create one around the first stage of the Camino de Santiago de Compostela (camino francés: from St-Jean-Pied-de-Port to Roncesvalles).

mat8854:
So I added GPX / KML data I found at www.gpsies.com, www.flightaware.com and self-recorded data. 

There still seems to be a minor bug. The line is not drawn as one big line:
Screenshot_1574809651

The borders move if you move the map. I am pretty sure it is some index offset mistake in the added code. The line drawer class has this constant of 256 set in the code `mLineDrawer = new LineDrawer(256);` in `PolyOverlayWithIW` (?). Otherwise its stable.

I opened the PR so other people can try out the feature and give feedback.

monsieurtanuki:
@mat8854 If I remember well the point of the "256" limit was performances.
We draw segments, computed from geopoints.
If we use the drawLines syntax, we need a limit for our array (256).
As we use the same color for all segments, we can safely say that the whole segments is a combination of several 256-point-max segments.
But with your new feature I understand there are side effects. Should we accept 1000 or no limit instead of 256: I don't think so, in first approach.
I know that I also implemented Polyline display as Path (and not drawLines): would that be helpful in your case?
Nice map by the way ;)

mat8854:
@monsieurtanuki  The limit 256 means 64 line segments if I understand it correctly. 

Yes even updating to 1000 makes no sense. The limit of 256 can stay the same. I can change the code to repaint the overdrawn spots. It becomes a little hacky.

I also look into this `Path` option. Right now the feature only works for `pUsePath = false`.

monsieurtanuki:
@mat8854 Perhaps a better option would be to add the following parameters to the drawLines(256) part:
* is it the very start?
* is it the very end?

With that in mind, you're able to decide how to display the first and the last of the 256(/64) segments without having to repaint on it afterwards.
Is that possible?

mat8854:
@monsieurtanuki Yes, only the start / beginning line segment must be redrawn afterwards. I hacked in a list to store the specific line segments and paint information.

For Polylines without border the current code works fine.

monsieurtanuki:
@mat8854 we'll always have the "alpha" issue with drawLines: if we use an "alpha'ed" Paint, as we paint twice the vertices therefore twice an alpha'ed spot the color on each vertex looks different (e.g darker). This is not specific to your new feature - though it's getting worse if you also repaint some start/end segments.
Does your new feature work with Path, too?
That being said, we can live with that restriction: darker vertices when using an alpha'ed Paint.

mat8854:
@monsieurtanuki Ah okay I see. I tested the feature mostly for polylines without border and no alpha for speed ranges. 

Actually other Libraries should face the same problem drawing polyline on canvases. I even looked deeper for modifactions to the Android Paint object (Shaders, Stroke / Fill, ...) for the border problem with out a good solution. Thats why border options uses two complete polylines now.

To make it  complete I will look into this Path draw option and alpha handling during then next days ...

mat8854:
* So with the last commit https://github.com/osmdroid/osmdroid/pull/1440/commits/e74735c87e0c831abe308c728bac34da42b117de the border painting problem is solved:
Screenshot_1575034787

I wanted to keep this feature state with advanced styling *only* works for `pUsePath = false`.

mat8854:
@monsieurtanuki Thanks, I will look over your code clean up suggestions.

mat8854:
@monsieurtanuki I was thinking about getting rid of mBorderMode. But `LineDrawer` has no information if it is drawing the border or the line when being called. The Paint object for the border is set all the time in `PolylineStyle` object.

I understand the other improvements for `PointAccepter` and so on.

monsieurtanuki:
@mat8854 I think we should think a bit more "meta" about `LineDrawer`.
Before your new features, what was the purpose of a `LineDrawer`?
* absorbing `Point`s up to 256 / 2 (method `LineBuilder.add`)
* drawing the corresponding segments with a unique `Paint` (method `flush`)

What you introduced was:
* the possibility to draw several times (e.g. twice) the same segments, with a different `Paint` (border and not border)
* the possibility to draw each segment with a different color (in order to show a progression, an elevation, ...)

This is what I suggest:
1. create an interface called `PaintList`, with methods `Paint getPaint(final int pIndex)` and `boolean isMonochromatic`
1. create a class called `UniquePaintList` that implements `PaintList` for the unique `Paint` case (`getPaint` returning always the same `Paint` as in the constructor, and `isMonochromatic` returning `true`)
1. in class `LineDrawer`, create a `List<PaintList>` member called `mPaintLists`, with a setter
1. in class `LineDrawer`, remove the `Paint mPaint` member and code current method `setPaint(Paint)` as setting `mPaintLists` as a list of one `UniquePaintList`
1. in class `LineDrawer.flush`, loop on `mPaintLists`: if the `PaintList` item is monochromatic, use `drawLines` with all the segments and with `PaintList.getPaint(0)`, if not, use a loop on each segment and use something like `PaintList.getPaint(i)`

Doing so:
* we have a clearer definition of a `LineDrawer`, that draws its segments with all the `Paint`s it's been given. No special case like border or `Paint` vs. `PolylineStyle`: this is managed by the `List<PaintList>`.
* we have the possibility to have borders with "evolving" `Paint`s (in your case you assume that borders are always monochromatic)
* we can use as many `Paint` layers as we want, like a 2 pixel red line in the middle, embedded in a 6 pixel white line, embedded in a 10 pixel blue line

What do you think of that?

mat8854:
@monsieurtanuki Yes, I also did not the hacky `LineDrawer` class. But it worked. 

Actually the code was moved around several times during development where to put the information and so on. The refinements for futher abstraction are good.

mat8854:
@monsieurtanuki  If I understand you correctly you want to keep a list of Paints / Colors in the `LineDrawer`.

The double call to `buildLinePortion()` from `drawWithLines()` in `PolyOverlayIW` will still be present and must be kept (!). There must be a logic to paint the "Border-Polyline(s)" first completely.

monsieurtanuki:
@mat8854 You don't have to store 64 Paints. All we need is that PaintList.getPaint(n) returns the correct Paint for the nth segment. Obviously for UniquePaintList it will return the same object. And that could be the same for other PaintList's: keeping an array of 64 colors, setting the nth color to the same Paint object, and returning it.

If we find a smart way, we may even imagine handling the "border side-effect" case within LineDrawer, in the flush method. Something like: flush was called because 256 points were added, therefore I keep the last pair of Point's and when the next pair of Point's is added, I can automatically draw the extra segment. Or something like that.

monsieurtanuki:
Alternate solution: using a List of LineDrawer's in PolyOverlayWithIW.
The side effect would be handled automatically if you draw with one LineDrawer then the other. But I'm not sure about the current code - actually I'm cooking...

mat8854:
@monsieurtanuki Okay, no I prefer keeping one `LineDrawer` class and using lists of Paints. The code in `flush()` gets already more compact. Only two statements calling `mCanvas.drawLines()` (one for monochromatic and one for segment by segment).

mat8854:
@monsieurtanuki Hello, you suggested to add a separate `add()` method to `PointAcceptor`:
```
public interface PointAccepter {

    void init();
    void add(final long pX, final long pY);
    void add(final long pX, final long pY, final int index);
    void end();
}
```
Any user has to implement both methods. Does it get really simpler? If the index has no meaning the user should just pass `0`.

monsieurtanuki:
@mat8854 Actually you removed add(long,long) and replaced it with add(long,long,int). But we're talking about public methods: in osmdroid itself we know who uses PointAccepter, but projects using osmdroid might have used PointAccepter.add(long,long) in their own code.
And it's not pretty either to say "oh this parameter? Just put 0 most of the time".
Something cleaner would be something like interface PointAccepterWithParam<T> extends PointAccepter, with add(long,long,T) because in your case your extra parameter is an int, but it could be anything. And keep PointAccepter as it was, without impact.
I think it is possible: can you please try that?

mat8854:
@monsieurtanuki I was looking at the call stack when points are added:

call_stack

Classes in the call are only: `LinearRing` (and `PathBuilder`), `SegmentClipper` and `LineBuilder`. I changed them using the new interface. `MilestoneXYZ` classes are changed back to old interface.

monsieurtanuki:
@mat8854 Currently reviewing...
There's at least one thing I'm not satisfied with: `PointAccepter` becoming `PointAccepterWithParam`, and the impacts.
I think there's something cleaner to do.
I'll keep you posted.

monsieurtanuki:
A lot of code to read...

Quick win: transform your `ColorHelper.HSLToColor(float[])` into `ColorHelper.HSLToColor(float h, float s, float l)`, in order to avoid to create an array each time.

Random thought: I didn't see where it made sense to keep the scalar value in the `ArrayList`. What we need is to say "the nth segment should have the nth color" (even with the `Shader`), but at that point we don't care about the scalar, do we?

I think the key would be to kick `item.setReferencesForAdvancedStyling(getColorIndexes(), getLines());` out of `LineDrawer`.
The `PaintList` should be able to know which index we're talking about, beyond the 256 (or 64) limit. That may mean something like `PolylineStyle` implementing `PaintList`.

Class `LineDrawer` would look something like that:
````java
public class LineDrawer extends LineBuilder{

    private Canvas mCanvas;
    private Paint mPaint;
    private PaintList mPaintList;

    public LineDrawer(int pMaxSize) {
        super(pMaxSize);
    }

    public void setCanvas(final Canvas pCanvas) {
        mCanvas = pCanvas;
    }

    public void setPaint(final Paint pPaint) {
        mPaint = pPaint;
    }

    public void setPaint(final PaintList pPaintList) {
        mPaint = null;
        mPaintList = pPaintList;
    }

    @Override
    public void flush() {
        if (getSize() < 4) {
            return;
        }
        if (mPaint != null) {
            mCanvas.drawLines(getLines(), 0, getSize(), mPaint);
            return;
        }
        if (mPaintList.isMonochromatic()) {
            mCanvas.drawLines(getLines(), 0, getSize(), mPaintList.getPaint(0));
            return;
        }
        for (int i = 0; i < getSize() / 4 ; i ++) {
            final int offset = (getOverallSize() - getSize()) / 4;
            mCanvas.drawLines(getLines(), i * 4, 4, mPaintList.getPaint(offset + i));
        }
    }
}
````

And those are edits for `LineBuilder`:
````java
    /**
     * Overall number of added items
     */
    private int mOverallIndex;

    @Override
    public void init() {
        mIndex = 0;
        mOverallIndex = 0; // added
    }

    @Override
    public void add(final long pX, final long pY) {
        mLines[mIndex ++] = pX;
        mLines[mIndex ++] = pY;
        mOverallIndex += 2; // added
        if (mIndex >= mLines.length) {
            innerFlush();
        }
    }

    /**
     * added
     */
    public int getOverallSize() {
        return mOverallIndex;
    }
````

I don't know if you see where I'm going (I don't either): I want to get rid of `PointAccepterWithParam` and its impacts in the existing code. Given that in your case, the param is more or less the index of the segment.

monsieurtanuki:
@mat8854 I think I have just solved everything. Details later, perhaps today, perhaps tomorrow.

mat8854:
@monsieurtanuki Good, I do not want this feature to be over-engineered. If the final merge back happens until the end of December that would be great.

monsieurtanuki:
@mat8854 I wrote a [wiki page](https://github.com/osmdroid/osmdroid/wiki/Polyline-advanced-display-styles) in order to explain the concepts of the advanced styles. It's based on your code, with simplifications on top.
Tell me If I missed anything important in that wiki.
As time is a factor and as I've spent the last days studying thoroughly your code, I suggest I rewrite part of your code (and then you review it), so that we can wrap things up quickly, perhaps by the end of the week. I'm waiting for your go/nogo.

mat8854:
@monsieurtanuki Okay. Yes, post your modifictions, I look over them and push them on this branch.

monsieurtanuki:
@mat8854 If you want to have a look at the code, you can do it now.
I won't change that much to the code (but I will: there's a side effect I need to have a look at in `IntegerAccepter`).

monsieurtanuki:
Latest test: a closed polyline with gradients of color
Screenshot_2019-12-18-11-34-25

mat8854:
Great, it is good now. It can be merged back.

PR #1445, opened 2019-12-07
bug/#1444 - setting back the correct icon bounds in ItemizedOverlay

monsieurtanuki:
Impacted class:
* `ItemizedOverlay`: set back the correct icon bounds after they were changed for the display part


monsieurtanuki:
I could not reproduce the issue, but I experienced another bug (icon and text flying away in `SampleMilitaryIconsItemizedIcons`). Maybe my fix fixes this issue too.

PR #1459, opened 2019-12-22
feature/#1409 - downgrade mode for PolyOverlayWithIW when the projected poly is too small

monsieurtanuki:
The downgrade management works with 2 methods:
* `PolyOverlayWithIW.setDowngradeMaximumPixelSize(int)` - if the poly projected at the screen is too small (width or height lower than this value), either we don't display the poly or we display a rectangle instead
* `PolyOverlayWithIW.setDowngradeDisplay(boolean)` - should we display the rectangle when the poly is too small?

It can be tested in `SampleShapeFile`.

Deleted class:
* `PathOverlay`, which was deprecated 4 years ago

Impacted classes:
* `BoundBoxTest`: unrelated refactoring
* `Counters`: added a more flexible access to counter features with methods `reset(String)`, `increment(String)` and `get(String)`
* `CustomPaintingSurface`: used the new method `PolyOverlayWithIW.getBounds`; unrelated performance refactoring
* `DefaultOverlayManager`: unrelated display optimization for MapSnapshots
* `LinearRing`: now computing the bounding box ASAP, in method `computeProjected`, which simplifies method `getCenter`; created another version of `getBestOffset`; created method `getCloserValue`; created methods `getBoundingBox` and `clear`
* `LineDrawer`: optimized the display
* `MilestoneLineDisplayer`: fixed a minor unrelated bug, where the last point was added as (X0,Y0), but without a correspnding (X1,Y1)
* `OsmMapShapeConverter`: used method `PolyOverlayWithIW.getActualPoints` instead of deprecated `Polygon.getPoints()`
* `Polygon`: deprecated method `getPoints()`, to be replaced by `PolyOverlayWithIW.getActualPoints`; moved methods `setPoints` and `addPoint` to `PolyOverlayWithIW`
* `Polyline`: deprecated method `getPoints()`, to be replaced by `PolyOverlayWithIW.getActualPoints`; moved methods `setPoints` and `addPoint` to `PolyOverlayWithIW`
* `PolyOverlayWithIW`: added in the `draw` the downgrade management; created methods `setDowngradeMaximumPixelSize`, `setDowngradeDisplay`, `isWorthDisplaying` and `displayDowngrade` for the downgrade management; moved `Polygon` and `Polyline` methods here - `getBounds`, `setPoints`, `addPoint` and `getActualPoints`
* `Projection`: computed `mProjectedMapSize` regardless of the tile size; created method `getWorldMapSize()`
* `SampleOsmPath`: removed the "deprecated 4 years ago" `PathOverlay` sample; refactored
* `SampleShapeFile`: set up the display (downgrade and style)
* `ShapeConverter`: used the new method `PolyOverlayWithIW.getBounds`
* `TileSystem`: deprecated now useless member `projectionZoomLevel`


monsieurtanuki:
@spyhunter99 Tell me if if works OK for you.

monsieurtanuki:
@spyhunter99 A better approximation than the default rectangle could be computed:
* we have all the Polygon coordinates as projected on a very big world map
* let's say that the target minimum size is 20 pixels - as a pre-check if the projection of the bounding box is below that size, we switch to the simplified version
* with a simple homothetic transformation, we can transform the very big world map coordinates into the small rectangle coordinates, with of course MANY duplicates that we remove
* doing so we compute once and for all a tiny Path, to be quickly drawn if needed

Same thing for the Polylines.

PR #1460, opened 2019-12-23
bug/#1360 - refactored icon anchoring in MyLocationNewOverlay

monsieurtanuki:
Impacted class:
* `MyLocationNewOverlay`: created methods `setPersonAnchor` and `setDirectionAnchor` while deprecating method `setPersonHotspot`; created method `setDirectionIcon` for consistency while deprecating method `setDirectionArrow`; refactored the constructor accordingly


spyhunter99:
working the merge issues now

spyhunter99:
so i did a test merge, the sample 'custom my location overlay' was waaaaay off....but so is master. I'm not sure what happened to it. Looks like the icon was resized and/or scaled and the scaling did not account for screen resolution or something

spyhunter99:
looks like the way off was introduced in 6.1.4

monsieurtanuki:
@spyhunter99 Actually it was just some refactoring when I was idle, not a fix.

Perhaps a sensitive way to handle this would be to split the code, creating a class that "just" displays the circle and the icon for a given `Location` and screen density - beyond the refactoring step, this is where the current issue probably is. And then `MyLocationNewOverlay` as a `IMyLocationConsumer` that calls that new class display method. In short: the GPS is not reliable on my smartphone and I cannot test it!

Issue #1458, opened 2019-12-22
Polygon: possible display with drawLines


monsieurtanuki:
## Issue Type

- [x] Improvement
- [x] Performance

## Description and/or steps/code to reproduce the problem

For `Polyline`s display, we can say if we want to use `Canvas.drawLines` (which is faster) or `Canvas.drawPath` (which is slower, but has added features).
For instance, if you want to draw with an "alpha-ed" Paint, it will look perfect with `drawPath`, but not nice with `drawLines` where some vertices will have several times the same "alpha-ed" Paint, therefore will look darker.

For `Polygon`s, we systematically use `Canvas.drawPath`.

We should be able to decide to use `drawLines` for `Polygon`s, if we want a faster display and if we don't use `Path` specific features like fill paint or holes.

Should be done after #1409.

## Environment

6.1.4

monsieurtanuki:
Done in #1483.

Issue #1447, opened 2019-12-07
refactor density calls


pasniak:
Maybe replace `mapView.getContext().getResources().getDisplayMetrics().density` chain-calls with a helper `mapView.getDisplayDensity()` ? This would make few lines shorter.

## Issue Type

[x] Improvement

### Version of osmdroid the issue relates to:

6.1.2

spyhunter99:
closing, looks like this was merged

Issue #1446, opened 2019-12-07
java.util.ConcurrentModificationException at org.osmdroid.tileprovider.MapTileProviderBase.mapTileRequestFailed(MapTileProviderBase.java:184)


pasniak:
## Issue Type

[ ] Question
[x] Bug
## Description and/or steps/code to reproduce the problem

java.util.ConcurrentModificationException
at java.util.LinkedHashMap$LinkedHashIterator.nextEntry(LinkedHashMap.java:346)
at java.util.LinkedHashMap$KeyIterator.next(LinkedHashMap.java:366)
at org.osmdroid.tileprovider.MapTileProviderBase.mapTileRequestFailed(MapTileProviderBase.java:184)
at org.osmdroid.tileprovider.MapTileProviderArray.mapTileRequestFailedExceedsMaxQueueSize(MapTileProviderArray.java:167)
at org.osmdroid.tileprovider.modules.MapTileModuleProviderBase$1.removeEldestEntry(MapTileModuleProviderBase.java:124)
at java.util.LinkedHashMap.addNewEntry(LinkedHashMap.java:189)
at java.util.HashMap.put(HashMap.java:404)
at org.osmdroid.tileprovider.modules.MapTileModuleProviderBase.loadMapTileAsync(MapTileModuleProviderBase.java:151)
at org.osmdroid.tileprovider.MapTileProviderArray.runAsyncNextProvider(MapTileProviderArray.java:227)
at org.osmdroid.tileprovider.MapTileProviderArray.mapTileRequestExpiredTile(MapTileProviderArray.java:179)
at org.osmdroid.tileprovider.modules.MapTileModuleProviderBase$TileLoader.tileLoadedScaled(MapTileModuleProviderBase.java:299)
at org.osmdroid.tileprovider.modules.MapTileModuleProviderBase$TileLoader.run(MapTileModuleProviderBase.java:345)
at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1113)
at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:588)
at java.lang.Thread.run(Thread.java:818)

My partner was trying my app and the map froze... after a while we got the stack above.  

I am at c245bcd129463a344606f24213acd4471dd5d35c
#906 restoring polyline addPoint function Alex O'Ree 10/26/19 12:27 PM

monsieurtanuki:
[...]

Issue #1409, opened 2019-09-29
polygon drawing optimizations at low numerical zoom levels


spyhunter99:
## Issue Type

[ ] Performance

## description

see pr https://github.com/osmdroid/osmdroid/pull/937 and #906 for reference.

When there is a sufficient quantity of small polygons on screen and the user zooms out, draw speed can drop to a crawl. https://github.com/osmdroid/osmdroid/pull/937 mitigates this somewhat by enabling and using the feature requested by @MKergall which uses an overlay bounds as a faster mechanism to skip draw cycles on overlays that are not in view. This makes rendering quick if nothing is in view.

i think using the point reducer at lower numerical zoom levels may reduce complex polygons from potentially 1000s of points down to 4 or less, which should help draw speed significantly.

the reduced point set can be cached for panning operations to reducing the recalculation necessary and the cache invalidated on zoom events.

## to reproduce...
 
see pr https://github.com/osmdroid/osmdroid/pull/937 for esri shape file download link, then import the shape file, specifically the "shapes" within the shape file. after its imported, the performance issue is obvious

monsieurtanuki:
I managed to reproduce the poor performances using https://catalog.data.gov/dataset/military-installations-ranges-and-training-areas
A solution could also be to use a `PointReducer` for pixels too, not only for `GeoPoint`s.
In order to avoid confusion and painful code merge, I'll work on that issue after #1440 is finished.

monsieurtanuki:
... and now, #1440 is merged. I can start!

monsieurtanuki:
For what it's worth, I've just run some tests on my smartphone. We're talking about 2035 ESRI Polygons, to be displayed in the world map.

Test 1 was the target test, without editing anything.
The very first draw took 3890ms, and the next draw 830ms.
The difference comes from the fact that the first draw initializes data (for instance the projections of GeoPoint into a virtual screen).

Test 2: I removed the actual "canvas.drawPath" call, in order to understand if we are slow computing things or Android is slow drawing Paths.
The very first draw took 3980ms, and the next draw 730ms.
Very similar to test 1, because the population of the Path is already optimized (we remove consecutive duplicates points).

Test 3: I removed the "display this Polygon" calls.
The very first draw took 20ms, and the next draw 30ms.
Same duration. We don't compute data for each draw, and don't even initialize any.

Test 4, 5 and 6 are the same as 1, 2 and 3, but pretending that we use `Polyline`s instead of `Polygon`s.

[...]

PR #1482, opened 2020-01-11
feature/#1480 - new method PolyOverlayWithIW.setDowngradePixelSizes

monsieurtanuki:
If the poly is smaller (in pixels) than size1, draw a precomputed simplified list of segments.
If it's even smaller (than size2), draw even faster the poly as a rectangle.
Of course this is only optional; the default being to always display the regular poly.
The demo is included in `SampleShapeFile`, and the important methods here are `setDowngradeDisplay` and `setDowngradePixelSizes`.

New class:
* `ListPointAccepter`: a `PointAccepter` that builds a `List` of `PointL` as a list of long, long

Impacted classes:
* `LinearRing`: new members `mProjectedWidth` and `mProjectedHeight`; new method `computeDowngradePointList` that computes the list of points of a polyline that would be the projection of the GeoPoints on a centered size*size square; new method `resetPrecomputations`
* `PolyOverlayWithIW`: implemented the "fast draw of a list of segments" as less downgraded than a rectangle; new member `float[] mDowngradeSegments`; new method `isVisible(Paint)`; method `setDowngradePixelSizes` replaced `setDowngradeMaximumPixelSize`; minor refactoring
* `SampleShapeFile`: now using the new `PolyOverlayWithIW.setDowngradePixelSizes` method

### DIFF
--- a/OpenStreetMapViewer/src/main/java/org/osmdroid/samplefragments/data/SampleShapeFile.java
+++ b/OpenStreetMapViewer/src/main/java/org/osmdroid/samplefragments/data/SampleShapeFile.java
@@ -117,7 +117,7 @@ public class SampleShapeFile extends SampleMapEventListener {
                     for (final Overlay item : folder) {
                         if (item instanceof PolyOverlayWithIW) {
                             final PolyOverlayWithIW poly = (PolyOverlayWithIW)item;
-                            poly.setDowngradeMaximumPixelSize(20);
+                            poly.setDowngradePixelSizes(50, 25);
                             poly.setDowngradeDisplay(true);
                             final Paint paint = poly.getOutlinePaint();
                             paint.setStyle(Paint.Style.STROKE);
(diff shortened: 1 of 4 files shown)


monsieurtanuki:
@spyhunter99 Now I think that the polygon display is VERY fast, even in the case of multiple polygons such as in #1409.
For the 2035 ESRI polygons on a world map, the display now takes 80ms, to be compared with 3890ms the first time then 830ms ("Test 1" with no downgrade mode). Even without the "downgrade poly" optim, it's faster now (2290ms the first time then 720ms) because of other optimizations implemented _en route_ like not drawing a transparent `Paint` (for the fill) and optimizing the Path regarding consecutive segments on the same side (cf. `SideOptimizationPointAccepter`).

spyhunter99:
whoa awesome work!

monsieurtanuki:
@spyhunter99 Thank you! I'm currently working on #1458; I should PR today.
As many recent PR's involved Poly's and as we're talking about big improvements regarding performances and display (like @mat8854's #1378), maybe we have enough material to create a new version one day, something like 6.2.0. What do you think about that?

spyhunter99:
Let me know when you are ready

monsieurtanuki:
@spyhunter99 I'm ready. I see that there are still 2 issues with milestone 6.1.6:
* #1447 - we can survive if it's not done
* #1446 - more problematic to postpone it as we're talking about app crashes

PR #1487, opened 2020-01-20
bug/#1463_2 - getting rid of Fragments related memory leaks

monsieurtanuki:


monsieurtanuki:
Not all classes have been fixed, but now we know how to deal with possible fragment memory leaks.
And it's just about the demo, not the core library.

PR #1517, opened 2020-03-11
Should call mapTileRequestCompleted instead

treenewbee:
Should call mapTileRequestCompleted otherwise it will just keep expiring and requesting tiles.


treenewbee:
Makes sense, so MapTileApproximater is designed to return a temporary "better than nothing" scaled image, not to provide raster scaling capability for tiles out of native zoom levels. 

New issue here
https://github.com/osmdroid/osmdroid/issues/1519

Issue #1518, opened 2020-03-11
ShapeFile support problem.


sun-jiao:
## Issue Type

[x] Bug

add shapefile to map view failed.

even the sample file TM_WORLD_BORDERS_SIMPL-0.3 is not supported.

some borders in the file can't show in map.

(And my own shapefile totally dosen't work.)

### Screenshot

https://github.com/sun-jiao/GitHubUse/blob/master/Screenshot_20200311-140846.png

### Code

```
File myshape = new File (getExternalFilesDir("TM_WORLD_BORDERS_SIMPL-0.3") +"/TM_WORLD_BORDERS_SIMPL-0.3.shp");
        List<Overlay>  folder = null;
        try {
            folder = ShapeConverter.convert(mMapView, myshape);
        } catch (Exception ex) {
            ex.printStackTrace();
        }
        if (folder == null)
            Log.i("SHP", "onCreate: null");
        else{
            mMapView.getOverlayManager().addAll(folder);
            mMapView.invalidate();
        }
```

### known affected version: 

Android 10. 

### hardware

Google Pixel 3

### Version of osmdroid the issue relates to:

6.1.6

mat8854:
@sun-jiao For further debugging you should post your own shp-file here. Since it is a binary file format you can use a third party tool for extracting the data (coordinates):
* https://mapshaper.org/
* https://wiki.openstreetmap.org/wiki/Shapefiles

The osmdroid `ShapeConverter` is a simple parser mapping the shape data to osmdroid equivalents. It returns a `List<Overlay>`.

sun-jiao:
> @sun-jiao For further debugging you should post your own shp-file here. Since it is a binary file format you can use a third party tool for extracting the data (coordinates):
> 
> * https://mapshaper.org/
> * https://wiki.openstreetmap.org/wiki/Shapefiles
> 
> The osmdroid `ShapeConverter` is a simple parser mapping the shape data to osmdroid equivalents. It returns a `List<Overlay>`.

the sample file TM_WORLD_BORDERS_SIMPL-0.3 is not fully supported, too.

sun-jiao:
> @sun-jiao For further debugging you should post your own shp-file here. Since it is a binary file format you can use a third party tool for extracting the data (coordinates):
> 
> * https://mapshaper.org/
> * https://wiki.openstreetmap.org/wiki/Shapefiles
> 
[...]

PR #1556, opened 2020-04-17
fix out of range & color, line width, bubble customization for ShapeConverter

sun-jiao:


sun-jiao:
> @sun-jiao Could you please bring some context to your changes?
> If I sum up, you slighly tune the values of longitude and latitude, saying that if they are just a little beyond the conventional limits, they should be truncated to this limit.
> My questions:
> 
> 1. When do we get values only just beyond the limits?
> 2. Are we happy with values like latitude 92?
> 3. I can understand the problem with latitude; do we have to fix longitudes too? Couldn't we accept those values, or just add/substract 360 degrees?
> 
> Thank you for your answers.

Yeah. You're right.
I'll fix it.

sun-jiao:
> @sun-jiao Could you please bring some context to your changes?
> If I sum up, you slighly tune the values of longitude and latitude, saying that if they are just a little beyond the conventional limits, they should be truncated to this limit.
> My questions:
> 
> 1. When do we get values only just beyond the limits?
> 2. Are we happy with values like latitude 92?
> 3. I can understand the problem with latitude; do we have to fix longitudes too? Couldn't we accept those values, or just add/substract 360 degrees?
> 
> Thank you for your answers.

Background info is in this issue: [https://github.com/osmdroid/osmdroid/issues/1518#event-3243566215](ShapeFile support problem.)

> This shapefile has some Geo Points close to the 180° longitude, whose longitude is out of range. These numbers are slightly larger than 180.00 or smaller than -180.00, such as 180.00000000010004.
> Because of which, the BondingBox throws a illegal argument exception.
> I have discussed with the GIS developer, who created the file using ArcGis and ENVI. We think it's due to an error during the floating-point arithmetic of these software.
> Considering the errors that these software may cause, I think it's a good idea to add an inspection operation: if the longitude and latitude is slightly out of range, we can adjust it to a critical value in the ShapeConverter.java.

I think it's unnecessary to add/substract 360 degrees to values like 180.00000000010004.

monsieurtanuki:
@sun-jiao I can understand the double precision rounding issue.

A safe way to deal with latitudes would be to truncate them anyway to [-90,90], regardless of the 90.0000000001004 or 92 value cases.

Regarding longitude, I don't how relevant it is to add/substract 360 degrees in the specific case of Shape files:
* is it Shape file compliant to have longitude beyond 180 degrees?
* how are Shape files supposed to deal with a sequence of +179 then -179 longitudes: just 2 more degrees or around the planet?

Please feel free to amend this PR (I noticed that you closed both the #1518 issue and this un-merged PR).

sun-jiao:
> @sun-jiao I can understand the double precision rounding issue.
> 
> A safe way to deal with latitudes would be to truncate them anyway to [-90,90], regardless of the 90.0000000001004 or 92 value cases.
> 
> Regarding longitude, I don't how relevant it is to add/substract 360 degrees in the specific case of Shape files:
> 
> * is it Shape file compliant to have longitude beyond 180 degrees?
> * how are Shape files supposed to deal with a sequence of +179 then -179 longitudes: just 2 more degrees or around the planet?
> 
> Please feel free to amend this PR (I noticed that you closed both the #1518 issue and this un-merged PR).

Thanks, I'm working on it.

sun-jiao:
It's ok to add/substract 360 degrees, osmdroid can find the closest line between two points. 
For example, {(10.00, 170.00), (10.00, -170.00), (-10.00, -170.00), (-10.00, 170.00)} is a square rather than a very long rectangle.

And I added a lot of customization, which will be so useful.

I have tested in my own app.

osmdroid-shape/src/main/java/org/osmdroid/shape/ShapeConverter.java
```diff
@@ -39,7 +41,22 @@
...
+     *
+     * @param map the MapView to which these overlays will be added.
+     * @param file the shape file to be converted.
+     * @param prefs  allows the client to relax the level of validation when reading a shape file.
+     * @param fillColor fill color of polygon.
+     * @param outLineColor out line color of polyOverlay.
+     * @param width out line width of polyOverlay.
+     * @param getBubbleText customize titles, snippets, and sub-descriptions of bubbles.
+     *                      remember to handling bounding box if your shape file has both points and poly-overlays.
+     *                      please also mind that getSnippet is called before getTitle, and before getSubDescription.
```
monsieurtanuki, line 54: We shouldn't care about getSnippet being called before getTitle and getSubDescription: we can do that in `ShapeMetadataReader.read` at once in the right order.

osmdroid-shape/src/main/java/org/osmdroid/shape/ShapeConverter.java
```diff
@@ -39,7 +41,22 @@
...
+     * @return
+     * @throws Exception
+     */
+    public static List<Overlay> convert(MapView map, File file, ValidationPreferences prefs,
```
monsieurtanuki, line 58: I guess `int strokeWidth` and `int strokeColor` would make more sense.

osmdroid-shape/src/main/java/org/osmdroid/shape/ShapeConverter.java
```diff
@@ -174,16 +203,68 @@
         return folder;
     }
 
-    private static String getSensibleTitle(String snippet) {
-        if (snippet.length() > 100) {
-            return snippet.substring(0, 96) + "...";
+    private static GetBubbleText defaultBubbleText = new GetBubbleText() {
```
monsieurtanuki, line 206: Remove `defaultBubbleText` and put the relevant code in `ShapeMetadataReader`.

osmdroid-shape/src/main/java/org/osmdroid/shape/ShapeConverter.java
```diff
@@ -174,16 +203,68 @@
...
+        return convert(map, file, pref, fillColor, outLineColor, width, defaultBubbleText);
+    }
+
+    public static List<Overlay> convert(MapView map, File file, GetBubbleText getBubbleText) throws Exception {
+        ValidationPreferences pref = new ValidationPreferences();
+        pref.setMaxNumberOfPointsPerShape(200000);
+        return convert(map, file, pref, Color.TRANSPARENT, Color.BLACK, 5, getBubbleText);
+    }
+
+    public static List<Overlay> convert(MapView map, File file, int fillColor, int outLineColor, int width, GetBubbleText getBubbleText) throws Exception {
```
monsieurtanuki, line 249: We don't need that: remove.

osmdroid-shape/src/main/java/org/osmdroid/shape/ShapeConverter.java
```diff
@@ -174,16 +203,68 @@
...
         ValidationPreferences pref = new ValidationPreferences();
         pref.setMaxNumberOfPointsPerShape(200000);
         return convert(map, file, pref);
     }
+
+    public static List<Overlay> convert(MapView map, File file, ValidationPreferences pref) throws Exception {
+        return convert(map, file, pref, Color.TRANSPARENT, Color.BLACK, 5, defaultBubbleText);
+    }
+
+    public static List<Overlay> convert(MapView map, File file, int fillColor, int outLineColor, int width) throws Exception {
```
monsieurtanuki, line 237: Remove this version, we don't need it.

osmdroid-shape/src/main/java/org/osmdroid/shape/ShapeConverter.java
```diff
@@ -174,16 +203,68 @@
...
+        }
+    };
 
     public static List<Overlay>  convert(MapView map, File file) throws Exception {
         ValidationPreferences pref = new ValidationPreferences();
         pref.setMaxNumberOfPointsPerShape(200000);
         return convert(map, file, pref);
     }
+
+    public static List<Overlay> convert(MapView map, File file, ValidationPreferences pref) throws Exception {
```
monsieurtanuki, line 233: Legacy version: OK to keep it.

osmdroid-shape/src/main/java/org/osmdroid/shape/ShapeConverter.java
```diff
@@ -174,16 +203,68 @@
...
-        return snippet;
-    }
+
+        @Override
+        public String getSnippet(String metadata, String boundingBox) {
+            return metadata;
+        }
+    };
 
     public static List<Overlay>  convert(MapView map, File file) throws Exception {
```
monsieurtanuki, line 227: Legacy version: OK to keep it, but use `getDefaultValidationPreferences()`.

sun-jiao:
> @sun-jiao Confusion with all those unrelated changes. I guess they make sense for a better customization, though.
> 
> Please have a look at my remarks; if you're OK with them, implement them and it's perfect.
> If you don't agree or have some questions, don't hesitate to reply.

Resolved. Thanks for your suggestions.

osmdroid-shape/src/main/java/org/osmdroid/shape/ShapeConverter.java
monsieurtanuki: Thank you for the fix.

Now that I'm thinking it would make much more sense if instead of the fill color, stroke color and stroke width we used 2 nullable parameters instead : `Paint fillPaint` and `Paint strokePaint`. And use them in the code with something like
````java
if (fillPaint != null) {
    polygon.getFillPaint().set(fillPaint);
}
````


It's better because we have much more control over the polygon paint (rounding the segments ending, dashed strokes, ...). And we don't have to create another version of `convert` if we say "now I'd like to add a style to all the shape polygons".

File: osmdroid-shape/src/main/java/org/osmdroid/shape/ShapeMetadataReader.java [hunk 8]
```diff
@@ -0,0 +1,26 @@
+package org.osmdroid.shape;
+
+import net.iryndin.jdbf.core.DbfRecord;
+
+import org.osmdroid.util.BoundingBox;
+
+import java.nio.charset.Charset;
+import java.text.ParseException;
+
+public abstract class ShapeMetadataReader {
```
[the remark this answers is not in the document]
sun-jiao: I think methods `setBoundingBox` and `getSubDescription` are useful.
Developers might wanna change sub-descriptions if the auto generated ones have wrong values due to the double precision rounding issue. They may also want to change the decimal places.

For example, our project:

20200418_160752-COLLAGE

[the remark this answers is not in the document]
monsieurtanuki: OK, I see the use case.
But then, it's like the `int fillColor` vs. `Paint fillPaint` competition: why do we use `BoundingBox` but not the number of points? Or other polygon details? Or why is do we use subDescription/BoundingBox for Polygons, but not for Polylines?

I think I've got the solution: instead of a `ShapeMetadataReader` class/interface, we should use something like that:
````java
public interface ShapeMetaSetter{
    void set(DbfRecord metadata, Polygon polygon);
    void set(DbfRecord metadata, Polyline polyline);
    void set(DbfRecord metadata, Marker marker);
}
````
and implement it with a `DefaultShapeMetaSetter` class, where you can do what you want with the metadata AND the bounding box or the number of points, whatever, for the title, the snippet and the sub description.

Do you get it so far?
Because I have an even better idea: why couldn't we also set the paint here?
What's the point? Well, maybe you are able to see in the metadata that this polygon is important (and paint it in red instead of black), or that this polygon's bounding box is tiny, therefore you need to give it a slightly bigger stroke width. Or you want the polygons and the polylines to have a different color, which is not possible for the moment.

That would be brillant, wouldn't it?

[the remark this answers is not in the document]
Comment (sun-jiao) [reply osmdroid-shape/src/main/java/org/osmdroid/shape/ShapeMetadataReader.java:10]: > OK, I see the use case.
> But then, it's like the `int fillColor` vs. `Paint fillPaint` competition: why do we use `BoundingBox` but not the number of points? Or other polygon details? Or why is do we use subDescription/BoundingBox for Polygons, but not for Polylines?
> 
> I think I've got the solution: instead of a `ShapeMetadataReader` class/interface, we should use something like that:
> 
> ```java
> public interface ShapeMetaSetter{
>     void set(DbfRecord metadata, Polygon polygon);
>     void set(DbfRecord metadata, Polyline polyline);
>     void set(DbfRecord metadata, Marker marker);
> }
> ```
> 
> and implement it with a `DefaultShapeMetaSetter` class, where you can do what you want with the metadata AND the bounding box or the number of points, whatever, for the title, the snippet and the sub description.
> 
> Do you get it so far?
> Because I have an even better idea: why couldn't we also set the paint here?
> What's the point? Well, maybe you are able to see in the metadata that this polygon is important (and paint it in red instead of black), or that this polygon's bounding box is tiny, therefore you need to give it a slightly bigger stroke width. Or you want the polygons and the polylines to have a different color, which is not possible for the moment.
> 
> That would be brillant, wouldn't it?

Good idea. That's right.

File: osmdroid-shape/src/main/java/org/osmdroid/shape/ShapeConverter.java [hunk 9]
```diff
@@ -66,13 +84,12 @@
...
                         PointShape aPoint = (PointShape) s;
                         Marker m = new Marker(map);
-                        m.setPosition(new GeoPoint(aPoint.getY(), aPoint.getX()));
+                        m.setPosition(fixOutOfRange(new GeoPoint(aPoint.getY(), aPoint.getX())));
                         if (metadata != null) {
-                            metadata.setStringCharset(Charset.defaultCharset());
-                            m.setSnippet(metadata.toMap().toString());
-                            m.setTitle(getSensibleTitle(m.getSnippet()));
-
+                            shapeMetadataReader.read(metadata);
```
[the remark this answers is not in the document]
sun-jiao: I don't think the first solution is a good idea.
Developers may want to use overlays to set bubble titles.

osmdroid-shape/src/main/java/org/osmdroid/shape/ShapeMetaSetter.java
```diff
@@ -0,0 +1,18 @@
...
+import android.graphics.Paint;
+
+import net.iryndin.jdbf.core.DbfRecord;
+
+import org.osmdroid.views.overlay.Overlay;
+
+import java.text.ParseException;
+
+public interface ShapeMetaSetter {
+    void set(DbfRecord metadata, Overlay overlay) throws ParseException;
```
monsieurtanuki, line 12: Just this, and that's enough:
````java
void set(DbfRecord metadata, Polygon polygon) throws ParseException;
void set(DbfRecord metadata, Polyline polyline) throws ParseException;
void set(DbfRecord metadata, Marker marker) throws ParseException;
````

File: osmdroid-shape/src/main/java/org/osmdroid/shape/ShapeConverter.java [hunk 11]
```diff
@@ -83,74 +91,74 @@
...
-                            m.setPosition(new GeoPoint(p.getY(), p.getX()));
-                            if (metadata != null) {
-                                metadata.setStringCharset(Charset.defaultCharset());
-                                m.setSnippet(metadata.toMap().toString());
-                                m.setTitle(getSensibleTitle(m.getSnippet()));
+                            m.setPosition(fixOutOfRange(new GeoPoint(p.getY(), p.getX())));
 
-
-                            }
+                            shapeMetaSetter.set(metadata, m);
```
monsieurtanuki, line 155: Just `shapeMetaSetter.set(metadata, m)`: the other two lines `m.setSnippet(shapeMetaSetter.getSnippet())` and `m.setTitle(shapeMetaSetter.getTitle())` will be performed inside `shapeMetaSetter.set(metadata, m)`.

osmdroid-shape/src/main/java/org/osmdroid/shape/ShapeConverter.java
```diff
@@ -83,74 +91,74 @@
...
                             for (PointData p : points) {
-                                GeoPoint pt = new GeoPoint(p.getY(), p.getX());
+                                GeoPoint pt = fixOutOfRange(new GeoPoint(p.getY(), p.getX()));
                                 pts.add(pt);
                             }
-                            line.setPoints(pts);
+
+                            line.setPoints(pts);//points out of range should be fixed before this line.
+
+                            shapeMetaSetter.set(metadata, line);
```
monsieurtanuki, line 135: Just `shapeMetaSetter.set(metadata, line)`: the other three lines `line.setSnippet(shapeMetaSetter.getSnippet())`, `line.setTitle(shapeMetaSetter.getTitle())` and `line.setSubDescription(shapeMetaSetter.getSubDescription())` will be performed inside `shapeMetaSetter.set(metadata, line)`.

osmdroid-shape/src/main/java/org/osmdroid/shape/ShapeConverter.java
```diff
@@ -83,74 +91,74 @@
...
                             }
                             pts.add(pts.get(0));    //force the polygon to close
 
-                            polygon.setPoints(pts);
-                            final BoundingBox boundingBox = polygon.getBounds();
-                            polygon.setSubDescription(boundingBox.toString());
-                            folder.add(polygon);
+                            polygon.setPoints(pts); //points out of range should be fixed before this line.
+
+                            shapeMetaSetter.set(metadata, polygon);
```
monsieurtanuki, line 107: Just `shapeMetaSetter.set(metadata, polygon)`: the other three lines `polygon.setSnippet(shapeMetaSetter.getSnippet())`, `polygon.setTitle(shapeMetaSetter.getTitle())` and `polygon.setSubDescription(shapeMetaSetter.getSubDescription())` plus the fillPaint and outlinePaint will be performed inside `shapeMetaSetter.set(metadata, polygon)`.

osmdroid-shape/src/main/java/org/osmdroid/shape/DefaultShapeMetaSetter.java
```diff
@@ -0,0 +1,74 @@
...
+    @Override
+    public void set(DbfRecord metadata, Overlay overlay) throws ParseException {
+        if (metadata != null){
+            metadata.setStringCharset(Charset.defaultCharset());
+            metaString = metadata.toMap().toString();
+        } else{
+            metaString = "";
+        }
+
+        shapeType = overlay.getClass().getName();
```
monsieurtanuki, line 31: `shapeType = overlay.getClass().getName()` is very ugly; this is not OOP.
Just implement
* `public void set(DbfRecord metadata, Polygon polygon) throws ParseException`
* `public void set(DbfRecord metadata, Polyline polyline) throws ParseException`
* `public void set(DbfRecord metadata, Marker marker) throws ParseException`
In those methods you compute whatever is relevant for each of the 3 possible Overlay classes, and you set it directly.

osmdroid-shape/src/main/java/org/osmdroid/shape/DefaultShapeMetaSetter.java
```diff
@@ -0,0 +1,74 @@
...
+            subString = ((Marker) overlay).getPosition().toString();
+        else if (shapeType.equals("org.osmdroid.views.overlay.Polygon") || shapeType.equals("org.osmdroid.views.overlay.Polyline"))
+            subString = overlay.getBounds().toString();
+        else
+            subString = "";
+
+        mOverlay = overlay;
+    }
+
+    @Override
```
monsieurtanuki, line 43: we don't need that.

osmdroid-shape/src/main/java/org/osmdroid/shape/DefaultShapeMetaSetter.java
```diff
@@ -0,0 +1,74 @@
...
+    public String getSnippet() {
+        return metaString;
+    }
+
+    @Override
```
monsieurtanuki, line 48: we don't need that.

osmdroid-shape/src/main/java/org/osmdroid/shape/DefaultShapeMetaSetter.java
```diff
@@ -0,0 +1,74 @@
...
+    public String getTitle() {
+        if (metaString.length() > 100) {
+            return metaString.substring(0, 96) + "...";
+        }
+        return metaString;
+    }
+
+    @Override
```
monsieurtanuki, line 56: we don't need that.

osmdroid-shape/src/main/java/org/osmdroid/shape/DefaultShapeMetaSetter.java
```diff
@@ -0,0 +1,74 @@
...
+    public String getSubDescription() {
+        return subString;
+    }
+
+    @Override
```
monsieurtanuki, line 61: we don't need that.

osmdroid-shape/src/main/java/org/osmdroid/shape/DefaultShapeMetaSetter.java
```diff
@@ -0,0 +1,74 @@
...
+    public Paint getFillPaint() {
+        if (shapeType.equals("org.osmdroid.views.overlay.Polygon"))
+            return ((Polygon) mOverlay).getFillPaint();
+        return null;
+    }
+
+    @Override
```
monsieurtanuki, line 68: we don't need that.

monsieurtanuki: we don't need that.

Oh, I understand the confusion: the point is not to retrieve the Overlay fill paint or stroke paint, the point is to set it.
Basically: standard case, I don't really care about the colors, I'm OK with the default values of Polygon and Polyline painting. It's cool: I do nothing, I don't override the defaults.
Refined case: I want the polygons to have a red stroke color and the polylines a blue one. For Polygons it's going to be written `polygon.getOutlinePaint().setColor(Color.RED)`. And something similar for polylines.

The point is: the code inside `ShapeConverter` doesn't care at all about what you want to change for your polygons, color, title, snippet, ..., because the `ShapeMetaSetter` will do everything about it (first step: computing the values, second step: setting the values for the polygon).

osmdroid-shape/src/main/java/org/osmdroid/shape/ShapeConverter.java
```diff
@@ -66,13 +76,11 @@
...
                         Marker m = new Marker(map);
-                        m.setPosition(new GeoPoint(aPoint.getY(), aPoint.getX()));
-                        if (metadata != null) {
-                            metadata.setStringCharset(Charset.defaultCharset());
-                            m.setSnippet(metadata.toMap().toString());
-                            m.setTitle(getSensibleTitle(m.getSnippet()));
+                        m.setPosition(fixOutOfRange(new GeoPoint(aPoint.getY(), aPoint.getX())));
 
-                        }
+                        shapeMetaSetter.set(metadata, m);
```
monsieurtanuki, line 81: Just `shapeMetaSetter.set(metadata, m)`: the other two lines `m.setSnippet(shapeMetaSetter.getSnippet())` and `m.setTitle(shapeMetaSetter.getTitle())` will be performed inside `shapeMetaSetter.set(metadata, m)`.

PR #1563, opened 2020-04-24
feature/#1551 - new class on top of ColorMappingForScalar for scalar range refresh

monsieurtanuki:
New class:
* `ColorMappingForScalarContainer`: Scalar container on top of any ColorMappingForScalar

Impacted classes:
* `ColorMappingForScalar`: created a `clear` method
* `ColorMappingVariation`: fixed a computing bug on method `mapScalar`; made method `init` public and fixed a side-effect when computing the slope
* `IConfigurationProvider`: unrelated micro-refactoring


spyhunter99:
sorry out of the loop, is this for doing fancy color fading effects on lines?

monsieurtanuki:
@spyhunter99 The fancy color effect is already there: this time, we make it more dynamic.
Typical use case: initially you don't know the altitude range of your hike so you put default values, but as you walk and as the GPS data arrive, you get a more relevant idea of the range.
Now it's easier to refresh the colors from the actual data (like min/max), on the fly.

monsieurtanuki:
> Could be useful to provide a sample / documentation as extra sample in the sample application. I would do that on a different PR.

Perfect!

Issue #1512, opened 2020-03-06
Android 10 map doesn't load


F3FFO:
# Issue Type

[ ] Question
[X] Bug
[X] Improvement
[ ] Build system related
[ ] Performance
[ ] Documentation

## Description and/or steps/code to reproduce the problem

Map doesn't load in android 10, but only in android 9.0 and previous version

### If it's a bug, version(s) of android this affects:

Android 10.0

### Version of osmdroid the issue relates to:

6.1.5

spyhunter99:
i just updated to 10, sample app works fine with gps over here.

F3FFO:
Ah ok ok, i have to download the source and not to implement in gradle is it true?

Conan42:
Hello, I confirm Osmdroid 6.1.5 does not load maps on Android 10. Why?

Conan42:
I just put osmdroid 6.1.6, it has refreshed gradle and its walk thank you spyhunter99

cbalster:
Without further investigating: One possible reason could be the scoped storage changes in API29. The code to determine cache locations seems a little outdated.
https://developer.android.com/about/versions/10/privacy/changes#scoped-storage
That said the sample app works just fine on an API29 emulator, so this is definitely something device specific.

mat-sienczyk:
Version 6.1.6 also does not work on Andorid 10 - app is targeting API 29 (some SQL exceptions). On Android 8 is fine.

cbalster:
I found the Issue.
If you're targeting API29, then the path returned by Environment.getExternalStorageDirectory() is no longer accessible (see [here](https://developer.android.com/reference/android/os/Environment.html#getExternalStorageDirectory)). Same is true for similar methods.
Since the default way of finding a cache directory as implemented in `DefaultConfigurationProvider` does not have access to a `Context`, there is currently no way to get a cache directory on API29 automatically.
The short-term solution is to explicitly set the cache directory before initializing your `MapView`. An Example might look like this:
```
IConfigurationProvider configuration = Configuration.getInstance();
File path = context.getFilesDir();
File osmdroidBasePath = new File(path, "osmdroid");
osmdroidBasePath.mkdirs();
File osmdroidTilePath = new File(osmdroidBasePath, "tiles");
noinspection ResultOfMethodCallIgnored
osmdroidTilePath.mkdirs();
configuration.setOsmdroidBasePath(osmdroidBasePath);
[...]

PR #1571, opened 2020-04-28
Bug/#1512 - Storage location on API29 and above

cbalster:
#1512 
Basically a complete overhaul of StorageUtils:

- lot's of refactoring
- < API19: Behaviour is untouched except fixing a minor bug, that lead to non-writable locations returned as writable
- 19 <= API < 29: legacy behaviour + additional storage locations when a context is provided
- \>API29: Only new behviour with fallback for targetSdk < API29 apps on >API29 devices.
- replacing references to now deprecated methods to the new alternatives.

On API29 and above a valid storage location can only be retrieved when a context is available. Therefore a hacky workaround has been implemented in MapView. This should cover most naive usages of the library.
In more advanced cases (e.g. own `MapTileModuleProviderBase` implementations) this can still lead to problems if the storage detection via `getOsmdroidBasePath(Context)` / `getOsmdroidTileCache(Context)` was not triggered before with a context or a valid directory was set via `setOsmdroidBasePath()` / `setOsmdroidTileCache()`.
This is not easily fixed due to the Configuration being a static singleton. This should probably be changed in a future release since it also causes lots of other issues (testing, hard to reason about code), but will mean a breaking change.

### DIFF
--- a/OpenStreetMapViewer/src/main/java/org/osmdroid/samplefragments/geopackage/GeopackageFeatureTiles.java
+++ b/OpenStreetMapViewer/src/main/java/org/osmdroid/samplefragments/geopackage/GeopackageFeatureTiles.java
@@ -245,9 +245,9 @@ public class GeopackageFeatureTiles extends BaseSampleFragment {
      *
      * @return
      */
-    protected static Set<File> findMapFiles() {
+    protected Set<File> findMapFiles() {
         Set<File> maps = new HashSet<>();
-        List<StorageUtils.StorageInfo> storageList = StorageUtils.getStorageList();
+        List<StorageUtils.StorageInfo> storageList = StorageUtils.getStorageList(getActivity());
         for (int i = 0; i < storageList.size(); i++) {
             File f = new File(storageList.get(i).path + File.separator + "osmdroid" + File.separator);
             if (f.exists()) {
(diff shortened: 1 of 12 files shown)


osmdroid-android/src/main/java/org/osmdroid/views/MapView.java
```diff
@@ -205,6 +205,12 @@ public MapView(final Context context,
 					  MapTileProviderBase tileProvider,
 					  final Handler tileRequestCompleteHandler, final AttributeSet attrs, boolean hardwareAccelerated) {
 		super(context, attrs);
+
+		// Hacky workaround: If no storage location was set manually, we need to try to be
```
monsieurtanuki, line 209: The workaround makes sense.

OpenStreetMapViewer/src/main/java/org/osmdroid/samplefragments/geopackage/GeopackageFeatureTiles.java
```diff
@@ -245,9 +245,9 @@ private void updateInfo() {
      *
      * @return
      */
-    protected static Set<File> findMapFiles() {
+    protected Set<File> findMapFiles() {
```
monsieurtanuki, line 248: I know we're talking only about samples, but that's not that great to see the same code in `GeopackageFeatureTiles`, `GeopackageFeatures`, `GeopackageSample` and `MapsforgeTileProviderSample`.

What about implementing `public static Set<File> findMapFiles(Activity activity, String endsWith)`? Somewhere in the sample app (perhaps `BaseFragment` - and we would even get rid of `static` and the `Activity` parameter).

cbalster: I actually started a cleanup off the sample app (formatting, lint errors etc.). After that I wanted to take a look at things like this. 👍 
But that's for another PR I think...

PR #1607, opened 2020-06-27
fix lint errors

hannesa2:


osmdroid-android/src/main/AndroidManifest.xml
```diff
@@ -8,5 +8,8 @@
 	<uses-feature android:name="android.hardware.telephony" android:required="false" />
 	<uses-feature android:name="android.hardware.wifi" android:required="false" />
 
+    <uses-permission android:name="android.permission.ACCESS_COARSE_LOCATION" />
```
spyhunter99, line 11: i thought we all agreed that not everyone needs this. While is convenient for some developers, it may add too many unneeded permissions in specific use cases

osmdroid-android/src/main/java/org/osmdroid/LocationListenerProxy.java
```diff
@@ -6,58 +6,56 @@
...
+            mListener.onProviderEnabled(arg0);
+        }
+    }
+
+    @Override
+    public void onStatusChanged(final String arg0, final int arg1, final Bundle arg2) {
+        if (mListener != null) {
+            mListener.onStatusChanged(arg0, arg1, arg2);
+        }
+    }
```
spyhunter99, line 60: is this change just a reformat or was there any substantive changes?

osmdroid-android/src/main/java/org/osmdroid/tileprovider/MapTileProviderBasic.java
```diff
@@ -19,204 +18,199 @@
...
+        if (approximationIndex < downloaderIndex && pOfflineFirst) {
+            return true;
+        }
+        if (approximationIndex > downloaderIndex && !pOfflineFirst) {
+            return true;
+        }
+        mTileProviderList.set(downloaderIndex, mApproximationProvider);
+        mTileProviderList.set(approximationIndex, mDownloaderProvider);
+        return true;
+    }
```
spyhunter99, line 215: same as previous comment

osmdroid-android/src/main/java/org/osmdroid/LocationListenerProxy.java
hannesa2: format

hannesa2: This repo would need a reformat of all code, but when I do it, I will touch almost every file 
.... on the level you trust, it's impossible for me to do this step 😄 

This is a pre-task to format all https://github.com/osmdroid/osmdroid/pull/1620

osmdroid-android/src/main/AndroidManifest.xml
hannesa2: I removed the permission to have some progress. But as follow up, 

* lint will show an issue 👎 
* a possible error source is now re-introduced 👎 
* most of the developer have to add this permission manually in there own repo 👎 , instead of a single place here

I would do it vice-versa and go with adding the permission and who don't want this can remove it

`<uses-permission android:name="android.permission.ACCESS_COARSE_LOCATION"" tools:node="remove"/>`

## Benefit:
It will work out of the box, nobody is force to do something and it removes a potential error source

PR #1618, opened 2020-07-21
Set TLS1.2 as the only protocol if supported

### LINKED ISSUE #1617: Android 4.x.x not downloading Titles
## Issue Type
[x] Bug

## Description and/or steps/code to reproduce the problem
Titles are set to MAPNIK.
When trying to render the map I get the following error 

```java.net.ConnectException: failed to connect to b.tile.openstreetmap.org/2a01:4f8:1c1c:bc54::1 (port 443): connect failed: ENETUNREACH (Network is unreachable)```

According to this [https://stackoverflow.com/questions/29916962/javax-net-ssl-sslhandshakeexception-javax-net-ssl-sslprotocolexception-ssl-han/50640113#50640113](https://stackoverflow.com/questions/29916962/javax-net-ssl-sslhandshakeexception-javax-net-ssl-sslprotocolexception-ssl-han/50640113#50640113) is a problem with TSL and Android 4.x.x but I'm not so sure if that's the answer.

## Environment
Device with 4.2 and any emulator using 4.x.x

### If it's a bug, version(s) of android this affects:
4.x.x

6.1.6

cbalster:
Keep old behaviour as fallback
fixes #1617

### DIFF
--- a/osmdroid-android/src/main/java/org/osmdroid/tileprovider/modules/TileDownloader.java
+++ b/osmdroid-android/src/main/java/org/osmdroid/tileprovider/modules/TileDownloader.java
@@ -344,29 +344,25 @@ public class TileDownloader {
             String[] enabledProtocols = socket.getEnabledProtocols();
             String[] newEnabledProtocols;
 
-            int sslEnabled = Arrays.binarySearch(enabledProtocols, "SSLv3");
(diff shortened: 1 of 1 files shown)


cbalster:
Just as a reminder: The whole protocol fix logic is only used on API<=20

PR #1622, opened 2020-07-26
Only set CompatiblitySocketFactory once

cbalster:
Fix Stackoverflow exceptions due to number of sockets created
@spyhunter99 This should be considered a hotfix as it fixes a critical bug on API <= 19 with encrypted connections.


spyhunter99:
i'll get a release out asap

PR #1635, opened 2020-08-31
Fix the map fling bounce from bug/#1612

MatJAchen:
What seemed to be the problem: 
1. The velocity for the fling needed to be rotated, but using Projection.unrotateAndScalePoint() would cause it to be rotated about the currently scrolled to center. The velocity needs to be rotated about (0, 0), rotating it about an offset point can cause an incorrect velocity.
2. The VelocityTracker inside GestureDetector seems to get confused when supplied with rotated touch events and return an incorrect velocity on occasion. To stop this, pass the touch events forward as normal and rotate them later when they will be used.

Tested on a device running Android 9. Fling, scroll, double tap, and single tap operations all worked as expected.

### DIFF
--- a/osmdroid-android/src/main/java/org/osmdroid/views/MapView.java
+++ b/osmdroid-android/src/main/java/org/osmdroid/views/MapView.java
@@ -54,7 +54,7 @@ import android.widget.ZoomButtonsController;
  * This is the primary view for osmdroid. <br><br>
  * As of version 6.0.0, please respect the android view lifecycle by calling
  * {@link MapView#onPause()} and {@link MapView#onResume()} respectively
- * 
+ *
  * @since the begining
  * @author plusminus on 17:45:56 - 25.09.2008
  * @author and many other contributors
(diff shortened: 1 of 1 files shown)


monsieurtanuki:
@InI4 What do you think of that?

InI4:
@MatJAchen, @monsieurtanuki: This is a testing problem: I have no idea to evaluate this. 

I can see, the code is now careful about creating rotated evens. OK.

My only in hardware phone moved from Android 9 to Android 10. Looking at the patches I sent, my latest patch essentially reverted my previous one. I.e. it looked like on my device (OnePlus6) Versions Android 8 and 10 worked one way and the Android 9 behaved differently in terms of these gesture directions. I also have a puzzle of reports of users of my app, that things work on Android 9, 10 or don't. It is hard to get user feedback though.

All I can say is by now, my patches work perfectly on my phone. Yes, I could try the new one, but I doubt that (yea, mine) >= 28 cases now.

How can we verify doing the right thing for the most devices? (I think it should run on Android 10 and 9 and 8, but from my user reports I am not even sure if all pieces of Android 9 hardware work the same).

monsieurtanuki:
If I remember well, there are only 2 cases: both work, but on distinct devices.

What about
1. implementing the most popular case by default
1. creating something like `MapView.setSpecificEventRotation(boolean)`
1. letting the developer add a "troubleshooting settings" section in their app, where the end-user can select the other non default behavior

Other idea: is there a way we could emulate programmatically the gesture and see what the result is? With that, we can run an "init" check and decide which solution is the correct one for the rest of the app.

We won't maintain a list of devices anyway, will we?

InI4:
A list of devices is bad, maximally a list of devices know to fail.

PR #1642, opened 2020-09-19
Apply screen rotation before calculating the bounding box

xyz32:
The bounding box will now accommodate for rectangular screens being rotated throughout 360 degrees.


xyz32:
BBOX before the fix with a 90 map rotation:
http://bboxfinder.com/#38.576770,-77.616325,49.534869,-68.651321

BBOX after the fix with a 90 map rotation:
http://bboxfinder.com/#41.000182,-80.812184,47.450336,-65.463763

monsieurtanuki (approved):
@xyz32 Thank you for the bug fix!

Issue #1657, opened 2020-11-09
Gray tile when zooming out


Saursinet:
## Issue Type

[X] Bug

## Description and/or steps/code to reproduce the problem

When user zoom out and that new tiles appear some of them take default color indicated by 
```
tilesOverlay.loadingBackgroundColor = context.getColor(R.color.loading_tile)
tilesOverlay.loadingLineColor = context.getColor(R.color.loading_tile)
```
But I have some others which appear gray and I cannot find a way to fix this. (cf: photo)
![Screenshot_20201029-155156](https://user-images.githubusercontent.com/10145464/98557062-06438000-22a4-11eb-984b-0d2258be5678.jpg)
On this photo we can see 2 colors. The yellowish is the default tile color while the gray is not set anywhere. 
Also after loading the tiles appear correctly but it's just in the meantime that we see this Gray color.

### If it's a bug, version(s) of android this affects:

All

### Version of osmdroid the issue relates to:

6.1.6

monsieurtanuki:
@Saursinet I guess it's when the tile is not available for the moment and is being "approximated" from relevant tiles on other zoom levels, cf. [MapTileProviderBase](https://github.com/osmdroid/osmdroid/blob/master/osmdroid-android/src/main/java/org/osmdroid/tileprovider/MapTileProviderBase.java#L473)
Does that make sense? Would you like to have control on the "approximation" background color?

Saursinet:
Ah indeed! It does make sense! And yes I would like to be able to set this "approximation" background color so that it would almost match empty land color!

monsieurtanuki:
Well, I guess a `static` setter would do the trick.
I've coded that, and then I got trapped into a git issue that could easily turn into a time-consuming nightmare. That would last much more than the 5 minutes I was able to allocate to this issue.
Feel free to send a PR.

Saursinet:
Thank you for checking and look into it!

monsieurtanuki:
@Saursinet Strange behavior...
If I sum up, if you chose color A, color B will be displayed wherever you expected color A as the default approximation color. Your only problem is that this is not the color you chose - but at least there's consistency as the same color B is displayed every time you expected color A. Is that a good description?

Saursinet:
So yes the summary is good there is only one nuance in that description. We seem to have the colour deeper in some places (cf: photo). This is the app with the color #EEF1E0 used [
[...]

PR #1687, opened 2021-01-26
Feature/#1657

Saursinet:
As suggested by @monsieurtanuki in the issue #1657, I added a static value and a setter to be able to change the color of the approximative tile.


monsieurtanuki (approved):
@Saursinet Thank you!

I can't remember why I initially suggested a `static` method, which can be problematic unless you frequently set again the color (at least each time you create your activity). It might be because all map providers should use the same value - makes sense.

Well, good enough is good enough!

PR #1705, opened 2021-03-19
Modernise the build to support recent Gradle, Android Gradle Plugin, and JDKs

alsutton:
I'm happy to bounce around some ideas on this, but wanted to get some eyeballs on it before I spend too much time on it.

When I checked out the project I hit some issues building because my default JDK is the latest LTS (jdk 11) which isn't compatible with the current osmdroid build setup. Updating the gradle wrapper to a version which supports JDK 11 sent me off on a bit of a yak-shaving expedition which has left me with;

- Gradle updated from 2.14.1 to 6.8.3 (latest stable)
- Android Gradle Plugin updated from 2.2.0 to 4.1.3 (latest stable).
- Removal of gradle-fury which is incompatible with the latest gradle and AGP and [appears to be unmaintained](https://github.com/gradle-fury/gradle-fury)
- Replacing findbugs with spotbugs, and removing jdepend, because the findbugs and jdepend plugins were deprecated in Gradle 5.0 [due to not supporting Java 9+ bytecode](https://github.com/gradle/gradle/issues/6652)
- Updating the build files to use `doLast` instead of `<<` because the left-shift notation was [removed in Gradle 5.0](https://github.com/gradle/gradle/issues/6266).
- Updating the travis configuration to use the build tools which are used by the updated Android Gradle Plugin.


alsutton:
I'm slowly reading through https://github.com/osmdroid/osmdroid/pull/1594 . The TLDR when compared to this is, so far as I can see, maven publishing needs to be added. Let me know if I've missed anything.

hannesa2:
In my point of view Travis is complete outdated. Github actions are a better way, especial for missing test-runs on CI

OpenStreetMapViewer/build.gradle
```diff
@@ -1,33 +1,10 @@
...
-                    || (task.name.equals("validateSigningRelease"))) {
-
-                task.setEnabled(canSign)
-            }
-        }
-    }
-}
 
 android {
+    compileSdkVersion findProperty('android.compileSdkVersion').toInteger()
```
hannesa2, line 6: There are much easier ways to to that

gradle.properties
```diff
@@ -40,7 +40,7 @@ org.gradle.jvmargs=-Xms256m -Xmx2048m -XX:MaxPermSize=256m
 gradleFury.version=1.0.14
 
 
-android-plugin.version=2.2.0
+android-plugin.version=4.1.3
```
hannesa2, line 43: I would complete remove the variables here. An probably upcoming dependabot job, will make this obsolete anyways

gradlew
```diff
@@ -6,12 +6,30 @@
 ##
 ##############################################################################
 
-# Add default JVM options here. You can also use JAVA_OPTS and GRADLE_OPTS to pass JVM options to this script.
-DEFAULT_JVM_OPTS=""
+# Attempt to set APP_HOME
```
hannesa2, line 9: Without a gradle verify job I would never trust this. e.g https://github.com/hannesa2/LiveEdgeDetection/blob/master/.github/workflows/gradle-wrapper-validation.yml

OpenStreetMapViewer/build.gradle
alsutton: Such as?

gradle.properties
alsutton: I'd like to keep this change focused and small, so a future PR could remove the variable.

gradlew
alsutton: This has come from the upstream version of Gradle. What threat model do you see from keeping this as is?

alsutton:
In this PR I'd like to focus on modernising the build and avoid conflating the change with alterations to CI or other systems (and I'd like to avoid a CI war :)).

OpenStreetMapViewer/build.gradle
hannesa2:
```
buildscript {
    ext {
        sdkCompileVersion = 29
        ....
```
```
android {
    compileSdkVersion sdkCompileVersion
```

No need having `gradle.properties`

gradlew
hannesa2: A blueprint as a link is too easy ?

hannesa2:
> In this PR I'd like to focus on modernising the build and avoid conflating the change with alterations to CI or other systems (and I'd like to avoid a CI war :)).

Good luck.
I would love to help/annoy you more, but I'll stop wasting my time here

OpenStreetMapViewer/build.gradle
alsutton: That doesn't seem to be a core requirement to updating the build tools used, so it's probably better left for a different PR.

gradlew
alsutton: Your statement doesn't seem to answer the question of what problem you're trying to solve with that change. To me there doesn't seem to be need to make your suggested change because it's not a requirement to achieve the goal of this PR.

alsutton:
All feedback is welcome, if you feel that your time is best spent elsewhere then thank you for your comments and cheerio 👋

PR #1715, opened 2021-04-18
Feature/create archive provider

pasniak:
Hi, I would like to be able to create a custom archive provider (in my own MapTileProvider inheriting from MapTileProviderBasic).

The PR has change which moves constructor to init() and adds `createApproximater()` `createAssetsProvider()` `createArchiveProvider()`  which can be overwritten in client code.


osmdroid-android/src/main/java/org/osmdroid/tileprovider/cachemanager/CacheManager.java
```diff
@@ -838,7 +838,7 @@ public CacheManagerAction getDownloadingAction() {
         return new CacheManagerAction() {
             @Override
             public boolean preCheck() {
-                if (mTileSource instanceof OnlineTileSourceBase) {
+                if (mTileSource!=null && mTileSource instanceof OnlineTileSourceBase) {
```
monsieurtanuki, line 841: If I remember well, in Java a `null` is never an instance of anything. That fix is not needed.

osmdroid-android/src/main/java/org/osmdroid/tileprovider/MapTileProviderBasic.java
```diff
@@ -41,13 +41,13 @@
...
-	private final INetworkAvailablityCheck mNetworkAvailabilityCheck;
+	private INetworkAvailablityCheck mNetworkAvailabilityCheck;
 
 	/**
 	 * @since 6.1.0
 	 */
-	private final MapTileDownloader mDownloaderProvider;
-	private final MapTileApproximater mApproximationProvider;
+	private MapTileDownloader mDownloaderProvider;
+	private MapTileApproximater mApproximationProvider;
```
monsieurtanuki, line 50: There's no obvious reason not to keep them `final`.
The point of this PR is to create these fields differently, not to make it possible to change them afterwards, isn't it?

osmdroid-android/src/main/java/org/osmdroid/tileprovider/MapTileProviderBasic.java
```diff
@@ -78,6 +78,14 @@ public MapTileProviderBasic(final IRegisterReceiver pRegisterReceiver,
 			final INetworkAvailablityCheck aNetworkAvailablityCheck, final ITileSource pTileSource,
 			final Context pContext, final IFilesystemCache cacheWriter) {
 		super(pTileSource, pRegisterReceiver);
+		init(pRegisterReceiver, aNetworkAvailablityCheck, pTileSource, pContext, cacheWriter);
+	}
+
+	protected void init(IRegisterReceiver pRegisterReceiver,
```
monsieurtanuki, line 84: I can't see the point of new method `init`. I thought the purpose was mainly to override the new `create*` methods.

monsieurtanuki (commented):
Hi @pasniak!
Still something I don't fully understand with the `init` method and removing the `final` keywords.

osmdroid-android/src/main/java/org/osmdroid/tileprovider/cachemanager/CacheManager.java
pasniak: I believe I was getting some exception along the lines `null is not instance of OnlineTileSourceBase` (instead going to `else` branch). 

But yes, folding init() into constructor makes this redundant as final cannot be null. I will update to PR.

Note: If I force `createAssetsProvider` in my `MapTileProviderCustom` class to return null:
```
java.lang.RuntimeException: Unable to start activity ComponentInfo{com.myapp.debug/com.myapp.ui.activity.fullscreenactivity.FullscreenActivity}: java.lang.NullPointerException: Attempt to invoke virtual method 'int org.osmdroid.tileprovider.modules.MapTileModuleProviderBase.getMinimumZoomLevel()' on a null object reference
	at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:2913)
	at android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:3048)
	at android.app.servertransaction.LaunchActivityItem.execute(LaunchActivityItem.java:78)
	at android.app.servertransaction.TransactionExecutor.executeCallbacks(TransactionExecutor.java:108)
	at android.app.servertransaction.TransactionExecutor.execute(TransactionExecutor.java:68)
	at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1808)
	at android.os.Handler.dispatchMessage(Handler.java:106)
	at android.os.Looper.loop(Looper.java:193)
	at android.app.ActivityThread.main(ActivityThread.java:6669)
	at java.lang.reflect.Method.invoke(Native Method)
	at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:493)
	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:858)
Caused by: java.lang.NullPointerException: Attempt to invoke virtual method 'int org.osmdroid.tileprovider.modules.MapTileModuleProviderBase.getMinimumZoomLevel()' on a null object reference
	at org.osmdroid.tileprovider.modules.MapTileApproximater.computeZoomLevels(MapTileApproximater.java:64)
	at org.osmdroid.tileprovider.modules.MapTileApproximater.addProvider(MapTileApproximater.java:57)
	at org.osmdroid.tileprovider.MapTileProviderBasic.createApproximater(MapTileProviderBasic.java:133)
	at org.osmdroid.tileprovider.MapTileProviderBasic.<init>(MapTileProviderBasic.java:106)
	at org.osmdroid.tileprovider.MapTileProviderBasic.<init>(MapTileProviderBasic.java:70)
	at org.osmdroid.tileprovider.MapTileProviderBasic.<init>(MapTileProviderBasic.java:63)
	at org.osmdroid.tileprovider.MapTileProviderBasic.<init>(MapTileProviderBasic.java:56)
```

monsieurtanuki: Oh I guess if it doesn't exist now you could create an empty `MapTileFileStorageProviderBase` that returns nothing.

Or systematically test for `null` before adding each provider.

Or remove altogether each local variable (like `assetsProvider`); basically you don't need an asset provider and a cache provider and so on, you just need a list of providers, to be added to the `mTileProviderList` and to the `MapTileApproximater`. What do you think of that?

monsieurtanuki (commented):
Basically you answered all my comments, thank you @pasniak!

But I've just answered one of your new comments, and I think there's something interesting regarding lists instead of local variables to have a look at. Tell me what you think about it.

osmdroid-android/src/main/java/org/osmdroid/tileprovider/cachemanager/CacheManager.java
monsieurtanuki: @pasniak ping

spyhunter99:
is this good to merge?

monsieurtanuki (approved):
Except for the format change that does not help code review, I approve!

Issue #1738, opened 2021-09-06
Issue with Markers resolution


MKergall:
I'm completely stucked with a strange issue related to Markers resolution on high-density screens. 
I define my Marker icons with only 1 resolution, in res/drawable-mdpi directory. 

On one application (OSMNavigator), Marker icons are automatically resized according to the screen density - as needed and expected. 

On another application (NooTous), this doesn't work at all: Marker icons are not resized. 

Same version of osmdroid (6.1.11). Same target SDK (30). Same test device (API 30, 420 dpi). 
The only difference I can see is that the first app is based on old-style Activity model, while the second app is based on Fragment. 

(both applications here: https://github.com/MKergall/osmbonuspack)

## Issue Type

[ X ] Question

monsieurtanuki:
It wouldn't be the first time that Android messes up with drawable expected behaviors; there are already a couple of related issues.
My suggestion: just populate the 5 dpi drawable folders. If that does not work, that's another story.

MKergall:
@monsieurtanuki I also used this workaround. My real problem is that I have the same issue with icons downloaded from Internet (typically: icons referenced in KML files). And for those icons, no way to handle multiple resolutions. 

(Could you point me to related issues?)

monsieurtanuki:
@MKergall I don't know if this problem is caused by the Android version (30+) or the code.

It could be interesting to have a look at an optimization I coded in [`Marker.drawAt`](https://github.com/osmdroid/osmdroid/blob/master/osmdroid-android/src/main/java/org/osmdroid/views/overlay/Marker.java#L574-L590), that I would sum up as: `canvas.drawBitmap( mIcon...)` or `mIcon.draw(pCanvas)`.
Could you test if one of them works in your KML case?
Then, it would make sense to be able to force the choice of the good solution - for the moment it's coded as `if (mIcon instanceof BitmapDrawable)`.
And probably being able to say: I want this Marker's icon to be displayed as big as possible in a specified [width, height] rectangle (cf. `final int markerWidth = mIcon.getIntrinsicWidth()`) is a path to the solution.

MKergall:
All my icons are BitmapDrawables. 
I will try to test the other case: mIcon.draw(pCanvas) - I will have to subclass the Marker. 

Something even stranger: I tried with another test device: API 27, 420 dpi => 
- Marker based on a resource drawable-mdpi is correctly resized (!!!). 
- KML Marker (icon downloaded at runtime) is still not resized. 

[...]

PR #1740, opened 2021-09-10
Solving #1738

MKergall:


osmdroid-android/src/main/java/org/osmdroid/views/overlay/Marker.java
```diff
@@ -571,7 +571,8 @@ protected void drawAt(final Canvas pCanvas, final int pX, final int pY, final fl
             pCanvas.save();
             pCanvas.rotate(pOrientation, pX, pY);
         }
-        if (mIcon instanceof BitmapDrawable) { // optimization 3: (about 15% faster)
+        /*
```
monsieurtanuki, line 574: I would add a single comment like "Unfortunate optimization with displayed size side effects"

monsieurtanuki (approved):
Thank you @MKergall!
I just have one minor comment: please read it, feel free to ignore it, and then merge this PR.

osmdroid-android/src/main/java/org/osmdroid/views/overlay/Marker.java
MKergall: Sure. But I don't know how to add that change in the PR :-(

MKergall: Found!

Issue #1693, opened 2021-02-18
GPLv3 Licence is not allowed under Apache 2.0


mltpn:
## Issue Type

[x] Question
[x] Improvement
[x] Legality issue 

## Description 
With my limited knowledge I understood from [here](https://www.apache.org/licenses/GPL-compatibility.html), that the usage of GPLv3  software must not be published under Apache 2.0 license. 

We found the following class under GPLv3 in osmdroid: 

https://github.com/osmdroid/osmdroid/blob/master/osmdroid-android/src/main/java/org/osmdroid/views/overlay/CopyrightOverlay.java

6.1.2

YousefED:
@billthefarmer, I noticed you gave OSMDroid permission to use your code, but maybe wasn't aware that OSMDroid is Apache licensed. The GPL license was (mistakenly?) kept in place.

Could you confirm the license of the code can be changed to Apache 2?

billthefarmer:
That's fine &ndash; GPLv3 isn't appropriate for a library anyway.

YousefED:
> That's fine – GPLv3 isn't appropriate for a library anyway.

Thanks bill! see https://github.com/osmdroid/osmdroid/pull/1754

PR #1754, opened 2021-12-16
update copyrightoverlay license header

YousefED:
Update license of file according to https://github.com/osmdroid/osmdroid/issues/1693#issuecomment-995530625


YousefED:
cc @billthefarmer

PR #1762, opened 2022-01-12
Android11 MBTiles fix

ryanbehr:
Allows user to pass URI of file if using android 11 or higher. This will read the file and copy it into a temporary location so we can work with the MBTiles file. Hopefully the user does this from a thread because this can take a while depending on the size of the MBTiles file.


ryanbehr:
Needs more work.

Issue #1764, opened 2022-01-14
Mapsforge max zoom


bkekelic2:
[x] Question

## Description and/or steps/code to reproduce the problem

Hi, 
I'm currently using osmdorid with mapsforge provider with offline maps. 
I'm curious if there is possibility to zoom more than 20 on the map?

From this https://github.com/osmdroid/osmdroid/issues/329 I can see that you are apparently supporting max zoom up to 29, but here in [Mapsforge implementation](https://github.com/osmdroid/osmdroid/blob/master/osmdroid-mapsforge/src/main/java/org/osmdroid/mapsforge/MapsForgeTileSource.java#L39) I can see that it is limited up to 20. 

Thanks for any kind of help.

### Version of osmdroid the issue relates to:

osmdroid - 6.1.11
mapsforge - 0.10.0

bkekelic2:
Hi @monsieurtanuki. What is the source referred to?
1. To osmdroid implementation of Mapsforge (link I sent above)?
2. To the acutall Mapsforge implementation (on their github repo)?
3. To the .map file?
4. Something else

Finally, is there any posibility to do sth like that on my side or? 
Thanks in advance.

monsieurtanuki:
I don't know much about mapsforge.
From what I can read [in the code](https://github.com/osmdroid/osmdroid/blob/master/osmdroid-mapsforge/src/main/java/org/osmdroid/mapsforge/MapsForgeTileSource.java#L39) there's no explicit max zoom at 20 in osmdroid - the max zoom would depend on mapsforge data itself (I guess it's the `.map` file you mentioned):
```java
// Reasonable defaults ..
public static int MIN_ZOOM = 3;
public static int MAX_ZOOM = 20;

// ...

public static MapsForgeTileSource createFromFiles(File[] file) {
  //these settings are ignored and are set based on .map file info
  int minZoomLevel = MIN_ZOOM;
  int maxZoomLevel = MAX_ZOOM;
// ...
```

If you have good reasons to think that your `.map` file has a max zoom higher than 20, and that if you use osmdroid you're limited to 20, I suggest you copy `MapsForgeTileSource` into your own class `MyOwnMapsForgeTileSource` and set broader values, for instance:
```java
// Broadest defaults
public static int MIN_ZOOM = 0;
public static int MAX_ZOOM = 29;
```

If that works, that will be a perfect opportunity for a new PR.
Keep us posted.

bkekelic2:
[...]

PR #1767, opened 2022-01-19
Changed osmdroid-mapsforge MAX ZOOM value to 29

bkekelic2:
As discussed in this #1764 issue, osmdroid-mapsforge plugin could provide max zoom as it is provided by osmdroid plugin. 

Currently, when using osmdroid with mapsforge we are limited to use max zoom as 20 because of this [lower constant value](https://github.com/osmdroid/osmdroid/blob/master/osmdroid-mapsforge/src/main/java/org/osmdroid/mapsforge/MapsForgeTileSource.java#L39) and not able to use max zoom as 29 which is provided from osmdroid. 

As discussed in the issue, I tested if the osmdroid-mapsforge plugin can work with max zoom = 29 and it works perfectly.


monsieurtanuki (approved):
As simple as that ;)
Thank you @bkekelic2!

Issue #1733, opened 2021-08-20
Zoom buttons zoom twice when tapped slowly


hauke96:
## Issue Type

[ ] Question
[x] Bug
[ ] Improvement
[ ] Build system related
[ ] Performance
[ ] Documentation

## Description and/or steps/code to reproduce the problem

When tapping a zoom button very slowly, there is a zoom in/out happening on the `ACTION_DOWN` and also on the `ACTION_UP` event, so there are two zooms happening. This doesn't feel right and also [confused an user](https://github.com/hauke96/GeoNotes/issues/47) of an app of mine.

As far as my research goes this is caused by pull request #1471 in the `CustomZoomButtonsDisplay` by [this line](https://github.com/osmdroid/osmdroid/blob/dbb20901cac30b28a6a9349da097c22d9a962311/osmdroid-android/src/main/java/org/osmdroid/views/CustomZoomButtonsDisplay.java#L232). Here both `UP` and `DOWN` events are considered a touch which causes the `CustomZoomButtonsController` to trigger two `onZoom()` calls on the map controller (see e.g. [here](https://github.com/osmdroid/osmdroid/blob/dbb20901cac30b28a6a9349da097c22d9a962311/osmdroid-android/src/main/java/org/osmdroid/views/CustomZoomButtonsController.java#L194)).
If this is not a bug, maybe it's better to somehow configure the behavior on the `CustomZoomButtonsController` class?

### Steps to reproduce

1. Tap and hold on the zoom in/out button
  * Now the map zooms in/out
2. Wait a second until the animation completed
3. Release your finger
  * Another zoom happens

### If it's a bug, version(s) of android this affects:

As far as I know all versions of android are affected.

### Version of osmdroid the issue relates to:

6.1.8

monsieurtanuki:
Hi @hauke96!
It does look like a bug to me.
Feel free to create a PR - I guess `ACTION_DOWN` would be enough, what do you think?

hauke96:
Hi, I think `ACTION_UP` is better suited as it's the default behavior on normal buttons (on desktop systems as well as on android). I currently have little time so it will take a while until I work on this, but it's on my bucket list.

monsieurtanuki:
Fixed by #1775.

PR #1775, opened 2022-02-23
Only zoom in/out on ACTION_UP event

hauke96:
This fixes issue #1733. I tested this by building the osmdroid-android project and manually loading the resulting library into my app. The unwanted behavior described in #1733 is gone: Only one zoom takes place when I tap long on the zoom in or out button.

### DIFF
--- a/osmdroid-android/src/main/java/org/osmdroid/views/CustomZoomButtonsDisplay.java
+++ b/osmdroid-android/src/main/java/org/osmdroid/views/CustomZoomButtonsDisplay.java
@@ -228,8 +228,7 @@ public class CustomZoomButtonsDisplay {
      * @since 6.1.3
      */
     public boolean isTouched(final MotionEvent pMotionEvent, final boolean pInOrOut) {
-        int action = pMotionEvent.getAction();
-        if (action == MotionEvent.ACTION_DOWN || action == MotionEvent.ACTION_UP) {
+        if (pMotionEvent.getAction() == MotionEvent.ACTION_UP) {
             return isTouched((int) pMotionEvent.getX(), (int) pMotionEvent.getY(), pInOrOut);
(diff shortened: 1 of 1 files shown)


monsieurtanuki (approved):
Thank you @hauke96!

tvaknin (commented):
Probably the root cause of this bug: https://github.com/osmdroid/osmdroid/issues/1822.
can you revert this, please?

hauke96:
I don't think my changes are the root cause of bug #1822 . I don't understand why a long-press event is fired anyway, the `isTouched` function has nothing directly to do with it. Probably the calling code misinterprets the returned `false` value and triggers a long-press event.

Simply reverting my PR will reintroduce the bug #1733, so I won't just undo my changes. Unfortunately I don't have the time to look into this right now.

InI4:
@hauke96 as I figure, your code changes the code where the zoom button decides to be responsible, not where the actual (at times doubled) zoom is being done. Thus, as the button claims not responsible, the one event you removed from detection, isn't captured and leaked to the next overlay. As buttons are often held, this can often result in LongPress events up the listener hierarchy. This is of course not happen for everyone, as many UIs might not respond at all to the resulting LongPress event. However for my app this messed up subsequent Osmdroid releases.

Would be helpful, someone verified and hopefully pulled my pull request, so I can move back to master branch. @monsieurtanuki ?

monsieurtanuki:
@InI4 You mean #1876?

InI4:
Oh, my projects run so unsmooth, I apologize. Yes, I mean pull #1876, @monsieurtanuki. osmdroid 6.1.17 still bugging for me.

monsieurtanuki:
@InI4 Actually I'm not active on osmdroid anymore :-(

Issue #1806, opened 2022-05-19
mbtiles vs. geopackage EPSG 3857 north must be in [-85.05112877980658,85.05112877980658]


TFreudi1:
OsmDroid 6.1.2, Geopackage 6.1.3.
I have a simple mbtiles map which loaded fine in Osmdroid, when I converted it to a geopackage and try to load that , I get an error about the bounding box:
north must be in [-85.05112877980658,85.05112877980658]
which is clear because the max-bounding box in all EPSG 3857 files is  -20037508.3427892,-20037508.3427892,20037508.3427892,
20037508.3427892. I can patch the table gpkg_tile_matrix_set from the correct EPSG 3857 bounding to -180,-85,180,85 than I don't have the error, but I never see a map. Here is a sample file from the geopackage web site:
[river.zip](https://github.com/osmdroid/osmdroid/files/8730526/river.zip)

So the problem is that I can't use any raster map with EPSG 3857 in a geopackage.

spyhunter99:
should be fixed in the next release

Issue #1794, opened 2022-04-22
removing range checks on geopoint


spyhunter99:
Please note: issues related to or caused by osmbonuspack will be closed with a
suggestion to open an issue at https://github.com/MKergall/osmbonuspack

(feel free to delete whatever doesn't apply)

## Issue Type

[ ] Question
[ ] Bug
[ ] Improvement
[ ] Build system related
[ ] Performance
[ ] Documentation

## Description and/or steps/code to reproduce the problem

Currently we have range checks in geopoint that matches the bounds of the web mercator projection which is +-85 lat. This has causes a lot of undue problems with apps not validating this,causing a npe and crashing. I think we should drop that range check or at lease move to 90 degree threshold

Any thoughts in this?

### If it's a bug, version(s) of android this affects:

All

### Version of osmdroid the issue relates to:

6.1.2

monsieurtanuki:
Hi @spyhunter99!

I know that in many cases developers opened issues about bounding box exceptions.
Actually the problem is not the exception, but the given value that is beyond the limits.

Here my suggestions:
* a more verbose message that include the "wrong" value, like `"north (" + north + ") must be in " + tileSystem.toStringLatitudeSpan()` (right now we don't print the off-limit value)
* a bounding box leniency mode in the preferences, with
  * a STRICT mode (like now) (default)
  * a WARNING mode (no exception, just `Log.w` when off limits)
  * a TRUNC mode (when off limit, put the exceeded limit value instead)

What do you think of that?

spyhunter99:
i think changing it back to just a warning is more appropriate. rendering near the poles with web mercator is and always will be a bit funky

spyhunter99:
after thinking about this some more, i think  > tile system bounds but <= lat/lon maxs should be allowed, but beyond the min/max of the lat/lon coordinates should not be allowed.

monsieurtanuki:
The initial point of the exception was not "thy data is evil therefore thou shall crash", but the mere fact that we cannot compute data beyond the limits (e.g. `tan(pi/2)`). If "our" exception does not crash first (with a verbose message), then it's going to crash later (without a verbose message).

If you really want to be lenient, the "trunc" mode would make sense: if the value is beyond the limit, we put the limit instead.
That could also mean creating a sibling to `TileSystemWebMercator`, where we never crash and apply this "trunc" fix. Of course there might be side-effects (there already are) if you expect some bijection between geopoint and X01/Y01.

[...]

Issue #1784, opened 2022-04-01
Cannot zoom to bounding box


ArnyminerZ:
## Issue Type

- [*] Bug

## Description and/or steps/code to reproduce the problem

When trying to move the camera to the following bounding box:
```
N:38.703256; E:-0.48536999999999997; S:38.69581; W:-0.49455
```
This error occurs:
```
java.lang.IllegalArgumentException: north must be in [-85.05112877980658,85.05112877980658]
    at org.osmdroid.util.BoundingBox.set(BoundingBox.java:74)
    at org.osmdroid.views.Projection.refresh(Projection.java:699)
    at org.osmdroid.views.Projection.<init>(Projection.java:111)
    at org.osmdroid.views.Projection.<init>(Projection.java:123)
    at org.osmdroid.views.MapView.zoomToBoundingBox(MapView.java:564)
    at org.osmdroid.views.MapView.zoomToBoundingBox(MapView.java:590)
```
I'm calling it the following way:
```kotlin
val boundingBox = points.getBoundingBox()
mapView.zoomToBoundingBox(boundingBox, animate, 30)
```
Points is a list of `GeoPoint`:
```
[
    38.69601, -0.49455, 0.0,
    38.703256, -0.486723, 0.0
]
```
Which is extended here:

https://gist.github.com/ArnyminerZ/5ef0c46c4c3d42045005dd3459735aec#file-utils-kt-L31

### If it's a bug, version(s) of android this affects:

All

### Version of osmdroid the issue relates to:

`6.1.11`

spyhunter99:
should be fixed in the next release

Issue #1751, opened 2021-11-21
Import shp file got java.lang.IllegalArgumentException: north must be in [-85.05112877980658,85.05112877980658]


wogno:
Please note: issues related to or caused by osmbonuspack will be closed with a
suggestion to open an issue at https://github.com/MKergall/osmbonuspack

(feel free to delete whatever doesn't apply)

## Issue Type

[ ] Question
[x ] Bug
[ ] Improvement
[ ] Build system related
[ ] Performance
[ ] Documentation

## Description and/or steps/code to reproduce the problem
I loaded polygon from shape(.shp) files
![image](https://user-images.githubusercontent.com/31076843/142773057-d47c30ef-2975-4c78-8619-d078a87d7294.png)

## Environment
Android Studio on Windows and real phone(Infinix Note 6) for simulation

### If it's a bug, version(s) of android this affects:

All

### Version of osmdroid the issue relates to:

6.1.11

eclectice:
I notice this non-fatal IllegalArgumentException only when the Mapnik database where at its extreme zoom level (global view)

monsieurtanuki:
@wogno What I can say is:
* there's probably some latitude that is too high in your data
* the range for latitude with the default projection on osmdroid is +-85.05112877980658 (cf. https://github.com/osmdroid/osmdroid/blob/master/osmdroid-android/src/main/java/org/osmdroid/util/TileSystemWebMercator.java#L9-L10)
* in shape converter we truncate latitude so they can fit, but unfortunately we are too generous and min/max latitudes are +-90 (cf. https://github.com/osmdroid/osmdroid/blob/master/osmdroid-shape/src/main/java/org/osmdroid/shape/ShapeConverter.java#L187-L202)
* if in shape converter we truncated latitudes to +-85.05112877980658 we should probably be OK

Does that make sense?

@eclectice I think your issue is slightly different: ok it's still a problem of latitudes, but in your case it's about "multi-world" side-effects.

spyhunter99:
should be fixed in the next release

PR #1807, opened 2022-05-22
#1794 throwing is now optional with the default being "don't throw" f…

spyhunter99:
…or bounding boxes beyond that of the current tile system.

#1806  #1784 #1751 related issues that should be resolve by this


monsieurtanuki (approved):
Hi @spyhunter99!
I would have set `true` as the default, but that's OK, let's merge and see!

Issue #1808, opened 2022-05-24
refactor osmdroid-android-it into the OpenStreetMaps directory


spyhunter99:
Please note: issues related to or caused by osmbonuspack will be closed with a
suggestion to open an issue at https://github.com/MKergall/osmbonuspack

(feel free to delete whatever doesn't apply)

## Issue Type

[ ] Question
[ ] Bug
[ ] Improvement
[X] Build system related
[ ] Performance
[ ] Documentation

## Description and/or steps/code to reproduce the problem

really just a build system cleanup. There was why the integration tests were separated but I don't remember what it is. I'm certain it is no longer relevant now. Clean up in aisle 6

PR #1809, opened 2022-05-25
#1808 first attempt. i can't get any avd's up and running. i think th…

spyhunter99:
…is will work however on device permissioins may still be an issue

hold to merge


spyhunter99:
@monsieurtanuki  do these look familiar?

org.osmdroid.tileprovider.modules.MapTileProviderTest > test_jump_queue[SM-G781U - 12] FAILED
        java.lang.AssertionError: Three tiles in the list expected:<3> but was:<0>
        at org.junit.Assert.fail(Assert.java:88)

org.osmdroid.tileprovider.modules.MapTileProviderTest > test_order[SM-G781U - 12] FAILED
        java.lang.AssertionError: Three tiles in the list expected:<3> but was:<0>
        at org.junit.Assert.fail(Assert.java:88)

monsieurtanuki:
@spyhunter99 It doesn't look like a "bad luck" error with bad internet connection as it's local: here we're just testing the logic of queueing the tiles with fake tile source and tile provider.

Something fishy with the tile provider: max zoom level is set to 0, and we're requesting tiles of zoom levels 1, 2 and 3. Setting max level to 3 may fix that error.

spyhunter99:
that fixed the previous error, now DelayTest.testDelayOne is failing. I'm not sure what's going on here

monsieurtanuki:
> that fixed the previous error, now DelayTest.testDelayOne is failing. I'm not sure what's going on here

@spyhunter99 Could be that we're too precise with our tests and that in some cases when the computer is busy, the tests fail. My suggestion: add 1000ms to each hard-coded duration in `DelayTest`.

spyhunter99:
@monsieurtanuki do these test code changes make sense? I'm not sure i understand what's going on here. maybe 10x is too much to have an effective test?

PR #1820, opened 2022-06-17
Feature/#1809

broscr:
default zoom level could not be adjust now can be adjusted

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,7 +24,7 @@ repositories {
 }
 
 dependencies {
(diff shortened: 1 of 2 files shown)


monsieurtanuki (commented):
Hi @broscr!
I'm a bit puzzled:
* the PR is named after #1809, but there's no obvious link to #1809
* you comment is about the "default" zoom level, but your change is about setting the zoom level
* you're supposed to set the zoom level via the controller

broscr:
I think I made a mistake and didn't study the details thoroughly.

PR #1834, opened 2022-08-25
set custom Interpolate for mapController

ankiimation:


osmdroid-android/src/main/java/org/osmdroid/views/MapController.java
```diff
@@ -48,6 +52,7 @@ public class MapController implements IMapController, OnFirstLayoutListener {
     private double mTargetZoomLevel = 0;
 
     private Animator mCurrentAnimator;
+    public TimeInterpolator interpolator = new AccelerateDecelerateInterpolator();
```
monsieurtanuki, line 55:
```suggestion
    private TimeInterpolator mInterpolator;
```

osmdroid-android/src/main/java/org/osmdroid/views/MapController.java
```diff
@@ -166,6 +171,7 @@ public void animateTo(final IGeoPoint point, final Double pZoom, final Long pSpe
             if (mCurrentAnimator != null) {
                 mapAnimatorListener.onAnimationCancel(mCurrentAnimator);
             }
+            mapAnimator.setInterpolator(interpolator);
```
monsieurtanuki, line 174:
```suggestion
            mapAnimator.setInterpolator(mInterpolator);
```

osmdroid-android/src/main/java/org/osmdroid/views/MapController.java
```diff
@@ -400,8 +406,9 @@ public boolean zoomToFixing(double zoomLevel, int xPixel, int yPixel, Long zoomA
             } else {
                 zoomToAnimator.setDuration(zoomAnimationSpeed);
             }
-
+            zoomToAnimator.setInterpolator(interpolator);
```
monsieurtanuki, line 409:
```suggestion
            zoomToAnimator.setInterpolator(mInterpolator);
```

monsieurtanuki (commented):
Hi @ankiimation!

Thank you for your PR. My remarks:
* we should have the same behavior as before by default, therefore as if there was no interpolator
* obviously we need a setter - maybe a getter
* not sure: should it be the same interpolator for map and zoom?

Issue #1839, opened 2022-09-03
Support custom TileWriter implementation in MapTileProviderBasic


RolandMostoha:
## Issue Type

[X] Improvement

## Description and/or steps/code to reproduce the problem

Steps: Try to use a custom `SqlTileWriter` implementation using a `MapTileProviderBasic` descendant.

### Problem

1. Even though it has a method to customize `MapTileFileStorageProviderBase`, it's static so it can't be overwritten
2. The passed `IFilesystemCache` parameter doesn't get injected into the corresponding `CacheProvider`

That means you have to hard copy at least 4 classes to make this work.

### Solution

1. Make `MapTileProviderBasic.getMapTileFileStorageProviderBase(...)` protected instead of static.
4. Enable `MapTileFilesystemProvider` and `MapTileSqlCacheProvider` to inject custom `TileWriter` implementation

### Usage

The following PR makes it as easy as:

```
class CustomMapTileProviderBasic(
    context: Context,
    tileSource: ITileSource,
    customTileWriter: TileWriter
) : MapTileProviderBasic(context, tileSource, customTileWriter)
```

### Version of osmdroid the issue relates to:

6.1.15

RolandMostoha:
I follow @monsieurtanuki  advice and copy the files separately.

PR #1840, opened 2022-09-03
Support custom TileWriter implementation in MapTileProviderBasic

RolandMostoha:
https://github.com/osmdroid/osmdroid/issues/1839

### Problem

1. Even though it has a method to customize `MapTileFileStorageProviderBase`, it's static so it can't be overwritten
2. The passed `IFilesystemCache` parameter doesn't get injected into the corresponding `CacheProvider`

That means you have to hard copy at least 4 classes to make this work.

### Solution

1. Make `MapTileProviderBasic.getMapTileFileStorageProviderBase(...)` protected instead of static.
4. Enable `MapTileFilesystemProvider` and `MapTileSqlCacheProvider` to inject custom `TileWriter` implementation

### DIFF
--- a/osmdroid-android/src/main/java/org/osmdroid/tileprovider/MapTileProviderBasic.java
+++ b/osmdroid-android/src/main/java/org/osmdroid/tileprovider/MapTileProviderBasic.java
@@ -203,15 +203,18 @@ public class MapTileProviderBasic extends MapTileProviderArray implements IMapTi
 	 * @since 6.0.3
 	 * cf. https://github.com/osmdroid/osmdroid/issues/1172
 	 */
-	public static MapTileFileStorageProviderBase getMapTileFileStorageProviderBase(
+	protected MapTileFileStorageProviderBase getMapTileFileStorageProviderBase(
 			final IRegisterReceiver pRegisterReceiver,
 			final ITileSource pTileSource,
 			final IFilesystemCache pTileWriter
 	) {
 		if (pTileWriter instanceof TileWriter) {
-			return new MapTileFilesystemProvider(pRegisterReceiver, pTileSource);
+			return new MapTileFilesystemProvider(pRegisterReceiver, pTileSource, (TileWriter) pTileWriter);
+		} else if (pTileWriter instanceof SqlTileWriter) {
+			return new MapTileSqlCacheProvider(pRegisterReceiver, pTileSource, (SqlTileWriter) pTileWriter);
+		} else {
+			return new MapTileSqlCacheProvider(pRegisterReceiver, pTileSource);
 		}
-		return new MapTileSqlCacheProvider(pRegisterReceiver, pTileSource);
 	}
 
 	/**
--- a/osmdroid-android/src/main/java/org/osmdroid/tileprovider/modules/MapTileFilesystemProvider.java
+++ b/osmdroid-android/src/main/java/org/osmdroid/tileprovider/modules/MapTileFilesystemProvider.java
@@ -32,7 +32,7 @@ public class MapTileFilesystemProvider extends MapTileFileStorageProviderBase {
     // Fields
     // ===========================================================
 
-    private final TileWriter mWriter = new TileWriter();
+    private TileWriter mWriter = new TileWriter();
     private final AtomicReference<ITileSource> mTileSource = new AtomicReference<ITileSource>();
 
     // ===========================================================
@@ -68,6 +68,19 @@ public class MapTileFilesystemProvider extends MapTileFileStorageProviderBase {
         setTileSource(pTileSource);
 
         mWriter.setMaximumCachedFileAge(pMaximumCachedFileAge);
+    }
+
+    /**
+     * Constructor that supports injecting custom {@link TileWriter} implementation.
+     *
+     * @since 6.1.15
+     */
+    public MapTileFilesystemProvider(final IRegisterReceiver pRegisterReceiver,
+                                     final ITileSource aTileSource,
+                                     final TileWriter tileWriter) {
+        this(pRegisterReceiver, aTileSource);
+
+        this.mWriter = tileWriter;
     }
     // ===========================================================
     // Getter & Setter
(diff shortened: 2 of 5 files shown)


osmdroid-android/src/main/java/org/osmdroid/tileprovider/MapTileProviderBasic.java
```diff
@@ -203,15 +203,18 @@ protected boolean isDowngradedMode(final long pMapTileIndex) {
 	 * @since 6.0.3
 	 * cf. https://github.com/osmdroid/osmdroid/issues/1172
 	 */
-	public static MapTileFileStorageProviderBase getMapTileFileStorageProviderBase(
+	protected MapTileFileStorageProviderBase getMapTileFileStorageProviderBase(
```
monsieurtanuki, line 206: We can't just do that, for legacy reason.
Please create a new `protected` method if needed but leave the `public static` one.

osmdroid-android/src/main/java/org/osmdroid/tileprovider/modules/MapTileFilesystemProvider.java
```diff
@@ -32,7 +32,7 @@ public class MapTileFilesystemProvider extends MapTileFileStorageProviderBase {
     // Fields
     // ===========================================================
 
-    private final TileWriter mWriter = new TileWriter();
+    private TileWriter mWriter = new TileWriter();
```
monsieurtanuki, line 35: Please leave it `final`, for code clarity.
If needed, put a default value in constructors.

osmdroid-android/src/main/java/org/osmdroid/tileprovider/modules/MapTileSqlCacheProvider.java
```diff
@@ -33,7 +33,7 @@ public class MapTileSqlCacheProvider extends MapTileFileStorageProviderBase {
     // ===========================================================
 
     private final AtomicReference<ITileSource> mTileSource = new AtomicReference<ITileSource>();
-    private SqlTileWriter mWriter;
+    private SqlTileWriter mWriter = new SqlTileWriter();
```
monsieurtanuki, line 36: I don't follow you here.

osmdroid-android/src/main/java/org/osmdroid/tileprovider/MapTileProviderBasic.java
```diff
@@ -203,15 +203,18 @@ protected boolean isDowngradedMode(final long pMapTileIndex) {
...
 			final ITileSource pTileSource,
 			final IFilesystemCache pTileWriter
 	) {
 		if (pTileWriter instanceof TileWriter) {
-			return new MapTileFilesystemProvider(pRegisterReceiver, pTileSource);
+			return new MapTileFilesystemProvider(pRegisterReceiver, pTileSource, (TileWriter) pTileWriter);
+		} else if (pTileWriter instanceof SqlTileWriter) {
+			return new MapTileSqlCacheProvider(pRegisterReceiver, pTileSource, (SqlTileWriter) pTileWriter);
+		} else {
+			return new MapTileSqlCacheProvider(pRegisterReceiver, pTileSource);
```
monsieurtanuki, line 216: Very strange.

monsieurtanuki (commented):
Hi @RolandMostoha!
I cannot say that I followed what you were trying to code - but it's been a long time since I actually coded in osmdroid.
There are some things that you should absolutely not do, like changing a `public static` method into a `protected` one. I'm happy if it solves your issues, you can understand that not everybody will be happy with that.
Plus some part of your code looked strange to me.
I suggest that you forget your own specific use case, and just code as if you wanted to add flexibility (actually I do not see why passing the correct `final IFilesystemCache cacheWriter` value in the `Basic` constructor would not work).
By the way `MapTileProviderBasic` was meant to be a _basic_ class, a starting point for developers to find inspiration on, not the be all end all that works in all cases.

osmdroid-android/src/main/java/org/osmdroid/tileprovider/MapTileProviderBasic.java
RolandMostoha: As I saw the original use of this method, the only reason it was static because we used it in `SampleLieFi` and `GeoPackageProvider` for creating `MapTileFileStorageProviderBase ` instances, which I think is unnecessary and blocks the overridability of the method.

You can check it out if you search for the usages of the original `getMapTileFileStorageProviderBase()` method

osmdroid-android/src/main/java/org/osmdroid/tileprovider/MapTileProviderBasic.java
RolandMostoha: The problem with the original method is that it does not inject the passed `IFilesystemCache pTileWriter` parameter to the created `MapTileSqlCacheProvider` instances. Therefore, even if I create my own CustomTileWriter and pass it, the corresponding `CacheProvider` won't use it -> the method is not doing what it is intended to do.

osmdroid-android/src/main/java/org/osmdroid/tileprovider/modules/MapTileSqlCacheProvider.java
RolandMostoha: The default implementation is `SqlTileWriter` and with the use of the new constructor, you can override it with your custom implementation.

osmdroid-android/src/main/java/org/osmdroid/tileprovider/modules/MapTileFilesystemProvider.java
RolandMostoha: Good point, thanks.

RolandMostoha:
Heya @monsieurtanuki!

Thanks for the quick review, I appreciate you response :) I can understand if you consider the changes risky and indeed, I can copy these classes without any serious effort so that might be right direction in my case.

Issue #1707, opened 2021-03-19
Align, if possible, geopackage support


at1839:
to  ngageoint/geopackage-android-map 5.0.0 

:)

Until 3.0 ...
geopackage-core version updated to 3.0.0
Package names in dependent classes must be updated

TyL. Paolo

Issue #1706, opened 2021-03-19
GeoPackage database import error; possible a bug?


at1839:
I’m dealing with a small test App to load and display a GeoPackage raster layer.
The source .gpkg file archive is on getExternalFilesDir(null) osmdroid/files path so must not be a matter of permission, but anyway I have evidences that the folder is writable and the source file is loaded and parsed.

The App works onboard an old Lg G2, Android 7 Api 25. The map shows up in the window.
On a newer Xiaomi Poco F1, Android 10 Api 29 the App fires up the below error

W/System.err: mil.nga.geopackage.GeoPackageException: Invalid GeoPackage database file
W/System.err: Caused by: mil.nga.geopackage.GeoPackageException: Failed to insert GeoPackage metadata. Name: cesaneMTB, External Path: null

Now I cannot figure out what’s on different Api … The source file is the same so why here it’s a valid gpkg and there it’s not … really confused and any debug strategy is landing to nowhere.

Gdal says 

gdalinfo cesaneMTB.gpkg -oo TABLE=cesaneMTB

Files: cesaneMTB.gpkg
            LENGTHUNIT["metre",1]]],
    PRIMEM["Greenwich",0,
        ANGLEUNIT["degree",0.0174532925199433]],
    CS[ellipsoidal,2],
        AXIS["geodetic latitude (Lat)",north,
            ORDER[1],
            ANGLEUNIT["degree",0.0174532925199433]],
        AXIS["geodetic longitude (Lon)",east,
            ORDER[2],
            ANGLEUNIT["degree",0.0174532925199433]],
    ID["EPSG",4326]]
Data axis to CRS axis mapping: 2,1
Origin = (12.668898420000000,43.749655154000003)
Pixel Size = (0.000009685365377,-0.000009685365303)
Metadata:
  IDENTIFIER=cesaneMTB
  ZOOM_LEVEL=13
Image Structure Metadata:
  INTERLEAVE=PIXEL
Corner Coordinates:
Upper Left  (  12.6688984,  43.7496552) ( 12d40' 8.03"E, 43d44'58.76"N)
Lower Left  (  12.6688984,  43.6603851) ( 12d40' 8.03"E, 43d39'37.39"N)
Upper Right (  12.8348378,  43.7496552) ( 12d50' 5.42"E, 43d44'58.76"N)
Lower Right (  12.8348378,  43.6603851) ( 12d50' 5.42"E, 43d39'37.39"N)
[...]

Issue #1684, opened 2021-01-23
Using geopackage tile occur SQLiteException


hello-willy:
According to the example, I write the following code：
```
Set<File> mapFiles = new HashSet<>();
mapFiles.add(new File("data/data/com.imagesky.app/123.gpkg"));
File[] maps = new File[mapFiles.size()];
maps = mapFiles.toArray(maps);

GeoPackageProvider geoPackageProvider = new GeoPackageProvider(maps, this);
List<GeopackageRasterTileSource> sources = geoPackageProvider.geoPackageMapTileModuleProvider().getTileSources();

XYTileSource tileSource = geoPackageProvider.getTileSource(sources.get(0).getDatabase(),
        sources.get(0).getTableDao());
BoundingBox bbox = sources.get(0).getBounds();
mMapView.setTileProvider(geoPackageProvider);
mMapView.setTileSource(tileSource);
```
There was an exception：
![123](https://user-images.githubusercontent.com/9379589/105568096-d28bd880-5d71-11eb-969e-e7fdf1ed1b54.jpg)

Version Info
```
Version Info:
implementation 'org.osmdroid:osmdroid-android:6.1.5'
implementation 'org.osmdroid:osmdroid-geopackage:6.1.5'
```
I need help, thank you!!

spyhunter99:
Is it possible that the geopackge file is on a read only partition or in a location where the app cannot write to it? permissions maybe? If not, it may be an upstream issue. We are using a 3rd party library for the geopackage support. NGA i think. It's on github

hello-willy:
> Is it possible that the geopackge file is on a read only partition or in a location where the app cannot write to it? permissions maybe? If not, it may be an upstream issue. We are using a 3rd party library for the geopackage support. NGA i think. It's on github

thank you for your reply.
I'm sure the path is accessible.

AndresLpzFdz:
Hello. Testing with Android 25 works fine, but I have the same issue here with:

- Android 29. 
- 'org.osmdroid:osmdroid-android:6.1.10'
- 'org.osmdroid:osmdroid-geopackage:6.1.10'

@spyhunter99 It´s not related with permissions or accessibility to filesystems. Same issue here: 
- https://github.com/osmdroid/osmdroid/issues/1706
- https://github.com/osmdroid/osmdroid/issues/1491

I made debug and found that:

[...]

Issue #1709, opened 2021-03-28
Offline map overlay over Mapnik


at1839:
Please note: issues related to or caused by osmbonuspack will be closed with a
suggestion to open an issue at https://github.com/MKergall/osmbonuspack

(feel free to delete whatever doesn't apply)

## Issue Type

[ x] Question

The goal is to overlay tiles from offline maps to online, say Mapnik. We start with a TIFF scan of a map in paper format. The .sqlite archive is by Mobac where the .gpkg is Qgis made. 

See below a bunch of wild code, just  for testing purpose. I run the code, via Android Studio, onboard a quite old LG G2 Mini with LineageOS, Android version 7.1.2; actually I tried with most recent ones but I have a lot of issues, as an example onboard Android 11 the Geopackage stuff fires up an obscure database error and refuse to import the map.

Do case …

Case 0 works.

Case 1 works. I have white tiles around my map but not an issue at now.

Case 2 works, again with white tiles around.

Case 3 mostly works. I see Mapnik’s tiles around my map but … those Mapnik tiles are just the previously cached ones, if I pan the map to a new square looks like online tiles are not downloaded, as if someone forced setUseDataConnection(false), ouch.

Case 4 doesn’t work. I do have nice Mapnik tiles and no trace of the Geopackage’s ones.

:)

Disclaimer. I’m quite a beginner. First at all I’d need to know if I’m on the right track or I badly misunderstood how to deal with the above goal.

And if I’m not dead wrong, what’s about my code? Please help :)

TyL. Paolo.

 
`    @Override
    public void onCreate(final Bundle savedInstanceState) {
           super.onCreate(savedInstanceState);

       Context ctx = getApplicationContext();
       Configuration.getInstance().load(ctx, PreferenceManager.getDefaultSharedPreferences(ctx));

       setContentView(R.layout.activity_main);
       mMapView = (MapView) findViewById(R.id.map);
       mMapView.setBuiltInZoomControls(true);
       mMapView.setMultiTouchControls(true);

// Sqlite stuff
       File fileSqlite = new File(getExternalFilesDir(null), "osmdroid/cesaneMTB.sqlite");
       OfflineTileProvider cesaneProvider = new OfflineTileProvider(new SimpleRegisterReceiver(this), new File[]{fileSqlite});
       ITileSource cesaneSource = FileBasedTileSource.getSource("cesaneMTB");
       TilesOverlay cesaneOverlay = new TilesOverlay(cesaneProvider, ctx);

[...]

PR #1853, opened 2022-10-12
Updated NGA GEOPackage core from 2.0.1 to 6.5.0

PWRxPSYCHO:
#1707 
#1684 
#1706 

Looking to update the NGA-Geopackage libraries within OSMDroid to the latest versions.

### DIFF
--- a/OpenStreetMapViewer/src/main/java/org/osmdroid/samplefragments/geopackage/GeopackageFeatureTiles.java
+++ b/OpenStreetMapViewer/src/main/java/org/osmdroid/samplefragments/geopackage/GeopackageFeatureTiles.java
@@ -34,8 +34,8 @@ import java.util.List;
 import java.util.Set;
 
 import mil.nga.geopackage.GeoPackage;
+import mil.nga.geopackage.GeoPackageFactory;
 import mil.nga.geopackage.GeoPackageManager;
-import mil.nga.geopackage.factory.GeoPackageFactory;
 
 import static org.osmdroid.samplefragments.events.SampleMapEventListener.df;
 
--- a/OpenStreetMapViewer/src/main/java/org/osmdroid/samplefragments/geopackage/GeopackageFeatures.java
+++ b/OpenStreetMapViewer/src/main/java/org/osmdroid/samplefragments/geopackage/GeopackageFeatures.java
@@ -43,7 +43,7 @@ import mil.nga.geopackage.features.user.FeatureCursor;
 import mil.nga.geopackage.features.user.FeatureDao;
 import mil.nga.geopackage.features.user.FeatureRow;
 import mil.nga.geopackage.geom.GeoPackageGeometryData;
-import mil.nga.wkb.geom.Geometry;
+import mil.nga.sf.Geometry;
 
 import static org.osmdroid.samplefragments.events.SampleMapEventListener.df;
 
--- a/OpenStreetMapViewer/src/main/java/org/osmdroid/samplefragments/geopackage/GeopackageFeatures.java
+++ b/OpenStreetMapViewer/src/main/java/org/osmdroid/samplefragments/geopackage/GeopackageFeatures.java
@@ -37,8 +37,8 @@ import java.util.List;
 import java.util.Set;
 
 import mil.nga.geopackage.GeoPackage;
+import mil.nga.geopackage.GeoPackageFactory;
 import mil.nga.geopackage.GeoPackageManager;
-import mil.nga.geopackage.factory.GeoPackageFactory;
 import mil.nga.geopackage.features.user.FeatureCursor;
 import mil.nga.geopackage.features.user.FeatureDao;
 import mil.nga.geopackage.features.user.FeatureRow;
--- a/OpenStreetMapViewer/src/main/java/org/osmdroid/samplefragments/tileproviders/OfflinePickerSample.java
+++ b/OpenStreetMapViewer/src/main/java/org/osmdroid/samplefragments/tileproviders/OfflinePickerSample.java
@@ -45,8 +45,8 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Set;
 
+import mil.nga.geopackage.GeoPackageFactory;
 import mil.nga.geopackage.GeoPackageManager;
-import mil.nga.geopackage.factory.GeoPackageFactory;
 
 /**
  * lets you pick one or more offline tile archives/providers
--- a/osmdroid-geopackage/build.gradle
+++ b/osmdroid-geopackage/build.gradle
@@ -25,7 +25,10 @@ android {
             withJavadocJar()
         }
     }
-
+    packagingOptions {
+        excludes += "DebugProbesKt.bin"
+        excludes += "com/j256/ormlite/core/LICENSE.txt"
+    }
 }
 
 dependencies {
@@ -39,9 +42,11 @@ dependencies {
         exclude group: 'com.google.android.gms', module:'play-services'
         exclude group: 'com.google.maps.android', module:'android-maps-utils'
         exclude group: 'com.android.support',module: 'support-v13'
+        exclude group: 'org.osmdroid.gpkg'
+        exclude module: 'ormlite-core'
+        exclude group: 'com.j256.ormlite'
     }
 
-    api 'com.j256.ormlite:ormlite-android:6.1'
     api 'mil.nga.geopackage:geopackage-core:6.5.0'
     api 'ar.com.hjg:pngj:2.1.0'
     api 'mil.nga:tiff:3.0.0'
--- a/osmdroid-geopackage/build.gradle
+++ b/osmdroid-geopackage/build.gradle
@@ -35,7 +35,7 @@ dependencies {
     testImplementation 'junit:junit:4.13.2'
 
     api  project(':osmdroid-android')
-    api ("mil.nga.geopackage:geopackage-android:2.0.1"){
+    api ("mil.nga.geopackage:geopackage-android:6.5.0"){
         exclude group: 'com.google.android.gms', module:'play-services'
         exclude group: 'com.google.maps.android', module:'android-maps-utils'
         exclude group: 'com.android.support',module: 'support-v13'
--- a/osmdroid-geopackage/build.gradle
+++ b/osmdroid-geopackage/build.gradle
@@ -44,10 +44,10 @@ dependencies {
     api 'com.j256.ormlite:ormlite-android:5.0'
     api 'mil.nga.geopackage:geopackage-core:6.5.0'
     api 'ar.com.hjg:pngj:2.1.0'
-    api 'mil.nga:tiff:2.0.0'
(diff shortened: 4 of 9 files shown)


eclectice:
Because **GeoPackage** now supports ***DocumentFile*** natively, we may be able to provide support for ***DocumentFile*** as an alternative to ***File*** (polymorphs methods to support both classes) so Geopackage files can be opened from the shared storage with respect to scoped storage introduction since Android 10+.

monsieurtanuki (approved):
Looks good to me @PWRxPSYCHO!

PWRxPSYCHO:
> Because **GeoPackage** now supports _**DocumentFile**_ natively, we may be able to provide support for _**DocumentFile**_ as an alternative to _**File**_ (polymorphs methods to support both classes) so Geopackage files can be opened from the shared storage with respect to scoped storage introduction since Android 10+.
> 
> Even, to make it more flexible, we can simply use the _**GeoPackage**_ class instead of the _**File**_ class. A Geopackage file can have multiple tiles tables storing different tile maps. With a single tile source approach, we can separately handle each tile table with a separate **GeoPackageMapTileModuleProvider()**.
> 
> ```java
> Map<String, MapTileModuleProviderBase> providers = new LinkedHashMap<>();
> //tiles object stores info about each geopackage file data (database, tile table, current tile table active state)
> //so, we can granularly control its data that we want to manipulate
> //key is database:table string format
> String key = String.join(":", tiles.getDatabase(), tiles.getName());
> //handle one geopackage
> GeoPackage geoPackage = geoPackageViewModel.getGeoPackage(tiles.getDatabase());
> GeoPackageProvider geoPackageProvider = new GeoPackageProvider(geoPackage, activity);
> 
> if (tileOverlays.get(key) == null) {
>     TilesOverlay tilesOverlay = new TilesOverlay(geoPackageProvider, activity); //with default MapTileProviderBase
>     tilesOverlay.setLoadingBackgroundColor(Color.TRANSPARENT);
>     tilesOverlay.setLoadingLineColor(Color.TRANSPARENT);
>     tilesOverlay.setLoadingDrawable(null);
> 
>     tileOverlays.put(key, tilesOverlay);
> }
> 
> //...
> if (!mapView.getOverlayManager().overlays().contains(tileOverlays.get(key))) {
>     mapView.getOverlayManager().add(tileOverlays.get(key));
>     mapView.setDrawingCacheBackgroundColor(Color.TRANSPARENT);
>     mapView.setBackgroundColor(Color.TRANSPARENT);
> }
> 
> //this geopackage may have multiple tile tables
> GeoPackageMapTileModuleProvider provider = geoPackageProvider.geoPackageMapTileModuleProvider();
> providers.put(key, provider);
> GeoPackageMapTileModuleProvider provider = (GeoPackageMapTileModuleProvider) providers.get(key);
> List<GeopackageRasterTileSource> tileSources = provider
>         .getTileSources(tiles.getDatabase(), tiles.getName(), tiles.isActive());
> tileSources.forEach((tileSource) -> {
>     if (tileSource != null) {
>         String tileSourceName = ((GeopackageRasterTileSource) tileSource).name();
>         //...
>         final MapTileFileStorageProviderBase cacheProvider =
>                 MapTileProviderBasic.getMapTileFileStorageProviderBase(simpleRegisterReceiver, tileSource, tileWriter);
>         providers.put(String.join("-", key, "cacheProvider"), cacheProvider);
>         //...
>         mapView.setTileSource(tileSource);
>         mapView.zoomToBoundingBox(((GeopackageRasterTileSource) tileSource).getBounds(), true);
>         mapView.getController().setZoom(((GeopackageRasterTileSource) tileSource).getMinimumZoomLevel()*1.0);
>     }
> });
> String[] providerNames = new String[providers.size()];
> int i = 0;
> for (Map.Entry<String, MapTileModuleProviderBase> entry : providers.entrySet()) {
> 	if (entry.getKey() != null && entry.getValue() != null) {
> 		providerNames[i++] = String.format("%s (%s)", entry.getKey(), entry.getValue().getClass().getSimpleName());
> 	}
> }
> 
> MapTileProviderArray obj = new MapTileProviderArray(tileSource, new SimpleRegisterReceiver(activity), providerArray);
> mapView.setTileProvider(obj);
> ```
> 
> This improvement is to limit _**GeoPackageProvider()**_ to a single _**GeoPackage**_ file so that we may control its tile source(s) more precisely with the required tileprovider(s) to manage a separate tile cache. It will be very simple to control the tile source hide/unhide feature (which is currently missing).
> 
> From
> 
> ```java
> public GeoPackageProvider(final IRegisterReceiver pRegisterReceiver,
>                               final INetworkAvailablityCheck aNetworkAvailablityCheck, final ITileSource pTileSource,
>                               final Context pContext, final IFilesystemCache cacheWriter, File[] databases)
> ```
> 
> to
> 
> ```java
> public GeoPackageProvider(final IRegisterReceiver pRegisterReceiver,
>                                  final INetworkAvailablityCheck aNetworkAvailablityCheck, final ITileSource pTileSource,
>                                  final Context pContext, final IFilesystemCache cacheWriter, GeoPackage geoPackage)
> ```
> 
> From
> 
> ```java
> public GeoPackageMapTileModuleProvider(File[] pFile,
>                                            final Context context, IFilesystemCache cache)
> ```
> 
> to
> 
> ```java
> public GeoPackageMapTileModuleProvider(final Context context, IFilesystemCache cache, GeoPackage geoPackage)
> ```
> 
> This is my version to idealize the above proposal. I've also hacked a way to control the range of zoom levels allowed on the tile source of each tile table based on the data from the GeoPackage file itself in the `getMapTile()` method rather than relying on the `getMaximumZoomLevel()` method. The latter's original method has a global effect: if there are multiple GeoPackage files loaded, the first one loaded will restrict the max zoom level allowed in the _**MapView**_ which is not desired. The same global effect occurs with the min zoom level of the original `getMinimumZoomLevel()` method.
> GeoPackageProvider
> GeoPackageMapTileModuleProvider
> GeopackageRasterTileSource
> 
> Discussions on my approach can be found here: [#1709 (comment)](https://github.com/osmdroid/osmdroid/issues/1709#issuecomment-1217772054)
> 
> Next support should be able to tell whether a table in a GeoPackage file is a features-based (vector type), a tiles-based (raster type), or feature tiles.

@eclectice Are these changes something that will be added to OSMDroid in the next release? Or is this a suggestion for developers who wish to use this feature going forward?

PWRxPSYCHO:
@monsieurtanuki The CI looks like it is failing for an API version. Should I address that? Or could that be a breaking change for the application as a whole?

monsieurtanuki:
> The CI looks like it is failing for an API version. Should I address that? Or could that be a breaking change for the application as a whole?

That does look like a breaking change.
Given that this could impact only the geopackage package (in its own `build.gradle`), and assuming that your PR is really worth it, we could upgrade only the geopackage package to sdk `32`.
Or, maybe you're able to identify what is causing this "min sdk 32" error - if it's something unimportant we can get rid of it.
Actually, I don't know very much about those things. @spyhunter99?

eclectice:
@PWRxPSYCHO I suggest it as a proposal to be implemented once your change has been successfully committed, as I was unable to push my change owing to an unknown OsmDroid build issue with Gradle-Fury last time. I'm also unfamiliar with the Git tool.

On minimum SDK version, there is a note regarding Android platform support 
> Starting with AGP 7.3.0-beta05, the highest supported minimum SDK version is 33 (you can use minSdk = 33). The minimum SDK represents the minimum version of Android that your app can run on and is set in the app-level build.gradle file.

https://developer.android.com/studio/releases/gradle-plugin#android-platform-support

monsieurtanuki:
> Starting with AGP 7.3.0-beta05

We don't need to go that high, do we?

eclectice:
> > Starting with AGP 7.3.0-beta05
> 
> We don't need to go that high, do we?

My current AGP 7.3.0 build (which I will upgrade soon) generated this lint error demanding minSDKversion 24. https://issuetracker.google.com/issues/219091668?hl=id

monsieurtanuki:
> My current AGP 7.3.0 build (which I will update this info soon) generated this lint error demanding minSDKversion 24.

That would be problematic for my Lollipop smartphone (21).

eclectice:
> Latest CI build issue is related to this: [j256/ormlite-android#141](https://github.com/j256/ormlite-android/issues/141)
> 
> The compileSDK bump is no longer required

It is just a variable but it is not applied in the `android {}` entry

eclectice:
> Latest CI build issue is related to this: [j256/ormlite-android#141](https://github.com/j256/ormlite-android/issues/141)
> 
> The compileSDK bump is no longer required

This is how I solved dependency duplicates between libraries that use different versions, and how I implemented my local version of OsmDroid Geopackage API support without using the same class package names. For example, I created ***SafGeoPackageProvider*** as a replacement for OsmDroid ***GeoPackageProvider*** version so that I could use the most recent GeoPackage API version.

```gradle
    api 'mil.nga.geopackage:geopackage-android:6.6.0'
    implementation('org.osmdroid:osmdroid-geopackage:6.1.14') {
        exclude group: 'org.osmdroid.gpkg'
        exclude module: 'ormlite-core'
        exclude group: 'com.j256.ormlite'
    }
```

PWRxPSYCHO:
Last issue seems to be on the DocumentFile

There are 2 different versions being used. Android.x 1.0.1 & android support: 28.0.0

eclectice:
> > Latest CI build issue is related to this: [j256/ormlite-android#141](https://github.com/j256/ormlite-android/issues/141)
> > The compileSDK bump is no longer required
> 
> It is just a variable but it is not applied in the `android {}` entry

Hmm, ***compileSdkVersion*** is still required in the `android {}` section since AGP complains about it if missing during the build.

PWRxPSYCHO:
> > > Latest CI build issue is related to this: [j256/ormlite-android#141](https://github.com/j256/ormlite-android/issues/141)
> > > The compileSDK bump is no longer required
> > 
> > 
> > It is just a variable but it is not applied in the `android {}` entry
> 
> Hmm, _**compileSdkVersion**_ is still required in the `android {}` section since AGP complains about it if missing during the build.

If you checkout the CI run here: https://github.com/ScriptTactics/osmdroid/actions/runs/3244600273/jobs/5320984386#step:5:327

You'll see that it's clashing with the document file issue. Android-Support brings in version 28.0.0 while the NGA-Geopackage-Android brings in Androidx which uses version 1.0.1.

I am also running CI on my newest PR: #1856 -> 
https://github.com/ScriptTactics/osmdroid/actions/runs/3245757323/jobs/5323696698

This has all the changes in from this PR and the other PR. (Removing all android-support libraries and updating to Androidx) and there does not seem to be any issues about dependencies or duplicates

eclectice:
Because of changes in the GeoPackage API since v2.0.1, this WIKI must be updated.
https://github.com/osmdroid/osmdroid/wiki/Geopackage-Support

spyhunter99:
merged, feel free to update the wiki

Issue #1453, opened 2019-12-18
from android.support to androidx


monsieurtanuki:
## Issue Type

- [x] Question
- [x] Improvement

## Description and/or steps/code to reproduce the problem

android.support library is kind of deprecated.
Shouldn't we roll out to [androidx](https://developer.android.com/jetpack/androidx/migrate)?
This may only impacts the demo app, not the core code.

### Version of osmdroid the issue relates to:

6.1.5

mat8854:
Migrating is easy. There is an option in Android Studio: https://developer.android.com/jetpack/androidx/migrate

Did it for outdated applications without problems.

InI4:
Well, I did this once and would say it was semi easy. The tool did not get everything right, manual corrections were easy, QA was heavy though.

mat8854:
Okay, the projects where I used the migration assistant were not complex.. But the osmdroid sample application is basically just a ListView and lot of Fragments.

monsieurtanuki:
A first step would be to switch to android support 28 in https://github.com/osmdroid/osmdroid/blob/master/gradle.properties#L44

monsieurtanuki:
The good news is that there's no reason why the core of osmdroid (`osmdroid-android`) should be impacted by this upgrade. That was my intuition, later confirmed by a `grep "android.support"`.
The upgrade makes sense only for the activities and demos, which may use a `ToolBar` or a `FloatingActionButton`.
@mat8854 Feel like upgrading to androidx?

mat8854:
@monsieurtanuki Yes, it is still on a different branch. An automated test suite is nice for such cases (for UI and logic). Other people have more experience with this migration task (@InI4 ).

monsieurtanuki:
I've just tried the first step: switch android.support version to 28.0.0 (from current value: 26.1.0).
That didn't go well.
Does anybody understand the world of gradle? Help!

InI4:
I've got no clue about gradle and I am quite offline lately, but I've managed to that occasionally.
What exactly happened after the change?

monsieurtanuki:
I can't remember exactly, but it's immediate: when you try to sync the gradle/project after typing 28.0.0 instead of 26.1.0, there's an error message at the gradle sync level, and you can't go further.

InI4:
I remember there were two places where you specify target and build sdk and you have to set them consistently. With recent gradle one of the settings can be removed though.

monsieurtanuki:
[...]

PR #1856, opened 2022-10-13
Android-Support to AndroidX Upgrade

PWRxPSYCHO:
After looking into upgrading NGA-Geopackage, it seems that there are too many clashing dependencies from android.support and androidx. I saw there was a feature request to update the older libraries to the newer libraries. 

PR: #1853 - should be done after this to resolve the duplicate file issues

Issue: #1453 

I followed this guide: https://developer.android.com/jetpack/androidx/migrate/artifact-mappings to update all legacy mappings to the equivalent AndroidX

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -9,7 +9,7 @@ org.gradle.configureondemand true
 # "Specifies the jvmargs used for the daemon process.  The setting is particularly useful for
 #  tweaking memory settings.  At the moment, the default settings are pretty generous with regards
 #  to memory."
-#org.gradle.jvmargs=-Xms256m -Xmx2048m -XX:MaxPermSize=256m
+org.gradle.jvmargs=-Xms256m -Xmx2048m -XX:MaxPermSize=256m
 
 
 # Dependency version configurations ----------------------------------------------------------------
--- a/gradle.properties
+++ b/gradle.properties
@@ -41,12 +41,12 @@ gradleFury.version=1.0.14
 
 
 android-plugin.version=7.1.3
-android-support.version=28.0.0
 # To be used for packages that depend on support lib (minSDK = 14 since 26.0.0)
 android-minSdkForSupportLib.version=14
 junit.version=4.13.2
 robolectric.version=4.7.3
 android.useAndroidX=true
+android.enableJetifier=true
 
 # Maven Repository (i.e. Sonatype Nexus Repository Manager) Configuration --------------------------
 
--- a/gradle.properties
+++ b/gradle.properties
@@ -9,7 +9,7 @@ org.gradle.configureondemand true
 # "Specifies the jvmargs used for the daemon process.  The setting is particularly useful for
 #  tweaking memory settings.  At the moment, the default settings are pretty generous with regards
 #  to memory."
-org.gradle.jvmargs=-Xms256m -Xmx2048m -XX:MaxPermSize=256m
+#org.gradle.jvmargs=-Xms256m -Xmx2048m -XX:MaxPermSize=256m
 
 
 # Dependency version configurations ----------------------------------------------------------------
@@ -47,7 +47,6 @@ junit.version=4.13.2
 robolectric.version=4.7.3
 android.useAndroidX=true
 android.enableJetifier=true
-
 # Maven Repository (i.e. Sonatype Nexus Repository Manager) Configuration --------------------------
 
 # username and pass can be stored local.properties with encryption
--- a/OpenStreetMapViewer/build.gradle
+++ b/OpenStreetMapViewer/build.gradle
@@ -98,8 +98,8 @@ dependencies {
     implementation 'com.squareup.leakcanary:leakcanary-android:2.2'
 
     //on device testing
-    androidTestImplementation 'androidx.multidex:multidex:2.0.1'
-    androidTestImplementation "androidx.annotation:annotation:1.5.0"
+    androidTestImplementation 'androidx.multidex:multidex:2.0.0'
+    androidTestImplementation "androidx.annotation:annotation:1.3.0"
 
 
     androidTestImplementation "androidx.test:runner:1.4.0"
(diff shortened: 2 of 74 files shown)


PWRxPSYCHO:
https://github.com/osmdroid/osmdroid/actions/runs/3245835164/jobs/5328503169#step:5:1338

I will investigate this unit test error.

PWRxPSYCHO:
> https://github.com/osmdroid/osmdroid/actions/runs/3245835164/jobs/5328503169#step:5:1338
> 
> I will investigate this unit test error.


It looks like all the tests are passing. There is one test skipped in the WMS parser but that module is still in beta so I don't see an issue with it skipping.
 
Screenshot 2022-10-14 at 08-58-47 Test results - Test Summary
Screenshot 2022-10-14 at 08-58-34 Test results - Test Summary
Screenshot 2022-10-14 at 08-58-16 Test results - Test Summary

gradle.properties
```diff
@@ -41,12 +41,12 @@ gradleFury.version=1.0.14
 
 
 android-plugin.version=7.1.3
-android-support.version=28.0.0
 # To be used for packages that depend on support lib (minSDK = 14 since 26.0.0)
 android-minSdkForSupportLib.version=14
 junit.version=4.13.2
 robolectric.version=4.7.3
 android.useAndroidX=true
+android.enableJetifier=true
```
monsieurtanuki, line 49: Just a minor remark from https://developer.android.com/jetpack/androidx/migrate:
> Additionally, the [enableJetifier](https://developer.android.com/jetpack/androidx/migrate#migrate_an_existing_project_using_android_studio) flag mentioned in this guide can lead to slower build times and should not be used unless it’s necessary.

monsieurtanuki (approved):
Hi @PWRxPSYCHO!
From what I read, this PR looks good to me.
The thing is that I have no idea what the impact of that migration would be on legacy apps.
I assume that in 2022 that should be OK, as long as the min sdk is not changed.

gradle.properties
PWRxPSYCHO: I will address this!

PWRxPSYCHO: Legacy Library

This seems to be the only thing that requires the enableJetifier flag on. The logging would need to be updated as newer versions of ACRA removed `setReportSender()` which is used in `OsmApplication.java`

PWRxPSYCHO: Overall the build time does not seem to be severely impacted by this:
As indicated here
Build

Total time

monsieurtanuki: Thank you @PWRxPSYCHO for your detailed report!

If I sum up:
* the impact of `android.enableJetifier=true` looks minimal, and is _de facto_ limited to the sample app
* we may create an issue about getting rid of `android.enableJetifier=true` that would involve an upgrade of `ch.acra` and some minor refactoring around it

PWRxPSYCHO:
> Hi @PWRxPSYCHO! From what I read, this PR looks good to me. The thing is that I have no idea what the impact of that migration would be on legacy apps. I assume that in 2022 that should be OK, as long as the min sdk is not changed.

@monsieurtanuki I did my best to maintain legacy mappings so they should be 1-1 and cause no breaking changes as far as legacy apps are concerned. The min sdk has not been touched. I did go through all the versions of the dependencies to make sure everything was still compatible.

eclectice:
What is the status of this pull request, and when will it be merged into the master branch? I'd like to write a simple app in Kotlin that makes use of **ViewPager2**.

Only the **osmdroid:master** branch has been tested with the Android Gradle plugin (AGP) 7.5.1, which requires a Java 11 JVM, and it works flawlessly. The broken Simple Map app in this OsmDroid project continues to puzzle me; I remain of the opinion that having the OpenStreetMapViewer app built is adequate.

On the other hand, did you encounter any ***prepareKotlinBuildScriptModel*** error when building **OpenStreetMapViewer** taken from **osmdroid:master** branch with Android Studio (Android Studio Dolphin | 2021.3.1 | Build #AI-213.7172.25.2113.9014738)? 

```
Task 'prepareKotlinBuildScriptModel' not found in project ':OpenStreetMapViewer'
```

To remove the error, I need to include this Gradle groovy snippet in OpenStreetMapViewer's build.gradle:

```groovy
tasks.register("prepareKotlinBuildScriptModel"){}
```

monsieurtanuki:
@spyhunter99 It's probably time to grant merge or/and admin privileges (whatever they're called in github) to contributors that have skills, time and motivation, like @eclectice.

spyhunter99:
invite sent

eclectice:
> @spyhunter99 It's probably time to grant merge or/and admin privileges (whatever they're called in github) to contributors that have skills, time and motivation, like @eclectice.

> invite sent

Thank you for the invitation, but I will KIV it for the time being because I am still unfamiliar with git and the GitHub versioning system; I am more familiar with the Subversion system.

@spyhunter99 I pulled your merge, but when I tested it on a physical Android 11 device, it had a lot of issues (specifically Samsung Galaxy Tab S7 FE SM-T733). Even for simple things like accessing shared preferences, getting a system service context, and so on, the OpenStreetMapViewer app receives a lot of StrictMode disk read/write violations. To reduce them, I use ***StrictMode.allowTgreadDiskWrites()*** whenever possible. 

```java
        StrictMode.ThreadPolicy oldPolicy = null;
        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.GINGERBREAD) {
            oldPolicy = StrictMode.allowThreadDiskWrites();
        }
        try {
            mSensorManager = (SensorManager) context.getSystemService(Context.SENSOR_SERVICE);
        } finally {
            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.GINGERBREAD) {
                StrictMode.setThreadPolicy(oldPolicy);
            }
        }
```

I'm also implementing various permission checks for code that requires runtime permissions. Because there are cleartext network traffic errors when accessing some OpenStreetMap links, I'm adding them to the app's network configuration file. 

My work is still in progress. I will keep you informed when it is done.

eclectice:
When running any sample demo that uses the **GpsMyLocationProvider** on an Android 11 device, it appears that the sample demo receives a leak notification from Leak Canary. Following further investigation, it appears that the Location Service obtains its context from the activity rather than the application context. This is well explained in this article: https://dev.to/pyricau/beware-packagemanager-leaks-223g.

As a result, I changed the majority of the system services in the sample demo to use the application context rather than the activity context, with the exception of ***Context.LAYOUT_INFLATER_SERVICE***, which requires the activity context.

PR #1873, opened 2022-12-03
Fix Mapbox rester image api url. Add support for hightDPI.

xyz32:
According to the MapBox raster images API docs here: https://docs.mapbox.com/api/maps/raster-tiles/

The url for this api has changed to
https://api.mapbox.com/v4/{tileset_id}/{zoom}/{x}/{y}{@2x}.{format}

### DIFF
--- a/osmdroid-android/src/main/java/org/osmdroid/tileprovider/tilesource/MapBoxTileSource.java
+++ b/osmdroid-android/src/main/java/org/osmdroid/tileprovider/tilesource/MapBoxTileSource.java
@@ -131,7 +131,7 @@ public class MapBoxTileSource extends OnlineTileSourceBase {
 		url.append(MapTileIndex.getX(pMapTileIndex));
 		url.append("/");
 		url.append(MapTileIndex.getY(pMapTileIndex));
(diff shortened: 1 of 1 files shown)


spyhunter99:
no idea why this failed ci. it's probably some testing oddity

tvaknin:
Hi,

Since this commit, the static tiles won't work. for example "streets-v11", "navigation-day-v1" that used to work.
What is the alternative method for static tiles? currently, I'm using the version before this commit and it works. 

I think there should be 2 APIs:

As the current one for Raster:

new MapBoxFixRasterAPi("MapBoxStreetsLabelled", MAPBOX_MIN_ZOOM_LEVEL, MAPBOX_MAX_ZOOM_LEVEL, Constants.MAPBOX_TILE_SIZE, ".png");

And for static, same as the previous version.

new MapBoxFixRasterStaticAPi("MapBoxStreetsLabelled", MAPBOX_MIN_ZOOM_LEVEL, MAPBOX_MAX_ZOOM_LEVEL, Constants.MAPBOX_TILE_SIZE, ".png");

PR #1811, opened 2022-05-30
#1783 resolves a NPE related to disableFollowLocation from MyLocation…

spyhunter99:
…NewOverlay, the rest of the changes are test code to reproduce the issue. Note: i tried to rig up an automated test for this but i was not successful.

PR #1877, opened 2022-12-30
Re-apply pull request #1811 with AndroidManifest.xml changes minified

jayaddison:
Reverts and then re-applies #1811 with a reduced -- although should be semantically-equivalent -- set of changes applied to [`OpenStreetMapViewer/src/main/AndroidManifest.xml`](https://github.com/osmdroid/osmdroid/blob/713f680f8e7c56418331415f7dd8adb9b6f43375/OpenStreetMapViewer/src/main/AndroidManifest.xml).

Relates to previous discussion at https://github.com/osmdroid/osmdroid/commit/a9cfce7b7659c2f2443cf739ac5311b3768f82d7#r90401586

In particular, this is intended to address some potentially-confusing relocation of XML comments due to reformatting of that file.


jayaddison:
cc @monsieurtanuki @spyhunter99 - this _should_ be a no-behaviour-change patch (but please don't take my word for that) - the only intent here is to make the XML format more human-readable.

jayaddison:
cc @PWRxPSYCHO in case you have any bandwidth for code review

jayaddison:
Closing as stale; please feel free to recover and merge these commits if you like, though (that offer also extends to any forks of this project).

PR #1879, opened 2023-01-15
DelayTest.check: skip the test if a sleep call exceeds the expected wait time

jayaddison:
The [DelayTest 'check' function's first call to Thread.sleep](https://github.com/osmdroid/osmdroid/blob/713f680f8e7c56418331415f7dd8adb9b6f43375/osmdroid-android/src/test/java/org/osmdroid/util/DelayTest.java#L41) is intended to wait for _some_ but _not all_ of the current delay (backoff) duration.

In practice this seems to be causing some [flaky unit test](https://github.com/osmdroid/osmdroid/actions/runs/3548810663/jobs/5960434258#step:5:1354) failures: my theory is that this occurs when the test thread is interrupted during the sleep, resulting in the [DelayTest.sleep](https://github.com/osmdroid/osmdroid/blob/713f680f8e7c56418331415f7dd8adb9b6f43375/osmdroid-android/src/test/java/org/osmdroid/util/DelayTest.java#L47-L53) call sometimes _exceeding_ the delay duration.

To accommodate that possibility, we can use [JUnit4's Assume class](https://junit.org/junit4/javadoc/4.12/org/junit/Assume.html) to test whether the condition that Delay.shouldWait is true.

If the condition does _not_ hold -- that is, the DelayTest.sleep exceeded the expected delay duration -- then ~~JUnit~~ the default JUnit test runner will skip the test (the condition for testing didn't hold, so we can't rely on the results).

Please note: untested locally; this is a theory, so far.

cc @monsieurtanuki

### DIFF
--- a/osmdroid-android/src/test/java/org/osmdroid/util/DelayTest.java
+++ b/osmdroid-android/src/test/java/org/osmdroid/util/DelayTest.java
@@ -1,6 +1,7 @@
 package org.osmdroid.util;
 
 import org.junit.Assert;
+import org.junit.Assume;
 import org.junit.Test;
 
 /**
@@ -39,7 +40,7 @@ public class DelayTest {
 
     private void check(final Delay pDelay, final long pMillis) {
         sleep((pMillis * 3) / 4);
-        Assert.assertTrue(pDelay.shouldWait());
+        Assume.assumeTrue(pDelay.shouldWait());
         sleep(pMillis / 2);
         Assert.assertFalse(pDelay.shouldWait());
     }


monsieurtanuki (commented):
Hi @jayaddison!
I'm not convinced by your fix, but I'm not familiar with "assume".
What your fix looks like is "when this particular test fails, just ignore it". I don't see the point of the test then. More specifically, you skip the rest of the tests, which may be more relevant.
If this specific test is sometimes false, it could be because the delay is too short to be consistently respected given possible side effects (slow cpu, gc) and the fact that we're talking about millis.
If needed, I would prefer to remove that test altogether, or comment it with "// would be the best case scenario but sometimes the cpu is too slow".
Does that make sense?

jayaddison:
Yep, thanks - that makes sense.  It's definitely not an ideal fix.

The goals I had in mind were:

- Increase the relevance of continuous integration failures (a failure should mean: stop and look at this; there is a real problem here)
- Improve safety for future fixes and refactors (developers should not accidentally break code without being informed)
- Catch external changes (environments, resources, dependencies) that cause regressions to the project

> If this specific test is sometimes false, it could be because the delay is too short to be consistently respected

Yep, agreed that the problem could be something like that.  In particular: if failures still occur with these changes applied, then that would increase the likelihood of the delay-too-short theory.

jayaddison:
One other thing that puzzled me: the zero-nanos argument _shouldn't_ make a difference, I suppose?  (in practice, though, it could do, depending on the implementation.  that could be another path to explore)

jayaddison:
Closing as stale; please feel free to recover and merge these commits if you like, though (that offer also extends to any forks of this project).

PR #1838, opened 2022-08-31
Upgrade to "tile.openstreetmap.org"

monsieurtanuki:
"a.", "b.", "c." are not needed anymore, cf. https://github.com/openstreetmap/operations/issues/737

PR #1906, opened 2023-06-30
Use preferred OSM tile URL

otbutz:
see https://github.com/openstreetmap/operations/issues/737


monsieurtanuki (approved):
Thank you @otbutz!
This PR looks very similar to #1838 ;)

spyhunter99:
yup, this is a duplicate, the change is already in place unless i'm missing something. i'll cut a release shortly

PR #2002, opened 2024-05-22
Implement equals and hashCode for BoundingBox

Maradox:
Implements `equals` and `hashCode` for the `BoundingBox`  class.


spyhunter99:
looks good, thanks

Issue #1950, opened 2023-10-25
How should I use user-supplied maps?


y20k:
## Issue Type

[x] Question

## Description

I am the developer of [Trackbook](https://codeberg.org/y20k/trackbook), a small app that records hikes. The app uses osmdroid. I recently got the [request](https://codeberg.org/y20k/trackbook/issues/121) to add offline maps. 

My approach: 
1. ask the user to provide a folder that contains .map files
2. render the offline maps using [MapsForgeTileSource](https://osmdroid.github.io/osmdroid/javadocAll/org/osmdroid/mapsforge/MapsForgeTileSource.html)

My problem: The files/directories, that the app receives from the system file picker, are of the type [DocumentFile](https://developer.android.com/reference/androidx/documentfile/provider/DocumentFile). MapsForgeTileSource only works with [File](https://developer.android.com/reference/java/io/File). 

I know that there is a (closed) issue #1591 for this problem. So my question rather would be: 

_Did I choose the right approach, given the premise, that I want to use user-supplied maps?_

y20k:
Hi @spyhunter99 

> might want to check with the maps force folks to see if they have alternate apis that we could potentially wire in an input stream instead of file.

As far as I can see the constructor for osmdroid's [MapsForgeTileSource](https://github.com/osmdroid/osmdroid/blob/72382651fc3a99f8ea89543913b822bd9508fbdf/osmdroid-mapsforge/src/main/java/org/osmdroid/mapsforge/MapsForgeTileSource.java#L64) takes [File[]](https://developer.android.com/reference/java/io/File) and uses it internally to create Mapsforge [MapFile](https://github.com/mapsforge/mapsforge/blob/master/mapsforge-map-reader/src/main/java/org/mapsforge/map/reader/MapFile.java)s.

Mapfile also has a constructor that accepts a [FileInputStream](https://github.com/mapsforge/mapsforge/blob/67be471da5da31b8d12ee0a602e2bdd4b4f4f3ee/mapsforge-map-reader/src/main/java/org/mapsforge/map/reader/MapFile.java#L275)

**My suggestion would be**: MapsForgeTileSource should be able to accept FileInputStream[].

In my scenario this would work: You can create a MapFile from a DocumentFile. I tested that in my app. (The Android ContentResolver only lets me open an InputStream, but that can be casted to FileInputsStream, since the InputStream comes from a file.) 

If implemented MapsForgeTileSource would work with the Android Storage Access Framework.

[...]

PR #2020, opened 2024-07-08
Create a MapsForgeTileSource from a FileInputStream

y20k:
This commit adds the ability to create a MapsForgeTileSource from a FileInputStream. This should add the ability to use MapsForgeTileSource with files coming from Android's new(ish) [Storage Access Framework](https://developer.android.com/guide/topics/providers/document-provider). It leverages the fact that MapFile also has a constructor that accepts a FileInputStream. See issue #1950


spyhunter99:
awesome thanks!

y20k:
Hi @spyhunter99  

Thanks. That is really helpful. I am happy to report that I now can load .map files that are returned from the Android file picker via the Storage Access Framework. (some theming might still be necessary 😄)
Screenshot_20240828_145518

PR #2040, opened 2024-08-30
Updated gradle to current 8.7 and sdk to 34

LostRat:
I temporarily comment out strict policy to install on local device and run. OpenStreetMapViewer app only. Random testing of a few options in the app. 

I updated gradle stuff and sdk incrementally. Making suggested changes to sucessfully sync and compile.

Had to leave leakcanary implimentation commented out. Otherwise Android Studio demands android:exported in a manifest for a leaklauncher file/entry that I cannot find.

I can only hope that you can at least see the changes I made.

My push attempt from local to fork tried to push a change I made last year. 

Someone else will have to run the routine to generate and test output.

### DIFF
--- a/OpenStreetMapViewer/build.gradle
+++ b/OpenStreetMapViewer/build.gradle
@@ -8,6 +8,7 @@ group = project.property("pom.groupId")
 version = project.property("pom.version")
 
 android {
+    namespace "org.osmdroid"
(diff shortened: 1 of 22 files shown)


LostRat:
It seems I need to first figure out how make my local code match exactly the latest master, before making changes. Will try to figure it out again.

